PR #11, opened 2022-09-27
Update godot-lib to master 84f2c68c84bde10026454292578e5fdcc9ab69a3

BastiaanOlij:
Updating the godot-lib entries to the current master now that our dynamic logic has been merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.iml
 .gradle
+.vscode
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
Cancelling this as not the right way forward. Will be changing the build scripts soon to download the godot-lib from Godots download pages.

PR #12, opened 2022-10-18
Download godot-lib and update build files to new name

BastiaanOlij:
Changed to new beta 3 naming for godot-lib and instead of including this file in our repo, downloading official release from the Godot website.
We will need to keep this updated when newer versions come out.


BastiaanOlij:
This is no longer needed, the solution in #13 is a better approach

PR #20, opened 2023-01-27
Add Generic KHR loader support

rbessems:
This MR adds support for the standard Khronos OpenXR loader. Tested on Magic Leap 2.


BastiaanOlij:
Thats amazing Ron!! Thanks! I'll do a proper review later in the weekend.

@m4gr3d can you review as well please?

godotopenxrkhr/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+        targetSdk versions.targetSdk
+        versionCode versions.versionCode
+        versionName versions.versionName
+
+        ndk {
+            abiFilters 'arm64-v8a', 'armeabi-v7a',  'x86',  'x86_64'
+        }
+    }
+
+    namespace = "org.godotengine.openxrloaders.godotopenxrkhr"
```
m4gr3d, line 19: To match the format used for the other loaders, here and in the rest of the PR, the namespace should be `org.godotengine.openxrloaders.khr`

m4gr3d (dismissed):
Looks good! Just some feedback about updating the namespace and package name and it should be good to merge.

godotopenxrkhr/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,23 @@
...
+    <application>
+        <meta-data
+            android:name="org.godotengine.plugin.v1.GodotOpenXRKHR"
+            android:value="org.godotengine.openxrloaders.khr.GodotOpenXRKHR" />
+
+        <activity android:name="com.godot.game.GodotApp" android:exported="true" tools:ignore="MissingClass">
+            <intent-filter>
+                <!-- Mark this activity as using OpenXR. -->
+                <!-- action tag is here to make sure this intent filter is loaded by AOSP. -->
+                <action android:name="org.khronos.openxr.action.MAIN" />
```
rbessems, line 17: AOSP filters intent-filters without actions out.

https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/content/pm/PackageParser.java#4575

So we need a 'fake' action here.

godotopenxrkhr/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,23 @@
...
+
+    <uses-permission android:name="org.khronos.openxr.permission.OPENXR_SYSTEM" />
+
+    <application>
+        <meta-data
+            android:name="org.godotengine.plugin.v1.GodotOpenXRKHR"
+            android:value="org.godotengine.openxrloaders.khr.GodotOpenXRKHR" />
+
+        <activity android:name="com.godot.game.GodotApp" android:exported="true" tools:ignore="MissingClass">
+            <intent-filter>
```
rbessems, line 14: Tried adding tools:node="merge" to intent-filter tag but bundletool refuses to merge the toplevel intent-filter with this. Hence the work-around below.

godotopenxrkhr/build.gradle
rbessems: Done!

BastiaanOlij (approved):
lgtm, I think all issues are addressed.

godotopenxrkhr/src/main/AndroidManifest.xml
m4gr3d: @rbessems Not sure I understand the reason for this addition, the `GodotApp` activity, which this intent-filter will be merged into already has a defined action, so this fake action should not be needed.

godotopenxrkhr/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,23 @@
...
+                <!-- Mark this activity as using OpenXR. -->
+                <!-- action tag is here to make sure this intent filter is loaded by AOSP. -->
+                <action android:name="org.khronos.openxr.action.MAIN" />
+                <category android:name="org.khronos.openxr.intent.category.IMMERSIVE_HMD" />
```
m4gr3d, line 18: @rbessems Is this a valid category? If it's, can you add a comment that links to its description.

Also it seems the purpose of this intent-filter is to merge this category with Godot's `GodotApp` activity. The `GodotApp` activity is specific to the Godot Editor, so these changes should be made directly to the core Godot project's AndroidManifest.
 
The `<activity` should not be include in this loader library since it can be integrated by projects that don't include Godot's default `GodotApp` activity. Instead the documentation should be updated to inform users of this library to add the necessary category to their main activity.

godotopenxrkhr/src/main/AndroidManifest.xml
rbessems: Two problems coming together here.

1. intent-filters are not merged. I've tried convincing build tool to do so by adding tools:node="merge" to the intent-filter. However the merged AndroidManifest always gets a 2nd intent-filter.
2. AOSP when loading the AndroidManifest will silently drop any intent filter without an action. This means the category never shows up and the application isn't marked as an OpenXR/AR application.

godotopenxrkhr/src/main/AndroidManifest.xml
rbessems: Yes valid: 

https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#android-runtime-category

Happy to put up another MR removing this. What is the best spot to add the documentation?

rbessems: It would be awesome if Godot adds this automatically for the default template,... wonder what the best way is to do that.

godotopenxrkhr/src/main/AndroidManifest.xml
m4gr3d: Both issues should be solvable by the option I mentioned below; the `category` addition to the `intent-filter` can be moved to the core Godot's [AndroidManifest generation logic](https://github.com/godotengine/godot/blob/master/platform/android/export/gradle_export_util.cpp), or better yet can be added directly to the [`AndroidManifest` just like we do for Meta's](https://github.com/godotengine/godot/blob/0a9e6e478e3183d7bf3c5cb895c706bc6275d3ea/platform/android/java/app/AndroidManifest.xml#L69) if its inclusion does not have side-effect for regular Godot Android apps running on regular devices (phone, tablet, etc).

godotopenxrkhr/src/main/AndroidManifest.xml
m4gr3d: Added a comment above describing the options you can use in core Godot to include this automatically. Let me know if you have questions.

godotopenxrkhr/src/main/AndroidManifest.xml
rbessems: Sounds good, I'll move this category over there.

godotopenxrkhr/src/main/AndroidManifest.xml
m4gr3d: > What is the best spot to add the documentation?

It can be added both in the `README`, and in `GodotOpenXRKHR.java` as class javadocs (the javadocs is exported when this project is used as a library, so integrators can reference it).

@BastiaanOlij Any other place where the documentation should be included?

godotopenxrkhr/src/main/AndroidManifest.xml
rbessems: @m4gr3d adding it to the AndroidManifest like it's done for Meta feels wrong, this tag should really only appear if this an OpenXR activity. 

I've added it to the generator logic, this adds a second intent-filter to the activity which tags the whole activity as an openxr activity. When an application is loaded Android removes any intent-filter without an action, so I add the placeholder action to avoid that.

I've tried to get bundletool to merge the two intent-filters but it is refusing.

m4gr3d: > adding it to the AndroidManifest like it's done for Meta feels wrong, this tag should really only appear if this an OpenXR activity.

@rbessems  yeah, you're correct. The Meta tag being there by default is also incorrect, so let's clean up the manifest.
This approach was previously used to allow compatibility with Godot's legacy build system for Android, but that build system is deprecated in Godot 4 for XR project, so the Meta category within the AndroidManifest can be removed.

> I've added it to the generator logic, this adds a second intent-filter to the activity which tags the whole activity as an openxr activity. When an application is loaded Android removes any intent-filter without an action, so I add the placeholder action to avoid that.

Sounds good, I'll take a look! Having the two intent-filters is supported and should work as expected, so we can leave it as is.

For reference, according to the doc, `intent-filter` are never merged - https://developer.android.com/studio/build/manage-manifests#merge_conflict_heuristics

godotopenxrkhr/src/main/AndroidManifest.xml
BastiaanOlij: I think the readme is the best location seeing this is of value to anyone who wishes to either add more to this repo or as we slowly improve the build system. 

We're probably too late to do major changes for 4.0 especially around the Android export in core, but as things evolve we'll want to refer back to decisions we've made here.

PR #29, opened 2023-02-14
Adding missing focusaware entry to meta manifest

decacis:
This PR adds a missing entry in the AndroidManifest.xml for the meta loader, that is required when uploading an application to either App Lab or the main Meta Quest store.

### DIFF
--- a/godotopenxrmeta/src/main/AndroidManifest.xml
+++ b/godotopenxrmeta/src/main/AndroidManifest.xml
@@ -15,6 +15,10 @@
         <meta-data
             android:name="com.oculus.supportedDevices"
             android:value="quest|quest2|questpro" />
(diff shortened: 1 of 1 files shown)


m4gr3d:
@decacis We removed that manifest entry because it's now enabled by default for all Meta apps. Did you get an error when uploading to the store?

decacis:
@m4gr3d Indeed, I get an error telling me that the focusaware entry is required:
Screamer-Distant-Flashy

I am using Godot 4.0 RC 2 with the latest release of the OpenXR loaders (alpha 4) and MQDH version 3.2.0. Adding the entry resolves the issue.

m4gr3d (approved):

PR #37, opened 2023-06-30
Add additional feature and permissions settings to Khronos manifest

BastiaanOlij:
Comparing our manifest with Khronos' hello-xr example, we're missing a feature and permission.

### DIFF
--- a/godotopenxrkhr/src/main/AndroidManifest.xml
+++ b/godotopenxrkhr/src/main/AndroidManifest.xml
@@ -3,6 +3,14 @@
     package="org.godotengine.openxrloaders.khr"
     xmlns:tools="http://schemas.android.com/tools">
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
@m4gr3d if you haven't included these already, you might want to add this into your refactor PR?

m4gr3d:
> @m4gr3d if you haven't included these already, you might want to add this into your refactor PR?

These entries can be left here since they don't require specific configurations; they will be added to the generated binary via manifest merging with the manifest that's being edited.

In the export plugin logic, we only want to include manifest entries that are either optional or have multiple values or that depend on a specific Godot manifest element (e.g: `GodotApp` activity element entry).

m4gr3d (approved):

PR #64, opened 2023-12-02
Build extension for all supported platforms

BastiaanOlij:
Working on altering build system to include all supported systems.

### DIFF
--- a/.github/workflows/build-addon-on-push.yml
+++ b/.github/workflows/build-addon-on-push.yml
@@ -6,8 +6,48 @@ on:
 
 jobs:
   build:
-    name: Assembling artifacts
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
I'm having issues cross compiling on linux, so disabled that for now. Still need to unpack the artifacts to create the final release but waiting for the builds to finish to see what they look like

BastiaanOlij:
Need to do some final tests but I think most of it is working

BastiaanOlij:
OK, as far as i've been told, we can't do a linux ARM64 or RV64 build on github so I'm leaving those out.

But it now build properly for Windows, Mac, Linux x86 and Android as far as I can tell.

m4gr3d:
I'll create a `2.0.2-stable` tag to push the fixes up.

PR #66, opened 2023-12-09
Add OpenXR extension wrappers for fb_scene, fb_spatial_entity, fb_spatial_entity_query, fb_spatial_entity_container

maunvz:
As title, following the pattern from the fb_scene_capture extension wrapper, I've added wrappers for the 4 extensions mentioned. They don't yet expose useful functionality to godot projects (beyond getting the singletons and checking if the extensions are supported), but provide most of the openxr related boilterplate so a later change can add an "XR Scene" abstraction that expresses a physical room's layout and surfaces to the a Godot scene.

### DIFF
--- a/godotopenxrmeta/src/main/cpp/export/meta_export_plugin.h
+++ b/godotopenxrmeta/src/main/cpp/export/meta_export_plugin.h
@@ -83,6 +83,7 @@ private:
 	Dictionary _hand_tracking_frequency_option;
 	Dictionary _passthrough_option;
 	Dictionary _use_anchor_api_option;
+	Dictionary _use_scene_api_option;
 	Dictionary _support_quest_1_option;
 	Dictionary _support_quest_2_option;
 	Dictionary _support_quest_3_option;
--- a/godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
+++ b/godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
@@ -1,0 +1,121 @@
+/**************************************************************************/
+/*  openxr_fb_scene_extension_wrapper.h                                   */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#pragma once
+
+#include <godot_cpp/classes/open_xr_extension_wrapper_extension.hpp>
+#include <openxr/openxr.h>
+#include <openxr/fb_scene.h>
+#include <godot_cpp/templates/hash_map.hpp>
+#include <godot_cpp/templates/vector.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+#include "util.h"
+
+#include <optional>
+#include <map>
+
+using namespace godot;
+
+struct XrSceneObjectInternal {
+	XrUuidEXT uuid;
+	XrSpace space;
+	std::optional<String> label;
+
+	// Vertices and lines on a plane, probably use this for a floor / ceiling as they are irregularly shaped
+	// We store a std::vector instead of XrBoundary2DFB to own the vertex memory
+	std::optional<Vector<XrVector2f>> boundary2D;
+	// A rectangle containing the whole thing, perfect for desks / tables / play surfaces
+	std::optional<XrRect2Df> boundingBox2D;
+	// 3D box for the whole thing, better for obstacles and other objects not used as a surface
+	std::optional<XrRect3DfFB> boundingBox3D;
+};
+
+// Wrapper for the set of Facebook XR scene extension.
+class OpenXRFbSceneExtensionWrapper : public OpenXRExtensionWrapperExtension {
+	GDCLASS(OpenXRFbSceneExtensionWrapper, OpenXRExtensionWrapperExtension);
+
+public:
+	Dictionary _get_requested_extensions() override;
+
+	void _on_instance_created(uint64_t instance) override;
+
+	void _on_instance_destroyed() override;
+
+	bool is_scene_supported() {
+		return fb_scene_ext;
+	}
+
+	static OpenXRFbSceneExtensionWrapper *get_singleton();
+
+	OpenXRFbSceneExtensionWrapper();
+	~OpenXRFbSceneExtensionWrapper();
+
+	std::optional<String> get_semantic_labels(const XrSpace& space);
+	void get_shapes(const XrSpace& space, XrSceneObjectInternal& object);
+
+protected:
+	static void _bind_methods();
+
+private:
+	EXT_PROTO_XRRESULT_FUNC3(xrGetSpaceBoundingBox2DFB,
+			(XrSession), session,
+			(XrSpace), space,
+			(XrRect2Df *), boundingBox2DOutput)
+
+	EXT_PROTO_XRRESULT_FUNC3(xrGetSpaceBoundingBox3DFB,
+			(XrSession), session,
+			(XrSpace), space,
+			(XrRect3DfFB *), boundingBox3DOutput)
+
+	EXT_PROTO_XRRESULT_FUNC3(xrGetSpaceSemanticLabelsFB,
+			(XrSession), session,
+			(XrSpace), space,
+			(XrSemanticLabelsFB *), semanticLabelsOutput)
+
+	EXT_PROTO_XRRESULT_FUNC3(xrGetSpaceBoundary2DFB,
+			(XrSession), session,
+			(XrSpace), space,
+			(XrBoundary2DFB *), boundary2DOutput)
+
+	EXT_PROTO_XRRESULT_FUNC3(xrGetSpaceRoomLayoutFB,
+			(XrSession), session,
+			(XrSpace), space,
+			(XrRoomLayoutFB *), roomLayoutOutput)
+
+	bool initialize_fb_scene_extension(const XrInstance instance);
+
+	HashMap<String, bool *> request_extensions;
+
+	void cleanup();
+
+	static OpenXRFbSceneExtensionWrapper *singleton;
+
+	bool fb_scene_ext = false;
+};
--- a/godotopenxrmeta/src/main/cpp/include/openxr_fb_spatial_entity_query_extension_wrapper.h
+++ b/godotopenxrmeta/src/main/cpp/include/openxr_fb_spatial_entity_query_extension_wrapper.h
@@ -1,0 +1,102 @@
+/**************************************************************************/
+/*  openxr_fb_spatial_entity_query_extension_wrapper.h                    */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#pragma once
+
+#include <godot_cpp/classes/open_xr_extension_wrapper_extension.hpp>
+#include <openxr/openxr.h>
+#include <openxr/fb_spatial_entity.h>
+#include <openxr/fb_spatial_entity_query.h>
+#include <godot_cpp/templates/hash_map.hpp>
+#include <godot_cpp/templates/vector.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+#include "util.h"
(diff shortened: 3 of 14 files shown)


godotopenxrmeta/src/main/cpp/openxr_fb_spatial_entity_container_extension_wrapper.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+std::vector<XrUuidEXT> OpenXRFbSpatialEntityContainerExtensionWrapper::get_contained_uuids(XrSpace space) {
+    XrSpaceContainerFB spaceContainer = {XR_TYPE_SPACE_CONTAINER_FB};
+    xrGetSpaceContainerFB(SESSION, space, &spaceContainer);
+    std::vector<XrUuidEXT> uuids(spaceContainer.uuidCountOutput);
+    spaceContainer.uuidCapacityInput = uuids.size();
+    spaceContainer.uuids = uuids.data();
+    xrGetSpaceContainerFB(SESSION, space, &spaceContainer);
+		return uuids;
+}
```
m4gr3d, line 109: nit: the indents used in this method are different from the ones used in the previous methods.

m4gr3d (changes requested):
Looks good overall as a first step to add support for these extensions!

@BastiaanOlij @Malcolmnixon As a follow-up to this PR, we'll need to evaluate how we want to expose this functionality on the Godot side in a manner that's consistent.

BastiaanOlij:
> @BastiaanOlij @Malcolmnixon As a follow-up to this PR, we'll need to evaluate how we want to expose this functionality on the Godot side in a manner that's consistent.

In Godot 3 we introduced Anchors based on how ARKit solved this, most of this logic is still in place but not all.
Whether detected by plane detection or otherwise detected, any location that was tracked in space was created as an `XRPositionalTracker` with the type set to `XRServer::TRACKER_ANCHOR`. While this location was mostly static, ARKit would update locations as it learned more about its surroundings.

So that part of the solution I think is sound and we should apply here too. I don't know the details of the flow Meta currently has but if any spatial location is detected, we should create a positional tracker and register it with the XRServer by calling `XRServer::add_tracker`. The XRServer already emits signals so the game can react to the creation of a new tracker and make any additions to the scene. This entails adding an `XRAnchor3D` node to the scene that is bound to the newly created tracker.

Using `XRAnchor3D` also ensures that the location is adjusted when we recenter the player. 

What is missing is what to do after this, I wasn't at all happy with how we did things in Godot 3 so I left this out in the conversion to Godot 4. We need a mechanism to expose further meta data about the anchor and provide a collision shape and/or occlusion mesh.

These could become `OpenXRCollisionShape` and `OpenXRMesh` classes where you specify the tracker, the problem is that this would suggest this would also apply for other trackers such as controllers.
This is how we did it in Godot 3 where OpenVR also allowed us to provide render models in this fashion however with OpenXR this is no longer sufficient. With OpenXR the vendor render model APIs are all GLTF based and it looks like this is a trend. We could paint ourselves in a corner if this trend continues into a more generic AR solution.

godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
```diff
@@ -0,0 +1,125 @@
...
+/*  openxr_fb_scene_extension_wrapper.h                                   */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Original contributed implementation:                                   */
+/*   Copyright (c) 2022-2023 MattaKis Consulting Kft. (Migeran)           */
```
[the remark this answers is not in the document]
maunvz: updated all headers to match that pattern (also switched to #pragma once)

maunvz:
addressed the formatting nits (now using tabs throughout, turned on whitespace rendering in my editor to catch these sooner lol). Is there / should there be a linter for this project?

BastiaanOlij:
> addressed the formatting nits (now using tabs throughout, turned on whitespace rendering in my editor to catch these sooner lol). Is there / should there be a linter for this project?

We use `clang-format` in most of our projects but doesn't look like we set that up here. Probably worth doing.

maunvz:
Added an export option to request the USE_SCENE permission as per https://developer.oculus.com/documentation/native/native-spatial-data-perm/

From a discussion in #68 the permission is already auto-requested by GodotActivity's existing logic

godotopenxrmeta/src/main/cpp/export/meta_export_plugin.h
```diff
@@ -52,6 +52,10 @@ static const int HAND_TRACKING_REQUIRED_VALUE = 2;
 static const int HAND_TRACKING_FREQUENCY_LOW_VALUE = 0;
 static const int HAND_TRACKING_FREQUENCY_HIGH_VALUE = 1;
 
+static const int BOUNDARY_ENABLED_VALUE = 0;
+static const int BOUNDARY_DISABLED_VALUE = 1;
+static const int BOUNDARY_CONTEXTUAL_VALUE = 2;
```
m4gr3d, line 57: Do these constants end up being used? I don't see them referenced in the code.

m4gr3d (approved):
Looks good!

godotopenxrmeta/src/main/cpp/openxr_fb_scene_extension_wrapper.cpp
```diff
@@ -0,0 +1,162 @@
...
+	request_extensions[XR_FB_SCENE_EXTENSION_NAME] = &fb_scene_ext;
+	singleton = this;
+}
+
+OpenXRFbSceneExtensionWrapper::~OpenXRFbSceneExtensionWrapper() {
+	cleanup();
+}
+
+void OpenXRFbSceneExtensionWrapper::_bind_methods() {
+	ClassDB::bind_static_method("OpenXRFbSceneExtensionWrapper", D_METHOD("get_singleton"), &OpenXRFbSceneExtensionWrapper::get_singleton);
```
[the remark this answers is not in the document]
m4gr3d: I wasn't aware of the `reigster_singleton(...)` approach; let's use that instead and update `OpenXRFbSceneCaptureExtensionWrapper` accordingly!

godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
```diff
@@ -0,0 +1,119 @@
...
+
+#include <optional>
+#include <map>
+
+using namespace godot;
+
+struct XrSceneObjectInternal {
+	XrUuidEXT uuid;
+	XrSpace space;
+	std::optional<std::string> label;
```
BastiaanOlij, line 47: Please use Godots implementations whenever possible to prevent needless conversions of data between internal data and Godot.

godotopenxrmeta/src/main/cpp/export/meta_export_plugin.h
maunvz: removed in latest update, leftover from when I split the export logic from my other pull request that added boundary-less mode and other features

godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
maunvz: This collection of OpenXR extensions use XrSpace and XrUuidEXT across multiple API calls to request different metadata and poses (for example, once you've queried for all available anchors, you can locate that anchor or ask about it's semantic labels / bounding volumes, all with uuid / space). The intent here is to store in the original OpenXR formats to facilitate later calls associated with the same anchor, and translate to a Godot type in some higher level "XrSceneManager" that gets used by gdscript (rather than using these wrappers directly)

godotopenxrmeta/src/main/cpp/openxr_fb_scene_extension_wrapper.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+std::optional<std::string> OpenXRFbSceneExtensionWrapper::get_semantic_labels(const XrSpace space) {
+	if (!OpenXRFbSpatialEntityExtensionWrapper::get_singleton()->is_component_enabled(space, XR_SPACE_COMPONENT_TYPE_SEMANTIC_LABELS_FB)) {
+		return std::nullopt;
+	}
+
+	static const std::string recognizedLabels =
+			"TABLE,COUCH,FLOOR,CEILING,WALL_FACE,WINDOW_FRAME,DOOR_FRAME,STORAGE,BED,SCREEN,LAMP,PLANT,WALL_ART,INVISIBLE_WALL_FACE,OTHER";
+	const XrSemanticLabelsSupportInfoFB semanticLabelsSupportInfo = {
```
m4gr3d, line 100: Should this list also include `GLOBAL_MESH` as mentioned in the documentation?

godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
```diff
@@ -0,0 +1,119 @@
...
+		return fb_scene_ext;
+	}
+
+	static OpenXRFbSceneExtensionWrapper *get_singleton();
+
+	OpenXRFbSceneExtensionWrapper();
+	~OpenXRFbSceneExtensionWrapper();
+
+	std::optional<std::string> get_semantic_labels(const XrSpace space);
+	void get_shapes(const XrSpace space, XrSceneObjectInternal& object);
```
m4gr3d, line 79: Here and throughout the PR, we should use a `const XrSpace &space` argument instead.

m4gr3d (changes requested):

godotopenxrmeta/src/main/cpp/openxr_fb_scene_extension_wrapper.cpp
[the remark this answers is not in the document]
m4gr3d: @dsnopek What's the earlier that `Engine::get_singleton()->register_singleton(...)` can be used?

I tried invoking it during `MODULE_INITIALIZATION_LEVEL_CORE` and I kept getting crash errors because the `Engine` singleton was not available.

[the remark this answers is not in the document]
m4gr3d: Also, doing `Engine.get_singleton(...)` does not retrieve the singleton instance.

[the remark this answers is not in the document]
m4gr3d: Here's a draft PR for reference: https://github.com/GodotVR/godot_openxr_vendors/pull/72

godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
maunvz: hm does that actually make a difference? the XrSpace, like many other XR handles, is just a pointer / uint64_t, not a full object. See openxr.h:

```
#if !defined(XR_DEFINE_HANDLE)
#if (XR_PTR_SIZE == 8)
    #define XR_DEFINE_HANDLE(object) typedef struct object##_T* object;
#else
    #define XR_DEFINE_HANDLE(object) typedef uint64_t object;
#endif
#endif

...

XR_DEFINE_HANDLE(XrSpace)
```
 Will update to match the style of the rest of the codebase

godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
```diff
@@ -0,0 +1,119 @@
...
+using namespace godot;
+
+struct XrSceneObjectInternal {
+	XrUuidEXT uuid;
+	XrSpace space;
+	std::optional<std::string> label;
+
+	// Vertices and lines on a plane, probably use this for a floor / ceiling as they are irregularly shaped
+	// We store a std::vector instead of XrBoundary2DFB to own the vertex memory
+	std::optional<std::vector<XrVector2f>> boundary2D;
```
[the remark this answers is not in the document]
maunvz: related: If I have "using namespace godot;" at the top, should I omit the godot:: prefix throughout the rest of the file (including in headers and variable types)?

godotopenxrmeta/src/main/cpp/openxr_fb_scene_extension_wrapper.cpp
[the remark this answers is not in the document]
maunvz: I'm getting segfaults when registering using this approach. Did a bit of digging, and I think it's due to Engine::get_singleton() not being available in a gdextension

```
ERROR: Failed to retrieve non-existent singleton 'Engine'.
   at: get_singleton_object (core/config/engine.cpp:270)
ERROR: Parameter "singleton_obj" is null.
   at: get_singleton (thirdparty/godot-cpp/gen/src/classes/engine.cpp:49)
```

Digging into the code, godot-cpp auto-generates bindings that eventually call into Engine::get_singleton_object(const StringName &p_name) and passes the class name (in this case "Engine"), but godot never actually registers Engine itself as a singleton in this form, within the godot/ source tree it's just a raw pointer, and since it doesn't extend Object it's not trivial to register like all the other Singleton instances. From godot/core/config/engine.cpp

```
Engine *Engine::singleton = nullptr;

Engine *Engine::get_singleton() {
	return singleton;
}
```

I think we'd need to address that in the engine before we can use Engine::get_singleton() from an extension? In contrast, main.cpp registers Performance as follows: ```engine->add_singleton(Engine::Singleton("Performance", performance));```

 Not sure if you encountered something similar in #73 @dsnopek

godotopenxrmeta/src/main/cpp/include/openxr_fb_scene_extension_wrapper.h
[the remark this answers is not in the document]
m4gr3d: Yeah, you can omit the `godot::` prefix.

godotopenxrmeta/src/main/cpp/openxr_fb_scene_extension_wrapper.cpp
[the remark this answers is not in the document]
m4gr3d: @maunvz There was some logic missing in `register_types.cpp` that @dsnopek fixed in https://github.com/GodotVR/godot_openxr_vendors/pull/73/files#diff-aaccbe3504fdb5822961fde8149e985f5539985f9e60a4a31d804f3bf5feb5e4.

In the updated [`register_types.cpp`](https://github.com/GodotVR/godot_openxr_vendors/pull/73/files#diff-aaccbe3504fdb5822961fde8149e985f5539985f9e60a4a31d804f3bf5feb5e4), the singleton registration is initialized at the `MODULE_INITIALIZATION_LEVEL_SCENE` level and the minimum initialization level was also updated to `MODULE_INITIALIZATION_LEVEL_SCENE`. 
With those changes in place, registration via the engine works as expected.

maunvz:
updated XrSpace usage, removed unnecessary godot:: prefixes, switched to Godot types throughout, and switched to new singleton registration

godotopenxrmeta/src/main/cpp/openxr_fb_scene_extension_wrapper.cpp
m4gr3d: @maunvz ping on this ^^

godotopenxrmeta/src/main/cpp/openxr_fb_scene_extension_wrapper.cpp
```diff
@@ -0,0 +1,162 @@
...
+void OpenXRFbSceneExtensionWrapper::_on_instance_destroyed() {
+	cleanup();
+}
+
+std::optional<String> OpenXRFbSceneExtensionWrapper::get_semantic_labels(const XrSpace& space) {
+	if (!OpenXRFbSpatialEntityExtensionWrapper::get_singleton()->is_component_enabled(space, XR_SPACE_COMPONENT_TYPE_SEMANTIC_LABELS_FB)) {
+		return std::nullopt;
+	}
+
+	static const std::string recognizedLabels =
```
m4gr3d, line 98: Another usage of `std:string`

godotopenxrmeta/src/main/cpp/openxr_fb_scene_extension_wrapper.cpp
```diff
@@ -0,0 +1,162 @@
...
+		XR_SEMANTIC_LABELS_SUPPORT_ACCEPT_INVISIBLE_WALL_FACE_BIT_FB,
+		recognizedLabels.c_str(),
+	};
+
+	XrSemanticLabelsFB labels = {XR_TYPE_SEMANTIC_LABELS_FB, &semanticLabelsSupportInfo, 0};
+
+	// First call.
+	xrGetSpaceSemanticLabelsFB(SESSION, space, &labels);
+	// Second call
+	std::vector<char> labelData(labels.bufferCountOutput);
```
m4gr3d, line 113: Another usage of `std::vector`

godotopenxrmeta/src/main/cpp/openxr_fb_spatial_entity_extension_wrapper.cpp
```diff
@@ -0,0 +1,156 @@
...
+		return true;
+	}
+
+	return false;
+}
+
+bool OpenXRFbSpatialEntityExtensionWrapper::is_component_supported(const XrSpace& space, XrSpaceComponentTypeFB type) {
+	uint32_t numComponents = 0;
+	xrEnumerateSpaceSupportedComponentsFB(space, 0, &numComponents, nullptr);
+	std::vector<XrSpaceComponentTypeFB> components(numComponents);
```
m4gr3d, line 116: Another usage of `std::vector`

godotopenxrmeta/src/main/cpp/include/openxr_fb_spatial_entity_query_extension_wrapper.h
```diff
@@ -0,0 +1,101 @@
...
+			(XrAsyncRequestIdFB), requestId,
+			(XrSpaceQueryResultsFB *), results)
+
+	bool initialize_fb_spatial_entity_query_extension(const XrInstance& instance);
+	void on_space_query_results(const XrEventDataSpaceQueryResultsAvailableFB* event);
+	void on_space_query_complete(const XrEventDataSpaceQueryCompleteFB* event);
+
+	std::map<XrAsyncRequestIdFB, Vector<XrSpaceQueryResultFB>> query_results;
+	std::map<String, bool *> request_extensions;
+	std::map<XrAsyncRequestIdFB, SpaceQueryCompleteCallback_t> query_complete_callbacks;
```
m4gr3d, line 94: You should be able to use `godot::HashMap` here instead

m4gr3d (approved):
Looks good overall!

Just some minor nits regarding types to address before merging.

maunvz:
addressed final type nits, added missing SCENE_MESH semantic label

m4gr3d:
Thanks for the contribution!

PR #68, opened 2023-12-22
Add additional export options for Meta Quest features

maunvz:
A lot of Meta Quest features are gated by AndroidManifest flags. Adding export options to support:
Experimental Features: https://developer.oculus.com/experimental/experimental-overview/
Overlay Keyboard: https://developer.oculus.com/documentation/unity/unity-keyboard-overlay/
Scene API: https://developer.oculus.com/documentation/native/native-spatial-data-perm/
Boundary Modes (can't find documentation, but already used by a few apps on the Quest Store)

### DIFF
--- a/godotopenxrmeta/src/main/cpp/export/meta_export_plugin.cpp
+++ b/godotopenxrmeta/src/main/cpp/export/meta_export_plugin.cpp
@@ -104,6 +104,46 @@ MetaEditorExportPlugin::MetaEditorExportPlugin() {
 			false,
 			false
 	);
+	_use_scene_api_option = _generate_export_option(
+			"meta_xr_features/use_scene_api",
+			"",
+			Variant::Type::BOOL,
+			PROPERTY_HINT_NONE,
+			"",
+			PROPERTY_USAGE_DEFAULT,
+			false,
+			false
+	);
+	_use_overlay_keyboard_option = _generate_export_option(
+			"meta_xr_features/use_overlay_keyboard",
+			"",
+			Variant::Type::BOOL,
+			PROPERTY_HINT_NONE,
+			"",
+			PROPERTY_USAGE_DEFAULT,
+			false,
+			false
+	);
+	_use_experimental_features_option = _generate_export_option(
+			"meta_xr_features/use_experimental_features",
+			"",
+			Variant::Type::BOOL,
+			PROPERTY_HINT_NONE,
+			"",
+			PROPERTY_USAGE_DEFAULT,
+			false,
+			false
+	);
+	_boundary_mode_option = _generate_export_option(
+			"meta_xr_features/boundary_mode",
+			"",
+			Variant::Type::INT,
+			PROPERTY_HINT_ENUM,
+			"Enabled,Disabled,Contextual",
+			PROPERTY_USAGE_DEFAULT,
+			BOUNDARY_ENABLED_VALUE,
+			false
+	);
 	_support_quest_1_option = _generate_export_option(
 			"meta_xr_features/quest_1_support",
 			"",
@@ -160,6 +200,10 @@ TypedArray<Dictionary> MetaEditorExportPlugin::_get_export_options(const Ref<Edi
 	export_options.append(_hand_tracking_frequency_option);
 	export_options.append(_passthrough_option);
 	export_options.append(_use_anchor_api_option);
+	export_options.append(_use_scene_api_option);
+	export_options.append(_use_overlay_keyboard_option);
+	export_options.append(_use_experimental_features_option);
+	export_options.append(_boundary_mode_option);
 	export_options.append(_support_quest_1_option);
 	export_options.append(_support_quest_2_option);
 	export_options.append(_support_quest_3_option);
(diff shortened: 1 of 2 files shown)


godotopenxrmeta/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -287,6 +343,32 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
 		contents += "    <uses-permission android:name=\"com.oculus.permission.USE_ANCHOR_API\" />\n";
 	}
 
+	// Check for scene api
+	bool use_scene_api = _get_bool_option("meta_xr_features/use_scene_api");
+	if (use_scene_api) {
+		contents += "    <uses-permission android:name=\"com.oculus.permission.USE_SCENE\" />\n";
```
m4gr3d, line 349: This should be moved within the https://github.com/GodotVR/godot_openxr_vendors/pull/66 PR so we can keep all related changes together.

Also do we need [additional logic to request the permission at runtime](https://developer.oculus.com/documentation/native/native-spatial-data-perm/#how-to-request-permission)?

m4gr3d (changes requested):

godotopenxrmeta/src/main/cpp/export/meta_export_plugin.cpp
maunvz: When testing, I see the permission dialog automatically appears at launch. Not quite sure why though...

1. Does Godot automatically request all dangerous permissions an app requires at launch, or is that the app's responsibility (i.e. in context permissions)? Ideally we do the second and add an API to request via the plugin, if Godot isn't already handling this.

2. On Quest 2 and earlier this is declared as a "normal" permission, on Quest Pro and later it's a "dangerous" permission. There's some compat mechanism that either auto-grants or auto-requests for older apps. This could be kicking in if it's not #1, and if so I'd need to figure out how the system determines that the compat mechanism is needed for a given app and how to opt out so we can just request in context.

Will look through engine / plugin source for answers to #1 unless you know off the top of your head, and follow up on #2 when I'm back at the office

m4gr3d: For (1), yes dangerous permissions are automatically requested on start.
See https://github.com/godotengine/godot/blob/master/platform%2Fandroid%2Fjava%2Flib%2Fsrc%2Forg%2Fgodotengine%2Fgodot%2FGodotActivity.kt#L68

maunvz: Awesome, in that case I'll move just the Scene permission to #66 to keep all that together and call it done since the permission is already requested appropriately, and address remaining comments here.

-------

Separate thought related to permissions: I see from the code pointer that "request at launch" isn't the only option:
>		// We exclude certain permissions from the set we request at startup, as they'll be
>		// requested on demand based on use-cases.

I see this list is hardcoded to just the microphone permission. It may be a good idea to define a "request" mode for all dangerous permissions that the developer can pick from:
- request at launch (when the game launches)
- request on demand (upon first access, if we can detect this automatically)
- manual (when the game chooses to in its game logic)

Manual would be used for "contextual" permissions, where you tell the user "hey we're gonna request a permission and use your data for XYZ, please accept the following prompt" and then request (increasingly common in Android apps, generally recommended for user education).

m4gr3d: Yeah, that'd be a good improvement to the current logic. I'd imagine the `request on demand` case would primarily be used by plugins / extensions so they can trigger the permission request when part of their functionality is used.

The engine does provide the ability to request permissions manually from the game logic, but that's moot because of the `request all` logic on start, so that'd need to be updated accordingly.

godotopenxrmeta/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -104,6 +104,46 @@ MetaEditorExportPlugin::MetaEditorExportPlugin() {
...
+	_use_scene_api_option = _generate_export_option(
+			"meta_xr_features/use_scene_api",
+			"",
+			Variant::Type::BOOL,
+			PROPERTY_HINT_NONE,
+			"",
+			PROPERTY_USAGE_DEFAULT,
+			false,
+			false
+	);
```
m4gr3d, line 116: This can be removed since it's being added in https://github.com/GodotVR/godot_openxr_vendors/pull/66

godotopenxrmeta/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -287,6 +343,32 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
...
+	// Check for scene api
+	bool use_scene_api = _get_bool_option("meta_xr_features/use_scene_api");
+	if (use_scene_api) {
+		contents += "    <uses-permission android:name=\"com.oculus.permission.USE_SCENE\" />\n";
+	}
+
+	// Check for overlay keyboard
+	bool _use_overlay_keyboard_option = _get_bool_option("meta_xr_features/use_overlay_keyboard");
+	if (_use_overlay_keyboard_option) {
+		contents += "    <uses-feature android:name=\"oculus.software.overlay_keyboard\" android:required=\"true\" />\n";
```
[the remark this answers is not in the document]
decacis: Just to share my experience using this feature, I use the following feature tag (hard-coded in my AndroidManifest.xml):

`<uses-feature android:name="oculus.software.overlay_keyboard" android:required="false"/>`

I can confirm that it's shown automatically when using Godot's text edit nodes.

godotopenxrmeta/src/main/cpp/export/meta_export_plugin.cpp
m4gr3d: fyi @maunvz, I'm reverting the automatic permissions requests behavior in https://github.com/godotengine/godot/pull/87080. So the PR will need to be updated so that the addon requests the necessary permissions when they're included in the manifest.

See https://github.com/GodotVR/godot_openxr_vendors/pull/71 for reference.

m4gr3d:
Superseded by https://github.com/GodotVR/godot_openxr_vendors/pull/82

PR #73, opened 2024-01-12
Register singletons the usual way

dsnopek:
Registers the singleton the usual way that singletons are registered in GDExtension, via `Engine::get_singleton()->register_singleton()`

Supersedes https://github.com/GodotVR/godot_openxr_vendors/pull/72

### DIFF
--- a/demo/main.gd
+++ b/demo/main.gd
@@ -9,8 +9,8 @@ func _ready():
 	if xr_interface and xr_interface.is_initialized():
 		var vp: Viewport = get_viewport()
 		vp.use_xr = true
-		
-	scene_capture = OpenXRFbSceneCaptureExtensionWrapper.get_singleton()
+
+	scene_capture = Engine.get_singleton("OpenXRFbSceneCaptureExtensionWrapper")
 	scene_capture.connect("scene_capture_completed", _on_scene_capture_completed)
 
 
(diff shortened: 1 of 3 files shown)


demo/main.gd
```diff
@@ -1,24 +1,22 @@
...
 func _ready():
 	xr_interface = XRServer.find_interface("OpenXR")
 	if xr_interface and xr_interface.is_initialized():
 		var vp: Viewport = get_viewport()
 		vp.use_xr = true
-		
-	scene_capture = OpenXRFbSceneCaptureExtensionWrapper.get_singleton()
-	scene_capture.connect("scene_capture_completed", _on_scene_capture_completed)
+
+	OpenXRFbSceneCaptureExtensionWrapper.scene_capture_completed.connect(self._on_scene_capture_completed)
```
m4gr3d, line 12: Let's use `Engine.get_singleton("OpenXRFbSceneCaptureExtensionWrapper")` instead as the current approach causes syntax error until the editor is restarted a second time.

@dsnopek On the syntax error, is that a known issue, or should I be create a GDExtension issue for it?

m4gr3d (approved):

demo/main.gd
dsnopek: Ok!

It's a known GDExtension issue. But it should only happen the first time you open the Godot editor with that GDExtension - all subsequent times should be fine (a file in the `.godot/` directory gets generated). This will get fixed sooner or later :-)

dsnopek: Updated in my latest push

m4gr3d:
Thanks for the fix!

PR #76, opened 2024-01-23
Add OpenXR fb_render_model extension wrapper and OpenXRFBRenderModel node

devloglogan:
Adds an extension wrapper for the OpenXR [XR_FB_render_model](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_render_model) extension and corresponding node implementing its usage. This extension provides access to gltf models of controllers/keyboards at runtime.

The extension wrapper caches a list of valid paths to these models (accessible with function `get_valid_paths()` in the `OpenXRFBRenderModel` node ~~or the same function in the extension wrapper singleton~~). One of these paths can be assigned to the `render_model_path` property of the `OpenXRFBRenderModel` node, which will trigger that model to be loaded and accessible via the nodes `get_render_model()` function. The only other time the model is loaded is on startup via the `OpenXrInterface` `session_begun` signal.

At present (tested on Quest 2 and Quest 3), it seems only the controller models are able to be successfully loaded. The virtual keyboard requires the XR_META_virtual_keyboard extension to be enabled, but I believe the other two _should_ be available. Unsure if there's another dependency missing for these. The controller models have been added to the demo main scene.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - Upgrade Android, Gradle, Godot and Kotlin dependencies
 - Add XR_FB_face_tracking support
 - Update to OpenXR 1.0.34 headers
+- Add XR_FB_render_model extension wrapper and OpenXRFBRenderModel node
 
 ## 2.0.3
 - Migrate the export scripts from gdscript to C++ via gdextension
--- a/common/src/main/cpp/classes/openxr_fb_render_model.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_render_model.cpp
@@ -1,0 +1,138 @@
+/**************************************************************************/
+/*  openxr_fb_render_model.cpp                                            */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "classes/openxr_fb_render_model.h"
+
+#include "extensions/openxr_fb_render_model_extension_wrapper.h"
+
+#include <godot_cpp/classes/gltf_document.hpp>
+#include <godot_cpp/classes/gltf_state.hpp>
+#include <godot_cpp/classes/open_xr_interface.hpp>
+#include <godot_cpp/classes/xr_server.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+using namespace godot;
+
+void OpenXRFbRenderModel::set_render_model_type(Model p_model) {
+	render_model_type = p_model;
+	if (is_inside_tree() && OpenXRFbRenderModelExtensionWrapper::get_singleton()->is_openxr_session_active()) {
+		load_render_model();
+	}
+}
+
+OpenXRFbRenderModel::Model OpenXRFbRenderModel::get_render_model_type() {
+	return render_model_type;
+}
+
+bool OpenXRFbRenderModel::has_render_model_node() {
+	return render_model_node != nullptr;
+}
+
+void OpenXRFbRenderModel::load_render_model() {
+	if (render_model_node != nullptr) {
+		render_model_node->queue_free();
+		render_model_node = nullptr;
+	}
+
+	String render_model_path;
+	switch (render_model_type) {
+		case MODEL_CONTROLLER_LEFT: {
+			render_model_path = "/model_fb/controller/left";
+		} break;
+
+		case MODEL_CONTROLLER_RIGHT: {
+			render_model_path = "/model_fb/controller/right";
+		} break;
+
+		default: {
+			render_model_path = "";
+		}
+	}
+
+	if (render_model_path.is_empty()) {
+		return;
+	}
+
+	PackedByteArray render_model_buffer = OpenXRFbRenderModelExtensionWrapper::get_singleton()->get_buffer(render_model_path);
+	if (render_model_buffer.is_empty()) {
+		UtilityFunctions::print_verbose("Failed to load render model buffer from path [", render_model_path, "] in OpenXRFbRenderModel node");
+		return;
+	}
+
+	Ref<GLTFDocument> gltf_document;
+	gltf_document.instantiate();
+	Ref<GLTFState> gltf_state;
+	gltf_state.instantiate();
+
+	Error err = gltf_document->append_from_buffer(render_model_buffer, "", gltf_state);
+	if (err != OK) {
+		UtilityFunctions::print_verbose("Failed to instance render model in OpenXRFbRenderModel node");
+		return;
+	}
+
+	render_model_node = Object::cast_to<Node3D>(gltf_document->generate_scene(gltf_state));
+	if (render_model_node) {
+		add_child(render_model_node);
+		emit_signal("openxr_fb_render_model_loaded");
+	}
+}
+
+Node3D *OpenXRFbRenderModel::get_render_model_node() {
+	return render_model_node;
+}
+
+void OpenXRFbRenderModel::_notification(int p_what) {
+	switch (p_what) {
+		case NOTIFICATION_POSTINITIALIZE: {
+			Ref<OpenXRInterface> openxr_interface = XRServer::get_singleton()->find_interface("OpenXR");
+			if (openxr_interface.is_valid()) {
+				openxr_interface->connect("session_begun", callable_mp(this, &OpenXRFbRenderModel::load_render_model));
+			}
+		} break;
+		case NOTIFICATION_ENTER_TREE: {
+			if (OpenXRFbRenderModelExtensionWrapper::get_singleton()->is_openxr_session_active()) {
+				load_render_model();
+			}
+		} break;
+	}
+}
+
+void OpenXRFbRenderModel::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("has_render_model_node"), &OpenXRFbRenderModel::has_render_model_node);
+	ClassDB::bind_method(D_METHOD("get_render_model_node"), &OpenXRFbRenderModel::get_render_model_node);
+	ClassDB::bind_method(D_METHOD("set_render_model_type", "render_model_type"), &OpenXRFbRenderModel::set_render_model_type);
+	ClassDB::bind_method(D_METHOD("get_render_model_type"), &OpenXRFbRenderModel::get_render_model_type);
+
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "render_model_type", PROPERTY_HINT_ENUM, "Left Controller,Right Controller"), "set_render_model_type", "get_render_model_type");
+
+	BIND_ENUM_CONSTANT(MODEL_CONTROLLER_LEFT);
+	BIND_ENUM_CONSTANT(MODEL_CONTROLLER_RIGHT);
+
+	ADD_SIGNAL(MethodInfo("openxr_fb_render_model_loaded"));
+}
--- a/common/src/main/cpp/include/classes/openxr_fb_render_model.h
+++ b/common/src/main/cpp/include/classes/openxr_fb_render_model.h
@@ -1,0 +1,67 @@
+/**************************************************************************/
+/*  openxr_fb_render_model.h                                              */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef OPENXR_FB_RENDER_MODEL_H
+#define OPENXR_FB_RENDER_MODEL_H
+
+#include <godot_cpp/classes/node3d.hpp>
+
+namespace godot {
+class OpenXRFbRenderModel : public Node3D {
+	GDCLASS(OpenXRFbRenderModel, Node3D)
+
+public:
+	enum Model {
+		MODEL_CONTROLLER_LEFT,
+		MODEL_CONTROLLER_RIGHT,
+	};
+
+private:
+	Model render_model_type = MODEL_CONTROLLER_LEFT;
+	Node3D *render_model_node = nullptr;
+
+	void load_render_model();
+
+protected:
+	void _notification(int p_what);
+
+	static void _bind_methods();
+
+public:
+	void set_render_model_type(Model p_model);
+	Model get_render_model_type();
+
+	bool has_render_model_node();
+	Node3D *get_render_model_node();
+};
+} //namespace godot
+
+VARIANT_ENUM_CAST(OpenXRFbRenderModel::Model);
+
+#endif
--- a/demo/main.gd
+++ b/demo/main.gd
@@ -1,5 +1,8 @@
 extends Node3D
 
+@onready var left_hand_mesh: MeshInstance3D = $XROrigin3D/LeftHand/LeftHandMesh
+@onready var right_hand_mesh: MeshInstance3D = $XROrigin3D/RightHand/RightHandMesh
+
 var xr_interface : XRInterface = null
 
 # Called when the node enters the scene tree for the first time.
@@ -23,3 +26,12 @@ func _on_left_hand_button_pressed(name):
 	if name == "menu_button" and scene_capture:
 		print("Triggering scene capture")
 		scene_capture.request_scene_capture()
+
+
+func _on_left_controller_fb_render_model_render_model_loaded() -> void:
+	left_hand_mesh.hide()
+
+
+func _on_right_controller_fb_render_model_render_model_loaded() -> void:
+	right_hand_mesh.hide()
+
--- a/demo/main.tscn
+++ b/demo/main.tscn
@@ -60,17 +60,19 @@ transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.352791, 0)
 [node name="LeftHand" type="XRController3D" parent="XROrigin3D"]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, -0.460909, 0.388594, -0.241118)
 tracker = &"left_hand"
-pose = &"aim"
+pose = &"grip"
 
 [node name="LeftHandMesh" type="MeshInstance3D" parent="XROrigin3D/LeftHand"]
 mesh = SubResource("BoxMesh_3kt6b")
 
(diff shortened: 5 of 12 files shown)


m4gr3d (changes requested):
Looks great! I've added some minor feedback to address.

demo/main.gd
```diff
@@ -23,3 +30,16 @@ func _on_left_hand_button_pressed(name):
 	if name == "menu_button" and scene_capture:
 		print("Triggering scene capture")
 		scene_capture.request_scene_capture()
+
+
+func _on_left_controller_fb_render_model_render_model_loaded() -> void:
+	left_hand.pose = "grip"
+	left_hand_mesh.rotate_x(-TAU/4)
```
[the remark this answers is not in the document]
devloglogan: It re-positions the hand tablet child node to the correct spot for the grip pose.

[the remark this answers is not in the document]
dsnopek: I think we could just make these changes permanently to the scene (except the visibility change, which needs to be in code)

godotopenxrmeta/src/main/cpp/register_types.cpp
```diff
@@ -72,11 +77,14 @@ void initialize_plugin_module(ModuleInitializationLevel p_level)
...
 
 		case MODULE_INITIALIZATION_LEVEL_SCENE: {
+			Engine::get_singleton()->register_singleton("OpenXRFbRenderModelExtensionWrapper", OpenXRFbRenderModelExtensionWrapper::get_singleton());
 			Engine::get_singleton()->register_singleton("OpenXRFbSceneCaptureExtensionWrapper", OpenXRFbSceneCaptureExtensionWrapper::get_singleton());
 			Engine::get_singleton()->register_singleton("OpenXRFbSpatialEntityExtensionWrapper", OpenXRFbSpatialEntityExtensionWrapper::get_singleton());
 			Engine::get_singleton()->register_singleton("OpenXRFbSpatialEntityQueryExtensionWrapper", OpenXRFbSpatialEntityQueryExtensionWrapper::get_singleton());
 			Engine::get_singleton()->register_singleton("OpenXRFbSpatialEntityContainerExtensionWrapper", OpenXRFbSpatialEntityContainerExtensionWrapper::get_singleton());
 			Engine::get_singleton()->register_singleton("OpenXRFbSceneExtensionWrapper", OpenXRFbSceneExtensionWrapper::get_singleton());
+
+			ClassDB::register_class<OpenXRFbRenderModel>();
```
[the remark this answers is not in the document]
dsnopek: Because it's a `Node`, I think it needs to be registered in `SCENE`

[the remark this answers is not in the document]
devloglogan: Initially that's what I tried, but the node isn't available in the editor if registered there. @dsnopek informed me that the scene level is the correct choice here.

godotopenxrmeta/src/main/cpp/openxr_fb_render_model_extension_wrapper.cpp
```diff
@@ -0,0 +1,243 @@
...
+}
+
+OpenXRFbRenderModelExtensionWrapper::~OpenXRFbRenderModelExtensionWrapper() {
+	cleanup();
+}
+
+void OpenXRFbRenderModelExtensionWrapper::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("is_enabled"), &OpenXRFbRenderModelExtensionWrapper::is_enabled);
+	ClassDB::bind_method(D_METHOD("get_valid_paths"), &OpenXRFbRenderModelExtensionWrapper::get_valid_paths);
+	ClassDB::bind_method(D_METHOD("get_buffer", "path"), &OpenXRFbRenderModelExtensionWrapper::get_buffer);
```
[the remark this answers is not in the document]
devloglogan: Whoops, definitely meant to remove `get_buffer()` from here. And it does feel a bit redundant to have `get_valid_paths()` available in both. I'll remove these.

godotopenxrmeta/src/main/cpp/openxr_fb_render_model_extension_wrapper.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+bool OpenXRFbRenderModelExtensionWrapper::is_enabled() const {
+	return fb_render_model_ext && system_render_model_properties.supportsRenderModelLoading;
+}
+
+void OpenXRFbRenderModelExtensionWrapper::populate_valid_paths() {
+	XrResult result;
+	uint32_t path_count = 0;
+
+	result = xrEnumerateRenderModelPathsFB((XrSession)get_openxr_api()->get_session(), 0, &path_count, nullptr);
```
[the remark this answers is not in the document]
devloglogan: Good to know, thank you!

godotopenxrmeta/src/main/cpp/openxr_fb_render_model.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+void OpenXRFbRenderModel::_bind_methods() {
+	ClassDB::bind_static_method("OpenXRFbRenderModel", D_METHOD("get_valid_paths"), &OpenXRFbRenderModel::get_valid_paths);
+	ClassDB::bind_method(D_METHOD("has_render_model"), &OpenXRFbRenderModel::has_render_model);
+	ClassDB::bind_method(D_METHOD("get_render_model"), &OpenXRFbRenderModel::get_render_model);
+	ClassDB::bind_method(D_METHOD("set_render_model_path", "render_model_path"), &OpenXRFbRenderModel::set_render_model_path);
+	ClassDB::bind_method(D_METHOD("get_render_model_path"), &OpenXRFbRenderModel::get_render_model_path);
+
+	ADD_PROPERTY(PropertyInfo(Variant::STRING, "render_model_path"), "set_render_model_path", "get_render_model_path");
```
Malcolmnixon, line 118: Is it possible to help the user enter a valid model path here. A similar issue arises with XRNode3D where there's a pose property string, but the list of poses cannot be statically entered (as it comes from the OpenXR Action Map) and so must be queried at runtime.

XRNode3D implements this by adding the `PROPERTY_HINT_ENUM_SUGGESTION` option as:
https://github.com/godotengine/godot/blob/74c32faa78b54863f8f25c538083907c2bf71791/scene/3d/xr_nodes.cpp#L230
This changes the node-editor to show a drop-down, but still lets the developer click a pen icon and type in a custom string should they choose to.

Then it implements the `_validate_property` method and uses it to populate the hint-strings from the `xr_server->get_suggested_pose_names`:
https://github.com/godotengine/godot/blob/74c32faa78b54863f8f25c538083907c2bf71791/scene/3d/xr_nodes.cpp#L240-L259
You should be able to use the `get_valid_paths` method to populate valid paths.

Malcolmnixon (approved):
This works well for me!

godotopenxrmeta/src/main/cpp/openxr_fb_render_model.cpp
dsnopek: Oh, that's a great idea! I actually didn't realize you could do a dropdown as a suggestion.

However, I don't think using `get_valid_paths()` would work - it'll only return results when we have a running OpenXR session, and that'll only be true in the XR editor. But we could certainly provide all the paths in the spec as suggestions!

BastiaanOlij: Worse yet, even if we get it to run in the editor it won't help you, because you'll get a list of valid paths on the PC. On Windows you might get lucky if you've got the Oculus Client configured as OpenXR, but you're dead in the water on Mac.

This is why in the action map our meta data ignores extensions, we can configure the action map regardless of the capabilities of the system on which the user is developing their game. What is important is what the capabilities of the headset are the user is running this on.

BastiaanOlij:
Looks really good at an early glance.

I'm concerned about the requirement of the rotation of the grip pose to position the render model, that should match up or the spec should provide information about correct placement. If it doesn't that's going to be bad and bite us down the road if other vendors follow suit but do properly position the meshes.

On that subject, exposing the render models to the end user should take a more global approach and not be FB centric. Here I'm unfortunately not at liberty to provide more detail yet, but we're going to back ourselves into a corner if we do not take a wider view.

devloglogan:
> I'm concerned about the requirement of the rotation of the grip pose to position the render model

To be clear, the render model isn't being repositioned by anything other than the setting of the grip pose. The hacky rotation is only there for repositioning the hand tablet correctly from one pose to the other. Would this not be expected? We could just set the pose for the controllers to be grip by default and reposition the hand tablet in the editor instead.

dsnopek:
> We could just set the pose for the controllers to be grip by default and reposition the hand tablet in the editor instead.

Yes, I think we should do exactly this!

devloglogan:
Alright, I changed the default controller poses to grip and added some (hard coded) `render_model_path` suggestions

dsnopek:
> On that subject, exposing the render models to the user should take a more global approach and not be FB centric. Here I'm unfortunately not at liberty to provide more detail yet, but we're going to back ourselves into a corner if we do not take a wider view.

If a vendor neutral OpenXR extension is added to the spec, we'd implement that in Godot itself'. So, personally, I don't think having the Meta-specific one here in 'godot_openxr_vendors' paints us into any corners. Folks not specifically targeting Meta headsets won't use their vendor extensions (they'll wait for the vendor neutral one), and Meta's extension will keep working on Meta devices even after the vendor neutral one exists.

godotopenxrmeta/src/main/cpp/openxr_fb_render_model.cpp
```diff
@@ -0,0 +1,141 @@
...
+TypedArray<String> OpenXRFbRenderModel::get_valid_paths() {
+	return OpenXRFbRenderModelExtensionWrapper::get_singleton()->get_valid_paths();
+}
+
+PackedStringArray OpenXRFbRenderModel::get_suggested_render_model_path_names() const {
+	PackedStringArray arr;
+	arr.push_back("/model_fb/controller/left");
+	arr.push_back("/model_fb/controller/right");
+	arr.push_back("/model_fb/keyboard/local");
+	arr.push_back("/model_fb/keyboard/remote");
```
m4gr3d, line 63: We should omit the keyboard paths until we add support for the keyboard extension. Maybe add a todo pointing to an issue to track.

m4gr3d (approved):
One minor comment to address but otherwise it looks good!

demo/main.gd
```diff
@@ -23,3 +30,13 @@ func _on_left_hand_button_pressed(name):
 	if name == "menu_button" and scene_capture:
 		print("Triggering scene capture")
 		scene_capture.request_scene_capture()
+
+
+func _on_left_controller_fb_render_model_render_model_loaded() -> void:
+	left_hand_mesh.hide()
+	hand_tablet.reparent(left_controller_render_model)
```
dsnopek, line 37: I think this should just be part of the scene too. There is no reason for the tablet to be a child of what it's a child of currently, so let's just move it in the editor permanently. Otherwise, folks may look at this demo thinking there's some reason for doing this, and copy it into their own projects :-)

demo/main.tscn
```diff
@@ -56,17 +56,20 @@ transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0.352791, 0)
...
+[node name="LeftControllerFbRenderModel" type="OpenXRFbRenderModel" parent="XROrigin3D/LeftHand"]
+render_model_path = "/model_fb/controller/left"
+
+[node name="HandTablet" type="MeshInstance3D" parent="XROrigin3D/LeftHand/LeftControllerFbRenderModel"]
+transform = Transform3D(1, 0, 0, 0, -0.392209, 0.919876, 0, -0.919876, -0.392209, 0, -0.236189, -0.208883)
 mesh = SubResource("QuadMesh_1oamj")
 surface_material_override/0 = SubResource("StandardMaterial3D_pmc5p")
 
-[node name="Viewport2Din3D" parent="XROrigin3D/LeftHand/LeftHandMesh/HandTablet" instance=ExtResource("2_7whgo")]
+[node name="Viewport2Din3D" parent="XROrigin3D/LeftHand/LeftControllerFbRenderModel/HandTablet" instance=ExtResource("2_7whgo")]
```
dsnopek, line 72: Why not just have this as a child of `LeftHand`? Ie

```suggestion
[node name="Viewport2Din3D" parent="XROrigin3D/LeftHand/HandTablet" instance=ExtResource("2_7whgo")]
```

devloglogan: Good point, moved it!

BastiaanOlij:
Just an intermediate update from me as I'm waiting on hearing back before we can make an informed decision on how to proceed here. Until I can share updates, best to press pause, hope to have info in the next couple days.

BastiaanOlij:
> If a vendor neutral OpenXR extension is added to the spec, we'd implement that in Godot itself. So, personally, I don't think having the Meta-specific one here in 'godot_openxr_vendors' paints us into any corners. Folks not specifically targeting Meta headsets won't use their vendor extensions (they'll wait for the vendor neutral one), and Meta's extension will keep working on Meta devices even after the vendor neutral one exists.

It's not that simple. The following scenario is fairly likely:
- we implement the meta extension as is
- in the near future we implement the official solution but there is no guarantee Meta will immediately support it as they already have a solution
- our users need to build something that uses the Meta solution on Quest, and the official solution on everything else, the most likely scenario is simply checking which one is available.
- Meta adds supports for the official solution but keeps the Meta solution around for backwards compatibility, both are now reported as available
- Existing Godot apps potentially will break as they try to activate both Meta and official solutions at the same time.

So we have a mandate to at least discuss whether we can come up with a solution that staves off this scenario (which is very likely due to reasons I'm not at liberty to share), or whether we should forge ahead and deal with the fallout later.

godotopenxrmeta/src/main/cpp/include/openxr_fb_render_model.h
```diff
@@ -0,0 +1,69 @@
...
+	void _notification(int p_what);
+
+	static void _bind_methods();
+
+	void _validate_property(PropertyInfo &p_property) const;
+
+public:
+	PackedStringArray get_suggested_render_model_path_names() const;
+
+	void set_render_model_path(String p_render_model_path);
```
m4gr3d, line 59: Let's remove this method. Through the api users should only have access to what we know is supported by the runtime.

dsnopek: What do you mean by this? The user needs some way to select which model should be loaded. Do you mean switching from a `String` to an `enum` so they can only use certain values, rather than a free-form string?

devloglogan:
@BastiaanOlij rebased!

godotopenxrmeta/src/main/cpp/include/openxr_fb_render_model.h
m4gr3d: Yeah I was referring to remove users' ability to enter a path themselves, so switching to an `enum` would work.

m4gr3d:
@BastiaanOlij @dsnopek From our discussions, it's likely we'll move forward with this approach as a `v0` api. Can you review in that context.

@devloglogan Can you rebase the PR for review.

devloglogan:
Rebased! Per Fredia's request I've swapped the model selector on the `OpenXRFbRenderModel` node to be an `enum` type. After doing that, it seemed pointless storing path names and exposing them, so I removed that functionality.

common/src/main/cpp/extensions/openxr_fb_render_model.cpp
```diff
@@ -0,0 +1,131 @@
+/**************************************************************************/
+/*  openxr_fb_render_model.cpp                                            */
```
m4gr3d, line 2: Can you move this implementation into a `classes` directory: `common/src/main/cpp/classes`. 

This will enable us to distinguish between extension wrappers implementations in the `extensions` directory and nodes / classes used to expose the functionality to end users.

common/src/main/cpp/extensions/openxr_fb_render_model.cpp
```diff
@@ -0,0 +1,131 @@
...
+	ClassDB::bind_method(D_METHOD("get_render_model"), &OpenXRFbRenderModel::get_render_model);
+	ClassDB::bind_method(D_METHOD("set_render_model_name", "render_model_name"), &OpenXRFbRenderModel::set_render_model_name);
+	ClassDB::bind_method(D_METHOD("get_render_model_name"), &OpenXRFbRenderModel::get_render_model_name);
+
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "render_model_name", PROPERTY_HINT_ENUM, "Left Controller,Right Controller"), "set_render_model_name", "get_render_model_name");
+
+	BIND_ENUM_CONSTANT(MODEL_CONTROLLER_LEFT);
+	BIND_ENUM_CONSTANT(MODEL_CONTROLLER_RIGHT);
+
+	ADD_SIGNAL(MethodInfo("render_model_loaded"));
```
m4gr3d, line 130: Let's make the signal name more explicit and use `openxr_fb_render_model_loaded`.

common/src/main/cpp/include/extensions/openxr_fb_render_model.h
```diff
@@ -0,0 +1,69 @@
+/**************************************************************************/
+/*  openxr_fb_render_model.h                                              */
```
m4gr3d, line 2: Similar comment as for `openxr_fb_render_model.cpp`; let's move this header under a new `include/classes` subdirectory: `common/src/main/cpp/include/classes`

m4gr3d (changes requested):

devloglogan:
@m4gr3d requests added!

dsnopek (approved):
This is looking great to me!

I wonder if the "classes" directory should really be "scene" because this is a node and that's what Godot names the directory that holds nodes? If we end up with some `RefCounted` classes it may be nice to have those in a different directory. But what we have here is certainly fine for now :-)

m4gr3d (approved):

CHANGES.md
```diff
@@ -4,6 +4,7 @@
 - Merge GDExtension into a single implementation
 - Upgrade Android, Gradle, Godot and Kotlin dependencies
 - Add XR_FB_face_tracking support
+- Add XR_FB_render_model extension warpper and OpenXRFBRenderModel node
```
m4gr3d, line 7: Typo in `wrapper`

devloglogan: Ahg, my bad, thank you!

BastiaanOlij (changes requested):
Ok Logan, just starting. Getting everything to compile and to a point where I thought I could run things took a lot longer than expected, and I ran into a number of snags I'm still looking into. Part of it was me getting familiar with things again (haven't had a deep dive into the vendors plugin), part is that we may have lost some instructions or that a clean build is throwing up some issues.

So there is more to come. So far it looks good, I'm mostly worried about some flow things around the model loading at the wrong time and in the wrong scenario.

Also my apologies that it took so long before I could really sink my teeth into. There are things happening on the core of OpenXR where doing something a little more elaborate to make this a bit more future proof may have been a necessary thing.

But looking at some structural differences and the fact that things will take longer than I hoped, I'm now of the belief that this is too soon and that an FB specific implementation as you've done here, is the right course of action. I also do not believe it likely that other vendors will come out with their own extension which would require us to think this through a little more.

That will mean that in the near future users who implement this extension and are planning to release games beyond the Meta platform, will likely need to refactor their code. We will deal with that, when the time is right.

common/src/main/cpp/classes/openxr_fb_render_model.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <godot_cpp/classes/gltf_state.hpp>
+#include <godot_cpp/classes/open_xr_interface.hpp>
+#include <godot_cpp/classes/xr_server.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+using namespace godot;
+
+void OpenXRFbRenderModel::set_render_model_name(Model p_model) {
+	render_model_name = p_model;
+	load_render_model();
```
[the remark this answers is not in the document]
devloglogan: I've made some changes, hope they will suffice! There's three scenarios at which a model can now be loaded.

1. The `session_begun` signal, for the `OpenXRFbRenderModel` nodes that are added to the scene tree in the editor.

2. The `NOTIFICATION_ENTER_TREE` notification, which checks to see if the OpenXR session is active (I'm now storing that in the extension wrapper). For the `OpenXRFbRenderModel` nodes that are added at runtime.

3. The property setting route, which now checks that it's in the scene tree and that the session is active. For changing the model type at runtime.

common/src/main/cpp/include/classes/openxr_fb_render_model.h
```diff
@@ -0,0 +1,69 @@
...
+	GDCLASS(OpenXRFbRenderModel, Node3D)
+
+public:
+	enum Model {
+		MODEL_CONTROLLER_LEFT,
+		MODEL_CONTROLLER_RIGHT,
+	};
+
+private:
+	Model render_model_name = MODEL_CONTROLLER_LEFT;
```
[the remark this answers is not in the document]
devloglogan: Renamed them. Personally, I prefer the more verbose names as just `type` and `node` feel a bit ambiguous to me. But if you feel strongly on dropping the prefix just let me know!

common/src/main/cpp/classes/openxr_fb_render_model.cpp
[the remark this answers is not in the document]
BastiaanOlij: Looks good to me :)

m4gr3d:
@BastiaanOlij Looks like all your feedback on the PR have been addressed. Are you good with merging the current implementation?

BastiaanOlij:
@m4gr3d I haven't gotten anywhere near actually testing the latest version, as I was discussing with David the other day I am getting fairly concerned with this, especially as the testing I did get around doing last week show that a few things have snuck in that are causing issues and right now the plugin is in a broken state for me.

Purely based on the code and the discussion around this however, I am happy. But I would like to actually take it for a drive before giving my final thumbs up.

common/src/main/cpp/classes/openxr_fb_render_model.cpp
```diff
@@ -0,0 +1,138 @@
...
+	ClassDB::bind_method(D_METHOD("get_render_model_node"), &OpenXRFbRenderModel::get_render_model_node);
+	ClassDB::bind_method(D_METHOD("set_render_model_type", "render_model_type"), &OpenXRFbRenderModel::set_render_model_type);
+	ClassDB::bind_method(D_METHOD("get_render_model_type"), &OpenXRFbRenderModel::get_render_model_type);
+
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "render_model_type", PROPERTY_HINT_ENUM, "Left Controller,Right Controller"), "set_render_model_type", "get_render_model_type");
+
+	BIND_ENUM_CONSTANT(MODEL_CONTROLLER_LEFT);
+	BIND_ENUM_CONSTANT(MODEL_CONTROLLER_RIGHT);
+
+	ADD_SIGNAL(MethodInfo("openxr_fb_render_model_loaded"));
```
BastiaanOlij, line 137: I'm not sure about the best implementation but the one thing I was missing is control/signals related to the controller no longer being tracked.

Right now in the demo you react to the `openxr_fb_render_model_loaded` signal, and hide the hand tracking but that seems to be too restrictive, and once we use proper hand tracking, it should actually show the hand properly holding the controller (and thus verifying the positioning is correct).

Off course this could be a limitation of the demo, and possibly the logic I'm looking for shouldn't be part of this but should be part of the general hand tracking implementation that is in the core of Godot.

If the latter, we should probably do a follow up PR that adds hand tracking into the demo with the hand meshes Malcolm has (I plan to do a version of this during the week for the `godot-demo-projects` repo).

But if there is more to handle here, especially if tracking of controllers stops and the controllers should be hidden, that is worth looking into before we merge this.

BastiaanOlij (approved):
OK, so thanks to some motivational help from Fredia I managed to get things working today. There are some issues unrelated to this PR but the big issues that got me stuck and ready to throw in the proverbial towel, seemed to have been solved.

The only one that worries me but is likely again unrelated to this PR, is that the Godot IDE hung while I was looking around at the setup.

There is one question that should be down below for what happens when controllers are let go and some suggestions for improving the demo. I'm happy for this to be merged and that issue to be raised as a separate issue to be handled in a follow up PR.

common/src/main/cpp/classes/openxr_fb_render_model.cpp
devloglogan: My initial thought: this extension exists to provide 3D models, so our implementation should just expose those 3D models, and any further logic should be done by end users (or at least not here). It's possible users may want to drop static controller models in other areas, not just use them for the tracked controllers.

dsnopek: I agree with that. There's also more models than just controllers, this can also be used for physics keyboards (for the tracked keyboard feature) and the advanced virtual keyboard (which isn't tracking anything at all).

m4gr3d:
@devloglogan Can you rebase the PR so we can proceed to merge it!

devloglogan:
@m4gr3d rebased

PR #78, opened 2024-01-26
Preparing release 2.0.3

BastiaanOlij:
Updating version number and changes document in preparation of release 2.0.3.

Note #61 and #66 were both merged but marked as `2.1.0` PRs. I've moved them under 2.0.3 in the changes document. We'll need to update the milestones if this was correct.

If either does not belong in 2.0.3, they need to be reverted until we release 2.0.3, and then resubmitted.


m4gr3d (approved):
Looks good!

Issue #67, opened 2023-12-11
Editor crash when closed


GuillaumeCa:
After upgrading from 2.0.1 to 2.0.2 on godot 4.2, the editor crashes when it is closed.

```
Godot(48212,0x7ff860710640) malloc: *** error for object 0x7feaf700d588: pointer being freed was not allocated
Godot(48212,0x7ff860710640) malloc: *** set a breakpoint in malloc_error_break to debug

```

I'm using macOS 13.4 on an intel macbook pro

starcin:
It doesn't matter if it is enabled or if I have an export template using OpenXR. And I've tried it removing the folder and replacing it with the one still in my downloads folder (the original).

It doesn't crash for the first time. First time, it says the editor needs a restart for some extensions to work properly. In this session, exiting Godot is without a problem. After opening it again, now it crashes every time I exit.

nontoxicguy:
I have the same issue on Linux mint 21.3 Cinnamon. When executing Godot from the command line I can see an error, even the first time. Here it is:
```
ERROR: Condition "!EditorSettings::get_singleton() || !EditorSettings::get_singleton()->has_setting(p_setting)" is true. Returning: Variant()
   at: _EDITOR_GET (editor/editor_settings.cpp:1140)
```
I hate the crash logs created by this.

nontoxicguy:
The issue still happens on version 2.0.3. The error I mentioned above is no longer here. It was therefore, unrelated. Since #64 is just an update on the automatic way the add-on is built and it would only be possible to cause an issue only if the builds are made with this automated script which would be very surprising.

To finish this comment, another info about the issue: by deleting every vendor directory, it disappears, which sounds logic said like that but if we only leave one vendor the issue persists. So it is not specific to a vendor.

m4gr3d:
The root cause for this issue has been addressed in the Godot core codebase. The fix will roll-out with the Godot 4.3 and Godot 4.2.2 release.

nontoxicguy:
I tested if the "fix" works on both Godot 4.2.2rc2 and 4.3dev5 by installing 2.0.3 in brand new projects and closing the editor. The editor crashed and created a crash log on both versions.

Am I missing something?

If not, then I want to bring back the fact that 2.0.1 works but not 2.0.2 (as I remember from testing on Godot 4.2.1), I'm assuming this means that the faulty code isn't in the engine.

m4gr3d:
[...]

PR #80, opened 2024-01-28
Fix double free

nontoxicguy:
I looked at the stack trace with gdb after reproducing #67 to see a double free error: 
```
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=139885056115072)
    at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=139885056115072)
    at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=139885056115072, signo=signo@entry=6)
    at ./nptl/pthread_kill.c:89
#3  0x00007f3987187476 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/posix/raise.c:26
#4  0x00007f398716d885 in __GI_abort () at ./stdlib/abort.c:100
#5  0x00007f39871ce676 in __libc_message (action=action@entry=do_abort, 
    fmt=fmt@entry=0x7f3987320b77 "%s\n") at ../sysdeps/posix/libc_fatal.c:155
#6  0x00007f39871e5cfc in malloc_printerr (
    str=str@entry=0x7f39873236f0 "free(): double free detected in tcache 2")
    at ./malloc/malloc.c:5664
#7  0x00007f39871e80ab in _int_free (av=0x7f398735fc80 <main_arena>, 
    p=0x1faab050, have_lock=0) at ./malloc/malloc.c:4473
#8  0x00007f39871ea453 in __GI___libc_free (mem=<optimized out>)
    at ./malloc/malloc.c:3391
#9  0x00007f397f8126e2 in ?? ()
   from /path/to/game/addons/godotopenxrvendors/pico/.bin/libgodotopenxrpico.linux.template_debug.x86_64.so
```
I've noticed that remove_export_plugin [frees the export plugin](https://github.com/godotengine/godot/blob/4.2/editor/export/editor_export.cpp#L162) meaning that the memfree called after is throwing the error.

I can't test it, but I'm almost sure this fixes the error. I haven't worked with C++ yet, first time.

### DIFF
--- a/godotopenxrkhronos/src/main/cpp/export/khronos_export_plugin.cpp
+++ b/godotopenxrkhronos/src/main/cpp/export/khronos_export_plugin.cpp
@@ -35,7 +35,7 @@ void KhronosEditorPlugin::_bind_methods() {}
 
 void KhronosEditorPlugin::_enter_tree() {
 	// Initialize the editor export plugin
-	khronos_export_plugin = memnew(KhronosEditorExportPlugin);
+	khronos_export_plugin.instantiate();
 	add_export_plugin(khronos_export_plugin);
 }
 
@@ -43,8 +43,7 @@ void KhronosEditorPlugin::_exit_tree() {
 	// Clean up the editor export plugin
 	remove_export_plugin(khronos_export_plugin);
 
-	memfree(khronos_export_plugin);
-	khronos_export_plugin = nullptr;
+	khronos_export_plugin.unref();
 }
 
 KhronosEditorExportPlugin::KhronosEditorExportPlugin() {
--- a/godotopenxrkhronos/src/main/cpp/export/khronos_export_plugin.h
+++ b/godotopenxrkhronos/src/main/cpp/export/khronos_export_plugin.h
@@ -58,5 +58,5 @@ protected:
 	static void _bind_methods();
 
 private:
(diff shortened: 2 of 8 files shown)


BastiaanOlij:
Looking through the code, `remove_export_plugin` only removes the entry from the list of plugins, it should not free the object. 

The real issue here is that plugins should be handed through references, we shouldn't be using `memnew` and `memfree` to create these objects. The result thus is that on the removal of the plugin from the array, the refcount hits 0, and the object is freed. But that's a fluke, not something to rely on.

To properly fix this,  `OpenXREditorExportPlugin *pico_export_plugin = nullptr;` should become `Ref<OpenXREditorExportPlugin>`, and we should use `pico_export_plugin.instantiate()` and `pico_export_plugin.unref()` to instantiate and free these objects.

Similar changes should be applied to the other plugins.

nontoxicguy:
Expected due to my very limited knowledge of C++. Effectively the build still has the issue, making that pr permitted to test my changes. Thanks for your understanding, going to update this pr with your recommended changes.

m4gr3d (approved):
I tested the changes and unfortunately they don't resolve https://github.com/GodotVR/godot_openxr_vendors/issues/67. 

I'll go ahead and merge the PR however because it's an enhancement on our current logic.

PR #81, opened 2024-01-30
Implement XR_FB_hand_tracking_aim extension

devloglogan:
This PR depends on Godot PR [87546](https://github.com/godotengine/godot/pull/87546)

Adds extension wrapper for the [XR_FB_hand_tracking_aim](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_hand_tracking_aim) extension. The added singleton has methods to determine gesture states / pinch strength while hand tracking, and also provides a number of signals indicating when certain gestures begin/end.

Since this extension wrapper will only be supported by 4.3 onward, as well as the others listed in the above Godot PR, maybe it would make since to merge this into a 4.3 branch instead of master?

Also, one thing that might be worth discussing is how we should implement demos of these extension wrappers moving forward. Right now they're being dropped in `main.tscn`, but these are only meta extensions, so it doesn't really make sense to have them always present. Maybe we want to start thinking of implementing a menu to navigate implemented extensions and load corresponding demos?

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
@@ -1,0 +1,193 @@
+/**************************************************************************/
+/*  openxr_fb_hand_tracking_aim_extension_wrapper.cpp                     */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "extensions/openxr_fb_hand_tracking_aim_extension_wrapper.h"
+
+#include <godot_cpp/classes/project_settings.hpp>
+#include <godot_cpp/classes/xr_pose.hpp>
+
+using namespace godot;
+
+OpenXRFbHandTrackingAimExtensionWrapper *OpenXRFbHandTrackingAimExtensionWrapper::singleton = nullptr;
+
+OpenXRFbHandTrackingAimExtensionWrapper *OpenXRFbHandTrackingAimExtensionWrapper::get_singleton() {
+	if (singleton == nullptr) {
+		singleton = memnew(OpenXRFbHandTrackingAimExtensionWrapper());
+	}
+	return singleton;
+}
+
+OpenXRFbHandTrackingAimExtensionWrapper::OpenXRFbHandTrackingAimExtensionWrapper() :
+		OpenXRExtensionWrapperExtension() {
+	ERR_FAIL_COND_MSG(singleton != nullptr, "An OpenXRFbHandTrackingAimExtensionWrapper singleton already exists.");
+
+	request_extensions[XR_FB_HAND_TRACKING_AIM_EXTENSION_NAME] = &fb_hand_tracking_aim_ext;
+	singleton = this;
+}
+
+OpenXRFbHandTrackingAimExtensionWrapper::~OpenXRFbHandTrackingAimExtensionWrapper() {
+	cleanup();
+}
+
+void OpenXRFbHandTrackingAimExtensionWrapper::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("is_enabled"), &OpenXRFbHandTrackingAimExtensionWrapper::is_enabled);
+}
+
+void OpenXRFbHandTrackingAimExtensionWrapper::cleanup() {
+	XRServer *xr_server = XRServer::get_singleton();
+
+	for (int i = 0; i < Hand::HAND_MAX; i++) {
+		if (xr_server != nullptr) {
+			xr_server->remove_tracker(trackers[i]);
+		}
+		trackers[i].unref();
+	}
+
+	fb_hand_tracking_aim_ext = false;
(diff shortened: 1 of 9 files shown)


dsnopek:
Note: This will fail CI until the Godot PR is merged, and we're using godot-cpp with an updated `extension_api.json` for it

BastiaanOlij:
`XR_FB_hand_tracking_aim` has the dubious honour of being one of the FB extensions that isn't following the design of OpenXR very well.
It's basically being replaced by `XR_EXT_hand_interaction` (which Meta sadly isn't supporting yet but for which we have an open PR that can be merged once we have a runtime on which we can test it). Together with the additions to the touch profiles Meta recently did, that means that there is a clean switch in a player controlling thing the game with a controller, or switching over to handtracking, with the action map.

I'm not against support for hand tracking aim, but it should come with a big warning that it's an older solution that is not portable, and that soon better alternatives are possible through the hand interaction API.

For getting this PR over the line I'd like to see two changes:
1. The feature needs to be gated behind a project setting. People need to opt in to this extension be activated. Its very likely that in the long term, many people will leave it off
2. I wouldn't go for direct access to the extension. Instead I would prefer if the extension manages two new positional trackers so normal `XRController3D` nodes can be used but with hardcoded actions that bypass the actionmap (similar to what we do with Skeleton). So we simply create two new paths, say `/user/fb_hand_aim/left` and `/user/fb_hand_aim/right` that only has a default pause (which is the provided aim pose) and various actions like `pinch_index`, `pinch_middle`, etc.

BastiaanOlij:
Note that for point 2, we probably need to make an amendment to core so [OpenXRInterface::get_suggested_tracker_names](https://github.com/godotengine/godot/blob/master/modules/openxr/openxr_interface.cpp#L152) calls into extension wrappers to ask for additional entries to return. We can then just have the 3 OpenXR paths returned, move the HTC entries into the HTC extension, and have this extension return the two new trackers. This will ensure they are selectable from the dropdown.

Similarly, but this should be for later as its a much bigger job and a more structural change, `get_suggested_pose_names` should be implemented to actually query the action map for poses mapped in the action map and call into extensions to return hardcoded action names.

common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
```diff
@@ -0,0 +1,164 @@
...
+		Transform3D transform = Transform3D(quat, origin);
+		Vector3 linear_velocity = Vector3(0.0, 0.0, 0.0);
+		Vector3 angular_velocity = Vector3(0.0, 0.0, 0.0);
+		XRPose::TrackingConfidence confidence = XRPose::TrackingConfidence::XR_TRACKING_CONFIDENCE_LOW;
+
+		if (!(aim_state[i].status & XR_HAND_TRACKING_AIM_VALID_BIT_FB)) {
+			confidence = XRPose::TrackingConfidence::XR_TRACKING_CONFIDENCE_NONE;
+		} else if (aim_state[i].status & XR_HAND_TRACKING_AIM_COMPUTED_BIT_FB) {
+			confidence = XRPose::TrackingConfidence::XR_TRACKING_CONFIDENCE_HIGH;
+		}
```
devloglogan, line 149: Still more work to be done on updating this PR but I rebased and swapped over to positional trackers as suggested. Not sure if in this context there is a point to checking anything for tracking confidence, but for the moment I threw this in. I don't really understand what `XR_HAND_TRACKING_AIM_COMPUTED_BIT_FB` really means yet, from the spec on what the flag means:

> Aiming data is computed from additional sources beyond the hand data in the base structure

m4gr3d (changes requested):

m4gr3d:
> Since this extension wrapper will only be supported by 4.3 onward, as well as the others listed in the above Godot PR, maybe it would make since to merge this into a 4.3 branch instead of master?

We follow the approach in Godot core where `master` tracks the tip of the development branch, so once approved this change can be merged into the `master` branch.
We'll need to update the `godot-cpp` submodule accordingly.

devloglogan:
Updated and rebased! Also see godotengine/godot#88311 for changes to `get_suggested_tracker_names()`

common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
```diff
@@ -0,0 +1,188 @@
...
+		uint64_t value = reinterpret_cast<uint64_t>(ext.second);
+		result[key] = (godot::Variant)value;
+	}
+	return result;
+}
+
+PackedStringArray OpenXRFbHandTrackingAimExtensionWrapper::_get_suggested_tracker_names() {
+	PackedStringArray arr = PackedStringArray();
+	arr.push_back("/user/fbhandaim/left");
+	arr.push_back("/user/fbhandaim/right");
```
m4gr3d, line 88: Can you turn these two into constants since they're being reused below as well.

m4gr3d (approved):

common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
devloglogan: Done 👍

m4gr3d:
@devloglogan Can you bump the [`compatibility_minimum` value in `plugin.gdextension`](https://github.com/GodotVR/godot_openxr_vendors/blob/d51b5031293e35cfc007d4f913cb4430aaea2d38/demo/addons/godotopenxrvendors/plugin.gdextension#L4) to `4.3`.

m4gr3d:
@devloglogan Can you bump the [`compatibility_minimum` value in `plugin.gdextension`](https://github.com/GodotVR/godot_openxr_vendors/blob/d51b5031293e35cfc007d4f913cb4430aaea2d38/demo/addons/godotopenxrvendors/plugin.gdextension#L4) to `4.3`.

devloglogan:
Rebased and bumped the `compatibility_minimum` value.

common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
BastiaanOlij: I think that means that if full optical tracking isn't available (hand is not visible) that the current position is based on indirect measurements. I believe Quest for instance just added a way to determine hand position based on the cameras still seeing the arm.

So my gut feeling says that if `XR_HAND_TRACKING_AIM_COMPUTED_BIT_FB` is set, our confidence would be low, not high.

Could be wrong though and I don't think many users check beyond confidence not being `NONE`.

devloglogan: I'd discussed with @dsnopek and he thought the same. At the moment I have it setting the confidence always to `LOW`, unless the flag `XR_HAND_TRACKING_AIM_VALID_BIT_FB` is not set, in which case I set confidence to `NONE`. So I'm not actually using the `XR_HAND_TRACKING_AIM_COMPUTED_BIT_FB` bit anywhere now.

dsnopek (approved):
I just retested this, and it's working wonderfully!

The latest code is looking good to me as well.

Great work!

common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+uint64_t OpenXRFbHandTrackingAimExtensionWrapper::_set_hand_joint_locations_and_get_next_pointer(int32_t p_hand_index, void *p_next_pointer) {
+	ERR_FAIL_INDEX_V_MSG(p_hand_index, Hand::HAND_MAX, reinterpret_cast<uint64_t>(p_next_pointer), vformat("Invalid hand index %d", p_hand_index));
+
+	if (!fb_hand_tracking_aim_ext) {
+		return reinterpret_cast<uint64_t>(p_next_pointer);
+	}
+
+	aim_state[p_hand_index].type = XR_TYPE_HAND_TRACKING_AIM_STATE_FB;
```
m4gr3d, line 127: Let's switch this to brace initialization as discussed on https://github.com/GodotVR/godot_openxr_vendors/pull/100

m4gr3d (approved):

common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
devloglogan: Updated in latest push

PR #84, opened 2024-02-03
Add XR_FB_face_tracking support

Malcolmnixon:
This pull request adds support for the XR_FB_face_tracking extension to the Meta vendor plugin. Users can:
* Select face-tracking features of None, Optional, and Required
* Access the tracking data through the `OpenXRFbFaceTrackingExtensionWrapper` singleton
* Test if face-tracking is enabled by the `is_enabled()` method
* Get all 63x expression weights using the `get_weights()` method (indexed via enums)
* Get the 2x tracking confidences using the `get_confidences()` method (indexed via enums)

It includes a face model from https://booth.pm/en/items/5420804 which contains almost all the blend-shapes necessary to test the face-tracking feature; although the files are quite large for this.

https://github.com/GodotVR/godot_openxr_vendors/assets/1863707/7a37bb99-86f5-4136-9ba9-11e112feb1f0


dsnopek:
Wow, this looks so cool! I'll review in depth when I have a chance :-)

m4gr3d (approved):
The logic looks good!

dsnopek (approved):
I just tested this with the Quest Pro and it worked great! (Except the face model in the demo is a little creepy :-)) Skimming the code, it looks good to me as well

m4gr3d:
@Malcolmnixon Just to confirm, this only works on the Quest Pro, right?

dsnopek:
In my testing, it only works on the Quest Pro. I also tried on the Quest 3, but it does nothing there - it doesn't even show the permission overlay about face tracking

Malcolmnixon:
> @Malcolmnixon Just to confirm, this only works on the Quest Pro, right?

I believe only the Quest Pro has the face-tracking cameras necessary to do this.

m4gr3d:
@Malcolmnixon We should consider updating the implementation to use [XR_FB_face_tracking2](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_face_tracking2) instead.

Malcolmnixon:
> @Malcolmnixon We should consider updating the implementation to use [XR_FB_face_tracking2](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_face_tracking2) instead.

That will need us to update openxr to 1.0.34 first. I'll get a separate PR ready for that.

PR #87, opened 2024-02-06
Make OpenXR vendor extension headers available to all

maunvz:
OpenXR extensions are designed to be fully optional, and for one binary to run on many systems using two runtime mechanisms:
 - XrCreateInstance taking a list of requested (optional) extension names
 - xrGetInstanceProcAddr returning XR_ERROR_FUNCTION_UNSUPPORTED if the extension isn't there

As such it's safe to include the full collection of extension headers for all different versions of these libraries without preprocessor macros because the extension would fail to load at runtime and the plugin must handle that.

For now this means the exact same source can compile for all platforms. It would be nice to make it a single library but since we're linking against different loaders we can't quite do that yet (until everyone supports the khronos loader).


maunvz:
converting to draft given two issues:
- need to review licensing implications of Meta extension headers
- pre-processor macros may still be useful in cases of header collisions, will restore these at least in the cmake files

BastiaanOlij:
XR Meeting 23/8/24

We've gone into a different direction, closing this PR. Thanks for the effort!

PR #97, opened 2024-02-27
Add XR_FB_body_tracking support.

Malcolmnixon:
This pull request depends upon https://github.com/godotengine/godot/pull/88798 to provide the XRBodyTracker and XRBodyModifier3D types. Until they are available and merged in, building the code requires:

Building Godot with the XRBodyTracker and XRBodyModifier3D nodes.
Running Godot with --dump-extension-api an overwriting the godot_cpp extension_api.json file
Building godot_openxr_vendors

The following videos demonstrates the results:

https://github.com/GodotVR/godot_openxr_vendors/assets/1863707/3af85576-725c-47c3-9b7b-aa0c9d0c4dd3

https://github.com/GodotVR/godot_openxr_vendors/assets/1863707/bcca52be-8ecc-4913-960a-2dd6425fc35d


common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,338 @@
...
+	{ XRBodyTracker::JOINT_RIGHT_LOWER_ARM, XR_BODY_JOINT_RIGHT_ARM_LOWER_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+
+	// Left hand joints
+	{ XRBodyTracker::JOINT_LEFT_HAND, XR_BODY_JOINT_LEFT_HAND_WRIST_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_PALM, XR_BODY_JOINT_LEFT_HAND_PALM_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_WRIST, XR_BODY_JOINT_LEFT_HAND_WRIST_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_THUMB_METACARPAL, XR_BODY_JOINT_LEFT_HAND_THUMB_METACARPAL_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_THUMB_PHALANX_PROXIMAL, XR_BODY_JOINT_LEFT_HAND_THUMB_PROXIMAL_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_THUMB_PHALANX_DISTAL, XR_BODY_JOINT_LEFT_HAND_THUMB_DISTAL_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_THUMB_TIP, XR_BODY_JOINT_LEFT_HAND_THUMB_DISTAL_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
```
m4gr3d, line 80: This should be `XR_BODY_JOINT_LEFT_HAND_THUMB_TIP_FB`

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,338 @@
...
+	{ XRBodyTracker::JOINT_LEFT_PINKY_FINGER_TIP, XR_BODY_JOINT_LEFT_HAND_LITTLE_TIP_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+
+	// Right hand joints
+	{ XRBodyTracker::JOINT_RIGHT_HAND, XR_BODY_JOINT_RIGHT_HAND_WRIST_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+	{ XRBodyTracker::JOINT_RIGHT_PALM, XR_BODY_JOINT_RIGHT_HAND_PALM_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+	{ XRBodyTracker::JOINT_RIGHT_WRIST, XR_BODY_JOINT_RIGHT_HAND_WRIST_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+	{ XRBodyTracker::JOINT_RIGHT_THUMB_METACARPAL, XR_BODY_JOINT_RIGHT_HAND_THUMB_METACARPAL_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+	{ XRBodyTracker::JOINT_RIGHT_THUMB_PHALANX_PROXIMAL, XR_BODY_JOINT_RIGHT_HAND_THUMB_PROXIMAL_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+	{ XRBodyTracker::JOINT_RIGHT_THUMB_PHALANX_DISTAL, XR_BODY_JOINT_RIGHT_HAND_THUMB_DISTAL_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+	{ XRBodyTracker::JOINT_RIGHT_THUMB_TIP, XR_BODY_JOINT_RIGHT_HAND_THUMB_DISTAL_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
```
m4gr3d, line 109: This should be `XR_BODY_JOINT_RIGHT_HAND_THUMB_TIP_FB`

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,338 @@
...
+	{ XRBodyTracker::JOINT_LEFT_PINKY_FINGER_METACARPAL, XR_BODY_JOINT_LEFT_HAND_LITTLE_METACARPAL_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_PINKY_FINGER_PHALANX_PROXIMAL, XR_BODY_JOINT_LEFT_HAND_LITTLE_PROXIMAL_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_PINKY_FINGER_PHALANX_INTERMEDIATE, XR_BODY_JOINT_LEFT_HAND_LITTLE_INTERMEDIATE_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_PINKY_FINGER_PHALANX_DISTAL, XR_BODY_JOINT_LEFT_HAND_LITTLE_DISTAL_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+	{ XRBodyTracker::JOINT_LEFT_PINKY_FINGER_TIP, XR_BODY_JOINT_LEFT_HAND_LITTLE_TIP_FB, Quaternion(0.5, -0.5, -0.5, -0.5) },
+
+	// Right hand joints
+	{ XRBodyTracker::JOINT_RIGHT_HAND, XR_BODY_JOINT_RIGHT_HAND_WRIST_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+	{ XRBodyTracker::JOINT_RIGHT_PALM, XR_BODY_JOINT_RIGHT_HAND_PALM_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
+	{ XRBodyTracker::JOINT_RIGHT_WRIST, XR_BODY_JOINT_RIGHT_HAND_WRIST_FB, Quaternion(0.5, 0.5, -0.5, 0.5) },
```
m4gr3d, line 105: `HAND` and `WRIST` have the same mapping, is that on purpose?

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
Malcolmnixon: Nice catch - I flubbed that one populating the table.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
Malcolmnixon: And this would be a copy/paste/rename duplication of the previous bug! ;)

Malcolmnixon:
We do try driving the following bones in the body skeleton:
- [Left/Right]Hand
- [Left/Right]Palm
- [Left/Right]Wrist

The [Left/Right]Hand is part of the Godot Humanoid; however the Palm and Wrist tracking points may be more valuable with regards to specific mount points.

Malcolmnixon:
Now that face-tracking has been merged; the following shows driving the avatar using both the face and body trackers simultaneously.

https://github.com/GodotVR/godot_openxr_vendors/assets/1863707/63440b1b-0611-494e-8681-cf93c0a30a33

dsnopek (commented):
This didn't work for me until I made this change:

```
diff --git a/common/src/main/cpp/export/meta_export_plugin.cpp b/common/src/main/cpp/export/meta_export_plugin.cpp
index 49e2ea2..c8d219c 100644
--- a/common/src/main/cpp/export/meta_export_plugin.cpp
+++ b/common/src/main/cpp/export/meta_export_plugin.cpp
@@ -361,9 +361,9 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
                contents += "    <uses-permission android:name=\"com.oculus.permission.BODY_TRACKING\" />\n";
 
                if (body_tracking_value == FACE_TRACKING_OPTIONAL_VALUE) {
-                       contents += "    <uses-feature tools:node=\"replace\" android:name=\"oculus.software.body_tracking\" android:required=\"false\" />\n";
+                       contents += "    <uses-feature tools:node=\"replace\" android:name=\"com.oculus.software.body_tracking\" android:required=\"false\" />\n";
                } else if (body_tracking_value == FACE_TRACKING_REQUIRED_VALUE) {
-                       contents += "    <uses-feature tools:node=\"replace\" android:name=\"oculus.software.body_tracking\" android:required=\"true\" />\n";
+                       contents += "    <uses-feature tools:node=\"replace\" android:name=\"com.oculus.software.body_tracking\" android:required=\"true\" />\n";
                }
        }
```

I don't know why hand tracking is `oculus.software.handtracking` (without the `com.`) whereas this one is `com.oculus.software.body_tracking` (with the `com.`), but it matches what I see in the sample applications in ovr_openxr_mobile_sdk_60.

Anyway, after that it seems to mostly working, although, it does seem to rotate the whole body around the X-axis in a weird way. That could just be what we're getting from the tracking data? Or, maybe this code in `XRBodyModifier3D` isn't right:

https://github.com/godotengine/godot/blob/master/scene/3d/xr_body_modifier_3d.cpp#L366

Maybe we shouldn't be transforming the whole `XRBodyModifier3D` node using the `JOINT_ROOT` data?

m4gr3d:
Can you update the [README](https://github.com/GodotVR/godot_openxr_vendors/blob/master/thirdparty/godot_cpp_gdextension_api/README.md) and update the commit for the Godot build you used to generate `extension_api.json`.

Malcolmnixon:
The avatar rotation issue comes from the fact that all blender/VMC rigging results in the characters facing in the Z+ direction, but Godot orients the camera in the Z- direction.

As such if you drop a mesh into a scene it faces in the Z+ direction, but the moment the XRBodyModifier3D starts moving the bones around, the avatar rotates to align with the players body and faces the Z- direction.

The only "solution" I can think of to extend the Godot Humanoid Importer. Currently it has an option to reset the character to a T-Pose, but I don't believe it rotates the character to face in the Z- direction. In theory it could be modified to do that, but I'm not sure it's the right answer.

Malcolmnixon:
I removed the unnecessary registration (Engine::get_singleton()->register_singleton) of the face and body tracking nodes, as the data is provided by the XRFaceTracker and XRBodyTracker objects.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,338 @@
...
+		xr_body_tracker->set_joint_transform(entry.xr_joint, transform);
+	}
+
+	// The following code adjusts the left and right shoulder back to align them
+	// with how models are designed, rather than the Meta positions of the tips
+	// of the clavicles.
+	if (locations.isActive) {
+		// Deduce the offset from the upper chest transform
+		Transform3D upper_chest = xr_body_tracker->get_joint_transform(XRBodyTracker::JOINT_UPPER_CHEST);
+		Vector3 offset = upper_chest.basis[2] * -0.07;
```
m4gr3d, line 305: What does the `-0.07` value correspond to?

Can you define it as a constant instead.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+	singleton = this;
+}
+
+OpenXRFbBodyTrackingExtensionWrapper::~OpenXRFbBodyTrackingExtensionWrapper() {
+	cleanup();
+}
+
+void OpenXRFbBodyTrackingExtensionWrapper::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("is_enabled"), &OpenXRFbBodyTrackingExtensionWrapper::is_enabled);
```
m4gr3d, line 155: This can be removed if the singleton is no longer exposed.

thirdparty/godot-cpp
m4gr3d, line 1: What's the update to `godot-cpp` for?

m4gr3d (changes requested):

thirdparty/godot-cpp
Malcolmnixon: That must have been from an earlier change. Reverted.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
Malcolmnixon: The _bind_methods() function cannot be removed as the GDCLASS macro references it; however I have removed the is_enabled registration.

m4gr3d (approved):
Looks good!

The checks are failing because of formatting.

dsnopek (commented):
I retested this, and for the most part it's working great!

However, I'd like to make the following suggestions:

- Could this update the export settings for the demo to set "Body Tracking" to "Optional"? This must be done manually in order for body tracking in the demo to work
- In an [earlier comment](https://github.com/GodotVR/godot_openxr_vendors/pull/97#issuecomment-1970277917), you suggested ways to work around the avatar rotating weirdly around the X-axis. Could you implement a fix for that in the demo? It'll present body tracking better.

Thanks!

Malcolmnixon:
> Could this update the export settings for the demo to set "Body Tracking" to "Optional"? This must be done manually in order for body tracking in the demo to work

If we do this, then non-avatar OpenXR games will start capturing and streaming body-tracking data by default. The developer would need to go to the export settings, pick the new property and disable the feature by selecting "None". I'm not sure if that'll be a problem with increased computational load when not needed/used.

dsnopek:
> If we do this, then non-avatar OpenXR games will start capturing and streaming body-tracking data by default.

No, I don't mean changing the default or anything. Just the export settings in the demo project. Right now, it doesn't have body tracking enabled, so to make it work in the demo, it needs to be manually enabled.

Malcolmnixon:
> Just the export settings in the demo project

Gotcha - it's now enabled in the export_presets and functions when run on the headset from a clean build.

Malcolmnixon:
Rather than using XR_BODY_JOINT_ROOT_FB to act as a root position under the character, Meta was not putting anything useful in it. As such I repurposed the XRBodyTracker::JOINT_ROOT to be positioned under the hips at floor level pointing forwards. The result is that the XRBodyModifier3D node will correctly move and orient with the player, making player-centric displays and mount-points much easier to implement.

dsnopek:
I just retested with the latest changes. This is definitely moving in a positive direction! Thanks for the continued dedication to this :-)

Is there a way to keep the avatar's feet rooted in place? When you bend forward, the whole avatar rotates forward, with the feet sliding around unnaturally. Unchecking "Lower Body" doesn't seem to have any effect on this. I tried putting the avatar under the `XRBodyModifier3D` and then the avatar moves around in X-Z space to overlap the player (which we don't want in this demo in any case), and while it's a little hard to tell, it seems like the feet still rotate around in this way.

What I would imagine happening when bending forward, is the upper body bending forward, with the feet staying in the same spot.

Malcolmnixon:
> Is there a way to keep the avatar's feet rooted in place? When you bend forward, the whole avatar rotates forward, with the feet sliding around unnaturally.

I don't see an easy way to do it. The upper body includes the hips (the first bone of the skeleton). When the user bends forwards the hips rotate forward as the first link of the spine, but through that the upper-legs, lower-legs, and feet also move in relationship.

In essence the only way to keep the legs stationary would be to actually bend the upper-legs in relation to the hips to preserve their orientation. As such we would need to "drive the legs" in order to keep them stationary. We don't get any leg information through XR_FB_body_tracking, but I guess we could try synthesizing virtual legs under the hips.

Hopefully this gets solved in a week or two with the next OpenXR release including the XR_FB_body_tracking_fullbody.

godotopenxrmeta/src/main/java/org/godotengine/openxr/vendors/meta/GodotOpenXRMeta.kt
```diff
@@ -153,6 +154,11 @@ class GodotOpenXRMeta(godot: Godot?) : GodotPlugin(godot) {
         if (PermissionsUtil.hasManifestPermission(activity, FACE_TRACKING_PERMISSION)) {
             permissionsToRequest.add(FACE_TRACKING_PERMISSION)
         }
+        // Request the body tracking permission if it's included in the manifest
+        if (PermissionsUtil.hasManifestPermission(activity, BODY_TRACKING_PERMISSION)) {
+            Log.d(TAG, "Requesting permission '${BODY_TRACKING_PERMISSION}'")
+            PermissionsUtil.requestPermission(BODY_TRACKING_PERMISSION, activity)
+        }
```
m4gr3d, line 161: @Malcolmnixon We've changed the logic used for requesting permissions; you need to include the permission in the `permissionsToRequest` list as done above instead.

See https://github.com/GodotVR/godot_openxr_vendors/pull/102 for reference.

common/src/main/cpp/extensions/openxr_fb_face_tracking_extension_wrapper.cpp
```diff
@@ -113,14 +112,14 @@ void OpenXRFbFaceTrackingExtensionWrapper::_on_session_created(uint64_t instance
...
+	XrFaceTrackingDataSource2FB dataSources[2] = {
+		XR_FACE_TRACKING_DATA_SOURCE2_VISUAL_FB,
+		XR_FACE_TRACKING_DATA_SOURCE2_AUDIO_FB
+	};
 	XrFaceTrackerCreateInfo2FB createInfo2 = { XR_TYPE_FACE_TRACKER_CREATE_INFO2_FB };
 	createInfo2.faceExpressionSet = XR_FACE_EXPRESSION_SET2_DEFAULT_FB;
-    createInfo2.requestedDataSourceCount = 2;
-    createInfo2.requestedDataSources = dataSources;
+	createInfo2.requestedDataSourceCount = 2;
+	createInfo2.requestedDataSources = dataSources;
```
m4gr3d, line 122: We're standardizing on using braced-initialization for enums initialization. See https://github.com/GodotVR/godot_openxr_vendors/pull/100#issuecomment-1981726339 for reference.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,359 @@
...
+/*  openxr_fb_body_tracking_extension_wrapper.cpp                         */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Original contributed implementation:                                   */
+/*   Copyright (c) 2024 Malcolm Nixon                                     */
```
m4gr3d, line 11: @Malcolmnixon We follow this approach for 'code donation' by third party, but we'd want to disincentivize this for contributions by team members.

cc @BastiaanOlij

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,359 @@
...
+}
+
+void OpenXRFbBodyTrackingExtensionWrapper::cleanup() {
+	fb_body_tracking_ext = false;
+}
+
+uint64_t OpenXRFbBodyTrackingExtensionWrapper::_set_system_properties_and_get_next_pointer(void *next_pointer) {
+	system_body_tracking_properties.type = XR_TYPE_SYSTEM_BODY_TRACKING_PROPERTIES_FB;
+	system_body_tracking_properties.next = next_pointer;
+	system_body_tracking_properties.supportsBodyTracking = false;
```
m4gr3d, line 164: We're standardizing on using braced-initialization for enums initialization. See https://github.com/GodotVR/godot_openxr_vendors/pull/100#issuecomment-1981726339 for reference.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+void OpenXRFbBodyTrackingExtensionWrapper::_on_session_created(uint64_t instance) {
+	// Skip if not enabled
+	if (!is_enabled()) {
+		return;
+	}
+
+	// Create the body-tracker handle
+	XrBodyTrackerCreateInfoFB createInfo = { XR_TYPE_BODY_TRACKER_CREATE_INFO_FB };
+	createInfo.bodyJointSet = XR_BODY_JOINT_SET_DEFAULT_FB;
```
m4gr3d, line 200: We're standardizing on using braced-initialization for enums initialization. See https://github.com/GodotVR/godot_openxr_vendors/pull/100#issuecomment-1981726339 for reference.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,359 @@
...
+	// Get the next frame time
+	const XrTime next_frame_time = get_openxr_api()->get_next_frame_time();
+	if (next_frame_time == 0) {
+		return;
+	}
+
+	// Construct the expression info struct.
+	XrBodyJointsLocateInfoFB locate_info = { XR_TYPE_BODY_JOINTS_LOCATE_INFO_FB };
+	locate_info.time = next_frame_time;
+	locate_info.baseSpace = (XrSpace)get_openxr_api()->get_play_space();
```
m4gr3d, line 252: We're standardizing on using braced-initialization for enums initialization. See https://github.com/GodotVR/godot_openxr_vendors/pull/100#issuecomment-1981726339 for reference.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+	// Construct the locations struct.
+	XrBodyJointLocationFB fb_locations[XR_BODY_JOINT_COUNT_FB] = {};
+	XrBodyJointLocationsFB locations = { XR_TYPE_BODY_JOINT_LOCATIONS_FB };
+	locations.jointCount = XR_BODY_JOINT_COUNT_FB;
+	locations.jointLocations = fb_locations;
```
m4gr3d, line 258: We're standardizing on using braced-initialization for enums initialization. See https://github.com/GodotVR/godot_openxr_vendors/pull/100#issuecomment-1981726339 for reference.

m4gr3d (changes requested):

godotopenxrmeta/src/main/java/org/godotengine/openxr/vendors/meta/GodotOpenXRMeta.kt
Malcolmnixon: Switched over to use permissionsToRequest.

common/src/main/cpp/extensions/openxr_fb_face_tracking_extension_wrapper.cpp
Malcolmnixon: Switched to brace initialization.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
Malcolmnixon: Removed these from source and header files.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
Malcolmnixon: Switched to brace initialization.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
Malcolmnixon: Switched to brace initialization.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
Malcolmnixon: Switched to brace initialization.

common/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
Malcolmnixon: Switched to brace initialization.

m4gr3d (approved):
Looks good!

Malcolmnixon:
So here we were waiting for the next Khronos OpenXR release with fullbody support... and it wasn't included. I think we have to just get this in as is and hope we get the full body soon to provide legs.

m4gr3d:
Switching this back to `draft` as discussed during the XR meeting

dsnopek:
@Malcolmnixon:

> So here we were waiting for the next Khronos OpenXR release with fullbody support... and it wasn't included. I think we have to just get this in as is and hope we get the full body soon to provide legs.

We have Meta's headers in the repo, which does include the full body extension:

https://github.com/GodotVR/godot_openxr_vendors/blob/master/thirdparty/ovr_openxr_mobile_sdk/OpenXR/Include/openxr/meta_body_tracking_full_body.h

So, even though it's not publish with the official OpenXR headers yet, you should still be able to use it.

There is a `META_VENDOR_ENABLED` define that I think is intended to wrap anything that depends on those headers.

m4gr3d:
@Malcolmnixon You moved the PR out of draft, is it intended to be merged in?

Malcolmnixon:
> is it intended to be merged in?

Yes - I have implemented the XRTracker changes necessary to make this work; although given the overlap between this and other PRs it's likely to need multiple passes to resolve merge conflicts.

PR #98, opened 2024-03-01
Add XR_FB_hand_tracking_mesh extension wrapper and OpenXRFbHandTrackingMesh node

devloglogan:
Adds support for [XR_FB_hand_tracking_mesh](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_FB_hand_tracking_mesh) via a new `OpenXRFbHandTrackingMesh` node.

https://github.com/GodotVR/godot_openxr_vendors/assets/95497005/9a830f73-940c-4a9a-9ecd-764a5f6d5232

This node is intended to work with the new [`XRHandModifier3D`](https://github.com/godotengine/godot/pull/88639) node. The `target` property of `XRHandModifier3D` needs to be directed to `OpenXRFbHandTrackingMesh`'s `Skeleton3D` after all of the bones have been constructed. So, after skeleton construction, `OpenXRFbHandTrackingMesh` will check to see if it is a child of `XRHandModifier3D`, and it will set its `target` property accordingly.

`OpenXRFbHandTrackingMesh` has properties (`use_scale_override` and `scale_override`) that correspond to the `XrHandTrackingScaleFB` struct. If `use_scale_override` is set to true, `scale_override` will be applied to the mesh/skeleton. At the moment I'm not convinced this is how the mesh is supposed to scale (the wrists don't seem to scale at all), but I'm not positive how it's supposed to look, I'll put a picture of an example of a hand scaled 0.5 and 1.75 below.

hand-scale

### DIFF
--- a/common/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
@@ -1,0 +1,130 @@
+/**************************************************************************/
+/*  openxr_fb_hand_tracking_mesh.cpp                                      */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "classes/openxr_fb_hand_tracking_mesh.h"
+
+#include "extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.h"
+
+#include <godot_cpp/classes/xr_hand_modifier3d.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+using namespace godot;
+
+void OpenXRFbHandTrackingMesh::setup_hand_mesh(Hand p_hand) {
+	ERR_FAIL_COND_MSG(skeleton != nullptr, "OpenXRFbHandTrackingMesh skeleton has already been set up");
+
+	if (p_hand == hand) {
+		skeleton = memnew(Skeleton3D);
+		OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->construct_skeleton(p_hand, skeleton);
+		add_child(skeleton);
+
+		mesh_instance = memnew(MeshInstance3D);
+		mesh_instance->set_mesh(OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->get_mesh(p_hand));
+		mesh_instance->set_material_override(material);
+		skeleton->add_child(mesh_instance);
+	}
+
+	XRHandModifier3D *parent = Object::cast_to<XRHandModifier3D>(get_parent());
+	if (parent) {
+		parent->set_target(parent->get_path_to(skeleton));
+	}
+}
+
+Skeleton3D *OpenXRFbHandTrackingMesh::get_skeleton() const {
+	return skeleton;
+}
+
+MeshInstance3D *OpenXRFbHandTrackingMesh::get_mesh_instance() const {
+	return mesh_instance;
+}
+
+void OpenXRFbHandTrackingMesh::set_hand(Hand p_hand) {
+	hand = p_hand;
+}
+
+XRHandTracker::Hand OpenXRFbHandTrackingMesh::get_hand() const {
+	return hand;
+}
+
+void OpenXRFbHandTrackingMesh::set_material(const Ref<Material> &p_material) {
+	material = p_material;
+	if (mesh_instance) {
+		mesh_instance->set_material_override(material);
+	}
+}
+
+Ref<Material> OpenXRFbHandTrackingMesh::get_material() const {
+	return material;
+}
+
+void OpenXRFbHandTrackingMesh::set_use_scale_override(bool p_use) {
+	OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->set_use_scale_override(hand, p_use);
+}
+
+bool OpenXRFbHandTrackingMesh::get_use_scale_override() const {
+	return OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->get_use_scale_override(hand);
+}
+
+void OpenXRFbHandTrackingMesh::set_scale_override(float p_scale) {
+	OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->set_scale_override(hand, p_scale);
+}
+
+float OpenXRFbHandTrackingMesh::get_scale_override() const {
+	return OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->get_scale_override(hand);
+}
+
+void OpenXRFbHandTrackingMesh::_notification(int p_what) {
+	switch (p_what) {
+		case NOTIFICATION_POSTINITIALIZE: {
+			OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->enable_fetch_hand_mesh_data();
+			OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->connect("openxr_fb_hand_tracking_mesh_data_fetched", callable_mp(this, &OpenXRFbHandTrackingMesh::setup_hand_mesh));
+		} break;
+	}
+}
+
+void OpenXRFbHandTrackingMesh::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("get_skeleton"), &OpenXRFbHandTrackingMesh::get_skeleton);
+	ClassDB::bind_method(D_METHOD("get_mesh_instance"), &OpenXRFbHandTrackingMesh::get_mesh_instance);
+	ClassDB::bind_method(D_METHOD("set_hand", "hand"), &OpenXRFbHandTrackingMesh::set_hand);
+	ClassDB::bind_method(D_METHOD("get_hand"), &OpenXRFbHandTrackingMesh::get_hand);
+	ClassDB::bind_method(D_METHOD("set_material", "material"), &OpenXRFbHandTrackingMesh::set_material);
+	ClassDB::bind_method(D_METHOD("get_material"), &OpenXRFbHandTrackingMesh::get_material);
+	ClassDB::bind_method(D_METHOD("set_use_scale_override", "use_scale_override"), &OpenXRFbHandTrackingMesh::set_use_scale_override);
+	ClassDB::bind_method(D_METHOD("get_use_scale_override"), &OpenXRFbHandTrackingMesh::get_use_scale_override);
+	ClassDB::bind_method(D_METHOD("set_scale_override", "scale_override"), &OpenXRFbHandTrackingMesh::set_scale_override);
+	ClassDB::bind_method(D_METHOD("get_scale_override"), &OpenXRFbHandTrackingMesh::get_scale_override);
+
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "hand", PROPERTY_HINT_ENUM, "Left Hand,Right Hand"), "set_hand", "get_hand");
+	ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "material", PROPERTY_HINT_RESOURCE_TYPE, "BaseMaterial3D,ShaderMaterial", PROPERTY_USAGE_DEFAULT), "set_material", "get_material");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "use_scale_override", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NO_EDITOR), "set_use_scale_override", "get_use_scale_override");
+	ADD_PROPERTY(PropertyInfo(Variant::FLOAT, "scale_override", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NO_EDITOR), "set_scale_override", "get_scale_override");
+
+	BIND_ENUM_CONSTANT(Hand::HAND_LEFT);
+	BIND_ENUM_CONSTANT(Hand::HAND_RIGHT);
+}
--- a/common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
@@ -1,0 +1,345 @@
+/**************************************************************************/
+/*  openxr_fb_hand_tracking_mesh_extension_wrapper.cpp                    */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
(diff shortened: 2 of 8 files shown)


common/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+#include "classes/openxr_fb_hand_tracking_mesh.h"
+
+#include "extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.h"
+
+#include <godot_cpp/classes/xr_hand_modifier3d.hpp>
+
+using namespace godot;
+
+void OpenXRFbHandTrackingMesh::setup_hand_mesh(Hand p_hand) {
```
m4gr3d, line 38: Looks like this may need logic to prevent this method from being called several times in a row, otherwise based on the current logic, this will cause the creation and addition of multiple `Skeleton3D` instances to this node.

common/src/main/cpp/include/classes/openxr_fb_hand_tracking_mesh.h
```diff
@@ -0,0 +1,78 @@
...
+namespace godot {
+class OpenXRFbHandTrackingMesh : public Node3D {
+	GDCLASS(OpenXRFbHandTrackingMesh, Node3D)
+public:
+	using Hand = XRHandTracker::Hand;
+
+	enum Mesh {
+		HAND_MESH_LEFT,
+		HAND_MESH_RIGHT
+	};
```
m4gr3d, line 46: What's the reason for another enum instead of reusing the `Hand` enum?

Also the name can be confusing.

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,309 @@
...
+	hand_mesh.type = XR_TYPE_HAND_TRACKING_MESH_FB;
+	hand_mesh.next = nullptr;
+	hand_mesh.jointCapacityInput = 0;
+	hand_mesh.vertexCapacityInput = 0;
+	hand_mesh.indexCapacityInput = 0;
+	XrHandTrackerEXT hand_tracker = reinterpret_cast<XrHandTrackerEXT>(get_openxr_api()->get_hand_tracker(p_hand));
+
+	XrResult result = xrGetHandMeshFB(hand_tracker, &hand_mesh);
+	if (XR_FAILED(result)) {
+		UtilityFunctions::print("Failed to retreive capacity inputs for OpenXR XR_FB_hand_tracking_mesh extension, error code: ", result);
```
m4gr3d, line 148: Typo on `retrieve`

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+	result = xrGetHandMeshFB(hand_tracker, &hand_mesh);
+	if (XR_FAILED(result)) {
+		UtilityFunctions::print("Failed to retreive hand mesh data for OpenXR XR_FB_hand_tracking_mesh extension, error code: ", result);
+		return false;
+	}
+
+	// convert to clockwise winding order to cull correct face side
+	for (int i = 0; i < hand_mesh.indexCapacityInput; i += 3) {
+		SWAP(hand_mesh.indices[i], hand_mesh.indices[i + 2]);
```
m4gr3d, line 174: Can you use descriptive names for `3` and `2`.

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+	hand_mesh_data[p_hand] = hand_mesh;
+
+	return true;
+}
+
+Ref<ArrayMesh> OpenXRFbHandTrackingMeshExtensionWrapper::get_mesh(Hand p_hand) {
+	ERR_FAIL_COND_V_MSG(!is_enabled(), Ref<ArrayMesh>(), "OpenXR extension XR_FB_hand_tracking_mesh is not available");
+
+	XrHandTrackingMeshFB hand_mesh = hand_mesh_data[p_hand];
```
m4gr3d, line 185: Should this validate that `hand_mesh_data` has been populated?

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,309 @@
...
+	for (int i = 0; i < hand_mesh.indexCapacityInput; i += 3) {
+		SWAP(hand_mesh.indices[i], hand_mesh.indices[i + 2]);
+	}
+
+	hand_mesh_data[p_hand] = hand_mesh;
+
+	return true;
+}
+
+Ref<ArrayMesh> OpenXRFbHandTrackingMeshExtensionWrapper::get_mesh(Hand p_hand) {
```
m4gr3d, line 182: Seems like this generates a new `Mesh` every time it's called. Since the underlying data is not supposed to change for the lifetime of the `XrInstance`, should this logic cache the generated `Mesh` and return it on subsequent calls?

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,309 @@
...
+	array_mesh->add_surface_from_arrays(Mesh::PRIMITIVE_TRIANGLES, arrays);
+	array_mesh->regen_normal_maps();
+
+	return array_mesh;
+}
+
+void OpenXRFbHandTrackingMeshExtensionWrapper::construct_skeleton(Skeleton3D *p_skeleton, Hand p_hand) {
+	ERR_FAIL_COND_MSG(!is_enabled(), "OpenXR extension XR_FB_hand_tracking_mesh is not available");
+
+	XrHandTrackingMeshFB hand_mesh = hand_mesh_data[p_hand];
```
m4gr3d, line 234: Should this validate the `hand_mesh_data` has been populated?

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,309 @@
...
+		"LittleDistal",
+		"LittleTip",
+	};
+
+	static const String bone_name_format[2] = {
+		"Left<bone>",
+		"Right<bone>",
+	};
+
+	p_skeleton->clear_bones();
```
m4gr3d, line 270: Should validate that `p_skeleton` is not `nullptr`.

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,309 @@
...
+	arrays[Mesh::ARRAY_WEIGHTS] = godot_bone_weights;
+	arrays[Mesh::ARRAY_INDEX] = godot_indices;
+
+	array_mesh->add_surface_from_arrays(Mesh::PRIMITIVE_TRIANGLES, arrays);
+	array_mesh->regen_normal_maps();
+
+	return array_mesh;
+}
+
+void OpenXRFbHandTrackingMeshExtensionWrapper::construct_skeleton(Skeleton3D *p_skeleton, Hand p_hand) {
```
m4gr3d, line 231: Seems like the `Skeleton3D` is the _return value_; in this case it should be the second argument, and the variable name should be updated to `r_skeleton`.

demo/main.tscn
```diff
@@ -62,12 +62,12 @@ tracker = &"left_hand"
 pose = &"grip"
 
 [node name="LeftHandMesh" type="MeshInstance3D" parent="XROrigin3D/LeftHand"]
+visible = false
 mesh = SubResource("BoxMesh_3kt6b")
 
-[node name="LeftControllerFbRenderModel" type="OpenXRFbRenderModel" parent="XROrigin3D/LeftHand"]
```
m4gr3d, line 67: If I'm reading right, the controller render model is replaced by the hand tracking mesh. Should we have logic to instead switch between the two based on whether the user is using a controller, or using hand tracking?

thirdparty/godot-cpp
m4gr3d, line 1: What version of `godot-cpp` is this updating to?

demo/main.tscn
dsnopek: Hand tracking data is available even when using a controller, which would just result in the hand being shown on the controller, which is what some apps want. I think how to make controllers and hands interact should be up to the app, and in the case of the demo app here, I think it'd be cool to show both the controller and the hand at the same time. :-)

devloglogan (commented):
Believe I've addressed all of the smaller comments. Thanks so much for the thorough review @m4gr3d! I am going to make some changes to what is cached in the extension wrapper.

devloglogan:
I've updated this PR so that all of the `XrHandTrackingMeshFB` data is now stored in `LocalVector`s (so no need to worry about memory cleanup I believe), and only the data relevant to skeleton construction will persist. Meanwhile the mesh data will be used to construct `ArrayMesh` resources, and those will be cached instead of the raw mesh data.

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,309 @@
...
+	ADD_SIGNAL(MethodInfo("hand_mesh_data_fetched", PropertyInfo(Variant::INT, "hand_index")));
+}
+
+void OpenXRFbHandTrackingMeshExtensionWrapper::cleanup() {
+	fb_hand_tracking_mesh_ext = false;
+
+	for (int i = 0; i < Hand::HAND_MAX; i++) {
+		is_hand_mesh_data_fetched[i] = false;
+		hand_tracking_scale[i].overrideHandScale = false;
+		hand_tracking_scale[i].overrideValueInput = 1.0;
```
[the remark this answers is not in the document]
dsnopek: Yeah, this should probably do:

```
if (hand_mesh[i].is_valid()) {
    hand_mesh[i].unref();
}
```

We may also want to clear out the `LocalVector`s on `bone_data`?

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,325 @@
...
+		if (!is_hand_mesh_data_fetched[i]) {
+			XrHandTrackerEXT hand_tracker = reinterpret_cast<XrHandTrackerEXT>(get_openxr_api()->get_hand_tracker(i));
+			if (hand_tracker != XR_NULL_HANDLE) {
+				if (fetch_hand_mesh_data(Hand(i))) {
+					is_hand_mesh_data_fetched[i] = true;
+					emit_signal("openxr_fb_hand_tracking_mesh_data_fetched", i);
+				}
+			}
+		}
+	}
```
dsnopek, line 124: Does this need to happen in `_on_process()`? Or, could we do it in `_on_state_ready()`? Or, maybe even only after `OpenXRFbHandTrackingMesh` has called a method asking for the mesh (or otherwise indicated that it's interested)?

Not all apps using 'godot_openxr_vendors' will want to use this extension even if it's available, so it'd be best to only do this on demand. Or, if that's not possible, then we should have a project setting to enable or disable it (but those are hard for developers to notice, so that option is not ideal).

dsnopek (commented):
Thanks! Sorry for the delay in reviewing this one.

For the most part, I think the current code looks great! I tested on the Meta Quest 3 and it worked great too (they are fairly aesthetically pleasing hand models!)

In addition to my code comments:

- The `OpenXRFbHandTrackingMesh` node should have a `Material` property that sets the material on the `MeshInstance3D`
- And, it should probably also have a `get_skeleton()` method that let's the developer easily grab the `Skeleton3D` node. Right now, it's easy to use this with `XRHandModifier3D` but if a developer wants to drive the skeleton with an animation player, they'd have to guess the node's path. For completeness, maybe a `get_mesh_instance()` method too?
- I already mentioned this in a response to one of Fredia's comments, but I think the demo project should show _both_ the controller models and the hands (assuming we have data for both - if we don't have hands or models, then we'll still want the cubes)

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
devloglogan: We discussed this a bit further and this will remain in `on_process()`, but I have changed it so it will only fetch the mesh data if `OpenXRFbHandTrackingMesh` nodes are used.

common/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+void OpenXRFbHandTrackingMesh::set_scale_override(float p_scale) {
+	scale_override = p_scale;
+	OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->set_scale_override(hand, p_scale);
+}
+
+float OpenXRFbHandTrackingMesh::get_scale_override() {
+	return scale_override;
+}
```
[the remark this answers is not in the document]
devloglogan: Updated to the suggested changes here, the properties are now using `PROPERTY_USAGE_NO_EDITOR` and can only be edited via code.

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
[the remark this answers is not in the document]
devloglogan: Fixed in latest push.

common/src/main/cpp/include/classes/openxr_fb_hand_tracking_mesh.h
```diff
@@ -0,0 +1,78 @@
...
+public:
+	using Hand = XRHandTracker::Hand;
+
+private:
+	Hand hand = Hand::HAND_LEFT;
+	Ref<Material> material;
+
+	Skeleton3D *skeleton = nullptr;
+
+	MeshInstance3D *mesh_instance = nullptr;
```
m4gr3d, line 50: Should we use a `Ref<>` reference instead for both `skeleton` and `mesh_instance`?

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,347 @@
...
+	request_extensions[XR_FB_HAND_TRACKING_MESH_EXTENSION_NAME] = &fb_hand_tracking_mesh_ext;
+	singleton = this;
+}
+
+OpenXRFbHandTrackingMeshExtensionWrapper::~OpenXRFbHandTrackingMeshExtensionWrapper() {
+	cleanup();
+}
+
+void OpenXRFbHandTrackingMeshExtensionWrapper::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("is_enabled"), &OpenXRFbHandTrackingMeshExtensionWrapper::is_enabled);
```
m4gr3d, line 59: You can remove this binding since the data/functionality will be exposed by the mesh node.

common/src/main/cpp/register_types.cpp
```diff
@@ -96,8 +101,10 @@ void initialize_plugin_module(ModuleInitializationLevel p_level) {
 			Engine::get_singleton()->register_singleton("OpenXRFbSpatialEntityContainerExtensionWrapper", OpenXRFbSpatialEntityContainerExtensionWrapper::get_singleton());
 			Engine::get_singleton()->register_singleton("OpenXRFbSceneExtensionWrapper", OpenXRFbSceneExtensionWrapper::get_singleton());
 			Engine::get_singleton()->register_singleton("OpenXRFbFaceTrackingExtensionWrapper", OpenXRFbFaceTrackingExtensionWrapper::get_singleton());
+			Engine::get_singleton()->register_singleton("OpenXRFbHandTrackingMeshExtensionWrapper", OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton());
```
m4gr3d, line 104: You can remove this registration since the api is exposed via the mesh node instead.

demo/main.tscn
m4gr3d: If we are able to animate the hand model so that it conforms to the controller, then yes that would be great to have. Otherwise I was thinking more about checking the source for the hand tracking data and showing the controller render models if the source is from controllers, and the hand mesh instead if the source is from actual hand tracking.

This will especially help illustrate the new Quest multimodal capability of being able to use both controller and hands at the same time.

https://www.uploadvr.com/quest-3-pro-hands-controllers-simultaneous-multimodal/

m4gr3d (changes requested):

m4gr3d (changes requested):

common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -0,0 +1,347 @@
...
+}
+
+Ref<ArrayMesh> OpenXRFbHandTrackingMeshExtensionWrapper::get_mesh(Hand p_hand) {
+	ERR_FAIL_COND_V_MSG(!is_enabled(), Ref<ArrayMesh>(), "OpenXR extension XR_FB_hand_tracking_mesh is not available");
+	ERR_FAIL_COND_V_MSG(hand_mesh[p_hand].is_null(), Ref<ArrayMesh>(), "OpenXR extension XR_FB_hand_tracking_mesh has not populated mesh data");
+
+	return hand_mesh[p_hand];
+}
+
+void OpenXRFbHandTrackingMeshExtensionWrapper::construct_skeleton(Hand p_hand, Skeleton3D *r_skeleton) {
```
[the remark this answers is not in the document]
dsnopek: World scale will be applied by `XRHandModifier3D` when animating the skeleton, I don't think it should be applied here.

common/src/main/cpp/include/classes/openxr_fb_hand_tracking_mesh.h
dsnopek: Those are nodes which aren't reference counted, so we can't use `Ref<>` with them.

demo/main.tscn
dsnopek: > If we are able to animate the hand model so that it conforms to the controller, then yes that would be great to have.

Yep, we can! When testing this while having controllers in my hands, the hands will be animated based on where I'm touching the controllers (you just can't see the controllers, because the controller models are gone). So, I think just keeping the controller models in there should allow this to work.

m4gr3d: Just tested and it looks great!

Although it looks weird when you're just using hand tracking and the controller render models remain attached to your hand.

I think when using controllers, we should show both controller render models and the hand meshes, but when only using hand tracking, we should hide the controller render models. We should be able to do that by checking the source of the hand tracking data.

common/src/main/cpp/register_types.cpp
devloglogan: Removed this as well as the `is_enabled` function binding in latest push.

demo/main.tscn
devloglogan: Added some logic in the latest push that utilizes `OpenXRInterface.get_hand_tracking_soruce()` to hide the controller models when hand tracking is being used vs. controllers. It would be handy to have a signal for when the sources change, maybe that's something we want to consider adding?

m4gr3d: Sounds like a good idea. Does the extension provide an event for it, or would we have to poll?

m4gr3d (approved):

demo/main.tscn
devloglogan: There doesn't seem to be an event related to hand tracking data sources but the `XR_TYPE_EVENT_DATA_INTERACTION_PROFILE_CHANGED` event seems to occur every time the source changes. `XRPositionalTracker` has a [`profile_changed`](https://github.com/godotengine/godot/blob/39f90600bc553f2273e72708eb2a8d37a8517e86/servers/xr/xr_positional_tracker.cpp#L73) signal. Maybe it could be bound to [`XRController3D`](https://github.com/godotengine/godot/blob/39f90600bc553f2273e72708eb2a8d37a8517e86/scene/3d/xr_nodes.cpp#L473) along with the others?

m4gr3d: Yes, let's do that!

PR #105, opened 2024-03-07
Make XR_FB_hand_tracking_aim project setting basic

devloglogan:
Project setting for `XR_FB_hand_tracking_aim` extension is currently an advanced project setting, but it was discussed in the xr contributors chat that it should be basic. This applies that!


m4gr3d (approved):

BastiaanOlij (approved):
lgtm

PR #107, opened 2024-03-11
Add a developer-facing API for interacting with scene anchors

dsnopek:
This builds on the work from PR https://github.com/GodotVR/godot_openxr_vendors/pull/66 which added extension wrappers for XR_FB_scene and all the related extensions, but didn't expose a nice developer-facing API.

**This PR adds that API!**

It's centered around a new `OpenXRFbSceneManager` node, where you configure scenes to be instantiated for each scene anchor depending on it's semantic labels:

Selection_124

For each scene anchor that it discovers, it creates an `XRAnchor3D` node (which gets its position from an `XRPositionTracker`) <-- this is the generic support for anchors already built into Godot, which is also used by the [ARKit plugin](https://github.com/godotengine/godot-ios-plugins/tree/master/plugins/arkit)

And then depending on the semantic labels of the scene anchor, it instantiates the given scene and makes it a child of the `XRAnchor3D`, and calls a method on it with an `OpenXRFbSpatialEntity` object, which the developer can use to query any information they want about the anchor, for example, including easily creating a `MeshInstance3D` or `CollisionShape3D`.

The demo project uses this to place some boxes and labels in the room based on the captured scene data:

e56d0a2a027db5dfb386cf3db8561067
**Press the Y button to toggle on/off passthrough (the scene anchors will only be shown when in passthrough) and the Menu button to do scene capture.**

I also took the opportunity to make a friendlier way to access the scene capture API. Rather than calling `OpenXRFbSceneCaptureExtensionWrapper.request_scene_capture()`, this PR adds a similar method on the `OpenXRFbSceneManager` node, so developers should be able to do everything related to scene anchors via that node (never having to interact with the singletons).

This PR was already getting huge, so there's a couple things I've left out for now, which I plan to add in follow-ups:

- Getting the triangle mesh for an anchor - this requires integrating with the `XR_FB_triangle_mesh` extension
- Developer-created spatial anchors (this PR only covers scene anchors)


common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
```diff
@@ -0,0 +1,390 @@
...
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "visible", PROPERTY_HINT_NONE, ""), "set_visible", "get_visible");
+
+	ADD_SIGNAL(MethodInfo("scene_anchor_created", PropertyInfo(Variant::Type::OBJECT, "scene_node"), PropertyInfo(Variant::Type::OBJECT, "spatial_entity")));
+	ADD_SIGNAL(MethodInfo("scene_data_missing"));
+	ADD_SIGNAL(MethodInfo("scene_capture_completed", PropertyInfo(Variant::Type::BOOL, "success")));
+}
+
+bool OpenXRFbSceneManager::_set(const StringName &p_name, const Variant &p_value) {
+	PackedStringArray parts = p_name.split("/", true, 2);
+	if (parts[0] == "scenes") {
```
m4gr3d, line 81: Do we need to check that `parts.size() == 2`?

common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
```diff
@@ -0,0 +1,390 @@
...
+		}
+	}
+	return false;
+}
+
+bool OpenXRFbSceneManager::_get(const StringName &p_name, Variant &r_ret) const {
+	PackedStringArray parts = p_name.split("/", true, 2);
+	if (parts[0] == "scenes") {
+		const PackedStringArray &semantic_labels = OpenXRFbSceneExtensionWrapper::get_supported_semantic_labels();
+		if (semantic_labels.has(parts[1])) {
```
m4gr3d, line 95: I'm curious why we need to check that`semantic_labels` contains the given key. Can't we just check that `scenes[parts[1]] != nullptr`?

common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
```diff
@@ -0,0 +1,390 @@
...
+void OpenXRFbSceneManager::show() {
+	set_visible(true);
+}
+
+void OpenXRFbSceneManager::hide() {
+	set_visible(false);
+}
+
+Error OpenXRFbSceneManager::create_scene_anchors() {
+	ERR_FAIL_COND_V(anchors_created, FAILED);
```
m4gr3d, line 213: Wondering if there's another error we can use to signify the anchors have already been created.

common/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+		AABB bounding_box = get_bounding_box_3d();
+		box_mesh->set_size(bounding_box.size);
+
+		mesh_instance = memnew(MeshInstance3D);
+		mesh_instance->set_mesh(box_mesh);
+		mesh_instance->set_position(bounding_box.get_center());
+
+	} else if (is_component_enabled(COMPONENT_TYPE_BOUNDED_2D)) {
+		Ref<PlaneMesh> plane_mesh;
```
m4gr3d, line 198: Out of curiosity, wouldn't a quad mesh be more appropriate here?

common/src/main/cpp/extensions/openxr_fb_scene_capture_extension_wrapper.cpp
```diff
@@ -96,16 +96,38 @@ void OpenXRFbSceneCaptureExtensionWrapper::_on_instance_destroyed() {
...
 	XrSceneCaptureRequestInfoFB request;
 	request.type = XR_TYPE_SCENE_CAPTURE_REQUEST_INFO_FB;
 	request.next = nullptr;
-	request.requestByteCount = 0;
-	request.request = nullptr;
-	XrResult result = xrRequestSceneCaptureFB((XrSession)get_openxr_api()->get_session(), &request, &requestId);
-	scene_capture_enabled = (result == XR_SUCCESS);
-	return scene_capture_enabled;
+	request.requestByteCount = cstr.size();
+	request.request = cstr.ptr();
```
m4gr3d, line 114: Can you update this initialization to braced-initialization.

common/src/main/cpp/extensions/openxr_fb_scene_capture_extension_wrapper.cpp
```diff
@@ -96,16 +96,38 @@ void OpenXRFbSceneCaptureExtensionWrapper::_on_instance_destroyed() {
...
+	XrResult result = xrRequestSceneCaptureFB((XrSession)get_openxr_api()->get_session(), &request, &request_id);
+	if (!XR_SUCCEEDED(result)) {
+		WARN_PRINT("xrRequestSceneCaptureFB failed!");
+		WARN_PRINT(get_openxr_api()->get_error_string(result));
+		p_callback(result, p_userdata);
+		return false;
+	}
+
+	scene_capture_enabled = true;
+	requests[request_id] = RequestInfo(p_callback, p_userdata);
```
m4gr3d, line 125: Seems like the only scenario where we can have multiple requests is when a scene capture fails and `scene_capture_enabled` is not set to `true`. Is that the intended behavior?

dsnopek: OpenXR allows for multiple requests at a time, and so do a number of other similar asynchronous APIs, for which I've used this same pattern. So, I did it this way largely for consistency with all the other places we're doing asynchronous requests.

However, from a developer experience perspective, we don't want to allow multiple requests at a time - if the scene capture interface has opened up, we want to let that completely finish before making another request. And unless I made a mistake somewhere, the `scene_capture_enabled` flag should prevent multiple requests in all cases (even if we get an error - it'll either be reset or not yet set in the first place).

So, if we don't care about the consistency thing, we could drop the `HashMap` and just keep a single `RequestInfo *` for the current request? I could go either way. Which do you prefer?

common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
dsnopek: Ah, yes, we do, thanks! Updated in my latest push

common/src/main/cpp/extensions/openxr_fb_scene_capture_extension_wrapper.cpp
dsnopek: Certainly! Updated in my latest push

common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
dsnopek: Ah, yeah, we could use `ERR_ALREADY_EXISTS`! Updated in my latest push

common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
```diff
@@ -0,0 +1,390 @@
...
+			scenes[parts[1]] = p_value;
+			return true;
+		}
+	}
+	return false;
+}
+
+bool OpenXRFbSceneManager::_get(const StringName &p_name, Variant &r_ret) const {
+	PackedStringArray parts = p_name.split("/", true, 2);
+	if (parts[0] == "scenes") {
```
m4gr3d, line 93: Similar comment as above here about checking for `parts.size() == 2`.

common/src/main/cpp/extensions/openxr_fb_scene_capture_extension_wrapper.cpp
m4gr3d: Let's keep it as is for consistency. 
Thanks for the clarification!

common/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
BastiaanOlij: Yeah `Quadmesh` is a subclass of `PlaneMesh` with just some different default settings, it's mostly a convenience as this is often used for billboarding. So I would use `PlaneMesh` in most scenarios.

BastiaanOlij (approved):
Sorry, didn't see the PR till now. I need to play around with this some day but purely looking at this from a code perspective, this looks really nice!

It needs documentation, both from a user POV and from someone who wants to understand the underlying approach, but we need to decide on a place to add this. Maybe we should add a help system similar to what we did with XR tools?

common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
```diff
@@ -0,0 +1,391 @@
...
+	ClassDB::bind_method(D_METHOD("get_spatial_entity", "uuid"), &OpenXRFbSceneManager::get_spatial_entity);
+
+	ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "default_scene", PROPERTY_HINT_RESOURCE_TYPE, "PackedScene"), "set_default_scene", "get_default_scene");
+	ADD_PROPERTY(PropertyInfo(Variant::STRING_NAME, "scene_setup_method", PROPERTY_HINT_NONE, ""), "set_scene_setup_method", "get_scene_setup_method");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "auto_create", PROPERTY_HINT_NONE, ""), "set_auto_create", "get_auto_create");
+	ADD_PROPERTY(PropertyInfo(Variant::BOOL, "visible", PROPERTY_HINT_NONE, ""), "set_visible", "get_visible");
+
+	ADD_SIGNAL(MethodInfo("scene_anchor_created", PropertyInfo(Variant::Type::OBJECT, "scene_node"), PropertyInfo(Variant::Type::OBJECT, "spatial_entity")));
+	ADD_SIGNAL(MethodInfo("scene_data_missing"));
+	ADD_SIGNAL(MethodInfo("scene_capture_completed", PropertyInfo(Variant::Type::BOOL, "success")));
```
m4gr3d, line 76: This is redefining a signal that's already present in `OpenXRFbSceneCaptureExtensionWrapper`, but including an additional parameter.

For this and the other signals in this PR, can you instead make the signal names more explicit (e.g: `openx_fb_scene_capture_completed`).

m4gr3d (approved):

common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
BastiaanOlij: Once this node works properly, that is the way users should be using this. We should strive to make this good enough so we don't need to expose the wrapper classes to Godot.

If we do decide to keep that status quo, I agree with Fredia here, the signals on the wrapper should be more explicit.

dsnopek: What is the advantage of making these super long signal names?

The class name is already `OpenXRFbSceneManager` so just having `OpenXRFbSceneManager.scene_capture_completed` seems like an OK signal name. In most native Godot nodes, the signal names aren't prefixed with some part of the class name. I see that we've been doing this on other classes in 'godot_openxr_vendors', but what's the rationale for it?

m4gr3d: For Godot native nodes, I agree that the signal names can (and should) be succinct. For vendor extensions on the other hand, we're making the signal names explicit for a couple of reasons:
- provide clear context on the origin and role of the signal: whether it's during development, or while reading documentation / guides / tutorial, this provides clear context to the developer that the functionality is vendor specific. This is especially relevant to apps / games that target multiple platforms as they may be using different set of signals depending on the platform they're running on.
- Unlike Godot native nodes, we make no backward compatibility promises on the vendor extensions. Instead we strive to deprecate them in favor of core or multi-vendor extensions the moment those are available. Making the signal names explicit makes it easy for developers to identify logic that should be updated because it's deprecated (especially bearing in mind that not all devs use the Godot Editor as their default dev environment). This PR has a good a example of what I describe; the `OpenXRFbSceneCaptureExtensionWrapper.scene_capture_completed` signal is technically being deprecated (as we don't intend for users to use the wrapper class) and replaced by the `OpenXRFbSceneManager.scene_capture_completed` signal, which also adds an additional parameter. As a dev making the switch to `OpenXRFbSceneManager`, it would be easy to forget updating the logic tied to the signal given the similarities between the old and the new. 
  - E.g: This is the code being removed in the demo; note that signal connection uses `scene_capture_completed` and not `OpenXRFbSceneManager.scene_capture_completed`
```
var scene_capture := Engine.get_singleton("OpenXRFbSceneCaptureExtensionWrapper")
if scene_capture:
	scene_capture.connect("scene_capture_completed", _on_scene_capture_completed)
```

m4gr3d: The pattern so far has been to use part of the class name. For example `openxr_fb_hand_tracking_mesh_data_fetched` for the `OpenXRFbHandTrackingMeshExtensionWrapper` class.

dsnopek: Signal names update in my latest push!

PR #109, opened 2024-03-13
Switch to composition layer API that allows multiple layers.

dsnopek:
This adapts the passthrough extension wrapper for the GDExtension API changes in PR https://github.com/godotengine/godot/pull/89460

~~Leaving this marked as a draft until that PR is merged.~~

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
@@ -210,13 +210,22 @@ bool OpenXRFbPassthroughExtensionWrapper::start_passthrough() {
 	return true;
 }
 
-uint64_t OpenXRFbPassthroughExtensionWrapper::_get_composition_layer() {
-	if (is_passthrough_enabled()) {
+int OpenXRFbPassthroughExtensionWrapper::_get_composition_layer_count() {
+	return is_passthrough_enabled() ? 1 : 0;
+}
+
+uint64_t OpenXRFbPassthroughExtensionWrapper::_get_composition_layer(int p_index) {
+	if (p_index == 0) {
 		composition_passthrough_layer.layerHandle = passthrough_layer;
 		return reinterpret_cast<uint64_t>(&composition_passthrough_layer);
 	} else {
 		return 0;
 	}
+}
+
+int OpenXRFbPassthroughExtensionWrapper::_get_composition_layer_order(int p_index) {
+	// Ensure the passthrough layer will be behind the projection layer.
+	return -100;
 }
 
 void OpenXRFbPassthroughExtensionWrapper::stop_passthrough() {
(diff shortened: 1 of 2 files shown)


dsnopek:
The upstream PR has been merged!

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -210,15 +210,24 @@ bool OpenXRFbPassthroughExtensionWrapper::start_passthrough() {
...
 		composition_passthrough_layer.layerHandle = passthrough_layer;
 		return reinterpret_cast<uint64_t>(&composition_passthrough_layer);
 	} else {
 		return 0;
 	}
 }
 
+int OpenXRFbPassthroughExtensionWrapper::_get_composition_layer_order(int p_index) {
+	// Ensure the passthrough layer will be below the projection layer.
+	return -9999;
```
m4gr3d, line 228: nit: Does it make sense to use `in front of` / `behind` instead of `below` / `above` when describing the layer ordering?

dsnopek: Hm, yeah, I guess "behind" is a little clearer! Updated in latest push

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -210,15 +210,24 @@ bool OpenXRFbPassthroughExtensionWrapper::start_passthrough() {
...
 		composition_passthrough_layer.layerHandle = passthrough_layer;
 		return reinterpret_cast<uint64_t>(&composition_passthrough_layer);
 	} else {
 		return 0;
 	}
 }
 
+int OpenXRFbPassthroughExtensionWrapper::_get_composition_layer_order(int p_index) {
+	// Ensure the passthrough layer will be behind the projection layer.
+	return -9999;
```
m4gr3d, line 228: I am concerned starting with a high number would set a bad precedent to the rest of the community. 

I think instead we should decide on a range based on use-cases and constrain layer ordering within that range with clearly define constants to indicate the ordering of core capabilities. 

To avoid blocking this PR, I'd suggest starting with -100 instead (we have until v3.0 is publicly release to update that value if needed), and let's add the proposal as a topic of discussion.

m4gr3d (approved):

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
dsnopek: Makes sense! Updated :-)

(FYI, I took the `-9999` from [here](https://github.com/godotengine/godot/pull/76877/files#diff-40ce218497b07758c62e07b8cf35e4dfe5d5f3029e27ad3a1f9cc27d942bdd2fR169), which actually defaults all other layers to `1`, which is what I've continued using in my continued work on composition layers.)

PR #111, opened 2024-03-24
Add Android Surface swapchain support

maunvz:
This is a draft for adding an AndroidSurfaceLayer object that extends Node3D and does the following:
 - Creates an Android Surface backed swapchain
 - Sets up an XrCompositionLayerQuad to submit (using Node3D pose, visibility, etc)
 - Stores an in handle for that Surface that can be passed to an Android plugin to retrieve the Surface

With this change, a project can use Android Surfaces to display content on OpenXR layers and control them with the Godot scene graph.

TODO: Add an example to the demo project

### DIFF
--- a/common/src/main/cpp/classes/android_surface_layer.cpp
+++ b/common/src/main/cpp/classes/android_surface_layer.cpp
@@ -1,0 +1,129 @@
+/**************************************************************************/
+/*  android_surface_layer.cpp                                             */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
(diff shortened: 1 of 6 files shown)


BastiaanOlij:
We'll need to see how this overlaps with the logic we're building to core around composition layers. Also seeing these are all KHR extensions it probably belongs in core, not in the vendors plugin.

dsnopek:
I already have a nearly finished PR for Godot that adds support for composition layers, including XrCompositionLayerQuad. I'll see if I can finish it up quick, or maybe just make a draft PR..

dsnopek:
Here's my composition layer PR: https://github.com/godotengine/godot/pull/89880

dsnopek:
I spent some more time looking into this!

If I'm understanding correctly, the `XR_KHR_android_surface_swapchain` extension allows creating a swap chain where the data drawn onto the swapchain is pulled from an Android `Surface`. So, if you've got an Android plugin that's drawing to an Android `Surface`, you could use this extension to display that on an OpenXR composition layer.

In that case, I think it would probably make sense to support this by building on PR https://github.com/godotengine/godot/pull/89880, to allow configuring the composition layer nodes to use an Android `Surface` rather than a Godot `SubViewport` as the source for the composition layer.

And, I agree with @BastiaanOlij that this should be done in Godot (rather than 'godot_openxr_vendors') because the core extension is `XR_KHR_android_surface_swapchain`, which is vendor neutral.

The PR here also uses `XR_FB_android_surface_swapchain` (a Meta vendor extension), but if we want to keep that bit out of Godot, we'll need to add some hook to Godot that allows GDExtensions to modify the swap chain creation. But that shouldn't be a big change, and we'll likely already need something similar to allow modifying composition layer creation (to support numerous other vendor extensions, like, for example, `XR_FB_composition_layer_secure_content`).

dsnopek:
I just posted PR https://github.com/godotengine/godot/pull/96185 which adds support for `XR_KHR_android_surface_swapchain` into Godot itself.

It doesn't include support for `XR_FB_android_surface_swapchain_create` since that's a vendor extension. In order to add that via GDExtension we'll need to make some other small Godot changes, but that'll come in a future PR.

Closing this PR as superseded by https://github.com/godotengine/godot/pull/96185

PR #115, opened 2024-04-03
Implement remaining functionality in XR_FB_passthrough extension wrapper

devloglogan:
This PR adds all of the remaining passthrough functionality that we're currently missing for the XR_FB_passthrough extension. Quick video showing some of the features:

https://github.com/GodotVR/godot_openxr_vendors/assets/95497005/0649d152-5e1e-45c9-9111-202ffee8883c

### Geometry Passthrough
I've added a new `OpenXRFbPassthroughGeometry` node, when visible, these nodes will register themselves with `OpenXRFbPassthroughExtensionWrapper` and begin geometry passthrough. If the `blend_mode` of `XRInterface` is set to `XR_ENV_BLEND_MODE_ALPHA_BLEND`, then "full passthrough" should always take place instead. These nodes can be moved and scaled, and their passthrough geometry will update accordingly.

### Passthrough Style
Some functions have been added to `OpenXRFbPassthroughExtensionWrapper` to update passthrough styles. Note that edge color is available in all styles, while the other three are mutually exclusive and will automatically enable when set.
- Edge Color: Pass a `Color` into the `set_edge_color()` function
- Color Map: Pass a `GradientTexture1D` into the `set_color_map()` function
- Mono Map: Pass a `Curve` into the `set_mono_map()` function
- Brightness/Contrast/Saturation: Pass three corresponding floats into the `set_brightness_contrast_saturation()` function.

### Passthrough Capabilities
`OpenXRFbPassthroughExtensionWrapper` also now has some functions that can be called to check the capabilities of the headset.

I've also added a signal that should fire if there are any `XrEventDataPassthroughStateChangedFB` events passed through, though I'm unsure at the moment if there's a way to test these.

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
@@ -30,12 +30,14 @@
 
 #include "extensions/openxr_fb_passthrough_extension_wrapper.h"
 
+#include <godot_cpp/classes/gradient.hpp>
 #include <godot_cpp/classes/main_loop.hpp>
 #include <godot_cpp/classes/open_xrapi_extension.hpp>
 #include <godot_cpp/classes/os.hpp>
 #include <godot_cpp/classes/scene_tree.hpp>
 #include <godot_cpp/classes/viewport.hpp>
 #include <godot_cpp/classes/xr_server.hpp>
+#include <godot_cpp/templates/local_vector.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
 using namespace godot;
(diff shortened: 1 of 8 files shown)


m4gr3d (approved):
Logic looks good!

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -65,12 +67,47 @@ OpenXRFbPassthroughExtensionWrapper::~OpenXRFbPassthroughExtensionWrapper() {
...
+	ClassDB::bind_method(D_METHOD("set_mono_map"), &OpenXRFbPassthroughExtensionWrapper::set_mono_map);
+	ClassDB::bind_method(D_METHOD("set_brightness_contrast_saturation"), &OpenXRFbPassthroughExtensionWrapper::set_brightness_contrast_saturation);
+
+	ClassDB::bind_method(D_METHOD("has_passthrough_capability"), &OpenXRFbPassthroughExtensionWrapper::has_passthrough_capability);
+	ClassDB::bind_method(D_METHOD("has_color_passthrough_capability"), &OpenXRFbPassthroughExtensionWrapper::has_color_passthrough_capability);
+	ClassDB::bind_method(D_METHOD("has_layer_depth_passthrough_capability"), &OpenXRFbPassthroughExtensionWrapper::has_layer_depth_passthrough_capability);
+
+	ADD_SIGNAL(MethodInfo("openxr_fb_projected_passthrough_layer_created"));
+	ADD_SIGNAL(MethodInfo("openxr_fb_passthrough_stopped"));
+	ADD_SIGNAL(MethodInfo("passthrough_state_changed", PropertyInfo(Variant::INT, "event_type")));
```
m4gr3d, line 92: Can you update this signal to `openxr_fb_passthrough_state_changed` to match the pattern of the above signals.

m4gr3d (changes requested):

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
devloglogan: Updated!

devloglogan:
Rebased!

m4gr3d (approved):

PR #117, opened 2024-04-05
Add support for XR_FB_composition_layer_secure_content

dsnopek:
This depends on PR https://github.com/godotengine/godot/pull/90237

It adds a new "External Output" property to `OpenXRCompositionLayer` nodes, which can be used to set the flags used by XR_FB_composition_layer_secure_content to configure the given composition layer:

Selection_138

I've tested this on the Meta Quest 3 and it works!

When casting to the web, setting the composition layer to "Replace" will display it blacked out in the cast, and setting to "Exclude" will remove it from the cast entirely. As expected, the setting has no effect in the headset.

Marking as draft until the Godot PR is merged.

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
@@ -1,0 +1,134 @@
+/**************************************************************************/
+/*  openxr_fb_composition_layer_secure_content_extension_wrapper.cpp      */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.h"
+
+#include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+using namespace godot;
+
+static const char *EXTERNAL_OUTPUT_PROPERTY_NAME = "XR_FB_composition_layer_secure_content/external_output";
+
+OpenXRFbCompositionLayerSecureContentExtensionWrapper *OpenXRFbCompositionLayerSecureContentExtensionWrapper::singleton = nullptr;
+
+OpenXRFbCompositionLayerSecureContentExtensionWrapper *OpenXRFbCompositionLayerSecureContentExtensionWrapper::get_singleton() {
+	if (singleton == nullptr) {
+		singleton = memnew(OpenXRFbCompositionLayerSecureContentExtensionWrapper());
+	}
+	return singleton;
+}
+
+OpenXRFbCompositionLayerSecureContentExtensionWrapper::OpenXRFbCompositionLayerSecureContentExtensionWrapper() :
+		OpenXRExtensionWrapperExtension() {
+	ERR_FAIL_COND_MSG(singleton != nullptr, "An OpenXRFbCompositionLayerSecureContentExtensionWrapper singleton already exists.");
+
+	request_extensions[XR_FB_COMPOSITION_LAYER_SECURE_CONTENT_EXTENSION_NAME] = &fb_composition_layer_secure_content;
+	singleton = this;
+}
+
+OpenXRFbCompositionLayerSecureContentExtensionWrapper::~OpenXRFbCompositionLayerSecureContentExtensionWrapper() {
+	cleanup();
+}
+
+void OpenXRFbCompositionLayerSecureContentExtensionWrapper::_bind_methods() {
+}
+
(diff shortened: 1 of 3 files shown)


common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
```diff
@@ -0,0 +1,131 @@
...
+Dictionary OpenXRFbCompositionLayerSecureContentExtensionWrapper::_get_requested_extensions() {
+	Dictionary result;
+	for (auto ext : request_extensions) {
+		uint64_t value = reinterpret_cast<uint64_t>(ext.value);
+		result[ext.key] = (Variant)value;
+	}
+	return result;
+}
+
+uint64_t OpenXRFbCompositionLayerSecureContentExtensionWrapper::_set_viewport_composition_layer_and_get_next_pointer(const void *p_layer, const Dictionary &p_property_values, void *p_next_pointer) {
```
m4gr3d, line 79: Here and below, should we check if the extension is available prior to executing the logic, or does the caller validates that?

m4gr3d (changes requested):

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
dsnopek: Ah, good catch! Yes, it should check that here - I'll update the PR when I have a chance

dsnopek: Updated in my latest push!

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
```diff
@@ -0,0 +1,135 @@
...
+			secure_content->flags = XR_COMPOSITION_LAYER_SECURE_CONTENT_REPLACE_LAYER_BIT_FB;
+		} break;
+	};
+
+	return reinterpret_cast<uint64_t>(secure_content);
+}
+
+TypedArray<Dictionary> OpenXRFbCompositionLayerSecureContentExtensionWrapper::_get_viewport_composition_layer_extension_properties() {
+	TypedArray<Dictionary> properties;
+
```
m4gr3d, line 113: We'd want to do a similar `fb_composition_layer_secure_content` check here so that no additional property is added if the extension is not enabled.

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
```diff
@@ -0,0 +1,135 @@
...
+		external_output["hint"] = PROPERTY_HINT_ENUM;
+		external_output["hint_string"] = "Display,Exclude,Replace";
+		properties.push_back(external_output);
+	}
+
+	return properties;
+}
+
+Dictionary OpenXRFbCompositionLayerSecureContentExtensionWrapper::_get_viewport_composition_layer_extension_property_defaults() {
+	Dictionary defaults;
```
m4gr3d, line 127: Similar comment here, no point returning a default if the extension is not enabled.

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
```diff
@@ -0,0 +1,135 @@
...
+	defaults[EXTERNAL_OUTPUT_PROPERTY_NAME] = (int)EXTERNAL_OUTPUT_DISPLAY;
+	return defaults;
+}
+
+void OpenXRFbCompositionLayerSecureContentExtensionWrapper::_on_viewport_composition_layer_destroyed(const void *p_layer) {
+	const XrCompositionLayerBaseHeader *layer = reinterpret_cast<const XrCompositionLayerBaseHeader *>(p_layer);
```
m4gr3d, line 133: Similar comment here about checking the extension is enabled. 
Here it's not as important since we're only trying to remove an inexistent value from the map, but it guards against future logic additions that may extend the current logic, and establish a pattern for other composition layer additions to follow.

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
```diff
@@ -0,0 +1,135 @@
+/**************************************************************************/
+/*  openxr_fb_composition_layer_secure_content_extension_wrapper.cpp      */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Original contributed implementation:                                   */
```
m4gr3d, line 10: Looks like you copied the wrong license header 🙂

common/src/main/cpp/include/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.h
```diff
@@ -0,0 +1,83 @@
...
+/*  openxr_fb_composition_layer_secure_content_extension_wrapper.h        */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Original contributed implementation:                                   */
+/*   Copyright (c) 2024 Malcolm Nixon                                     */
```
m4gr3d, line 11: Looks like you copied the wrong license header 🙂

m4gr3d (changes requested):

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
dsnopek: Nope, we need to return these always - it's for the editor, not runtime. Otherwise, we'd only provide these options when running the editor in VR on a headset that supports this extension.

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
dsnopek: In this case, it's for both the editor and runtime, but because we need this in the editor, we have to return it always.

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
dsnopek: Ok, this one is only for runtime, so we can add the check here :-)

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
dsnopek: Eep! I've been meaning to make a template of the `.h` and `.cpp` files, so I can copy a nice clean template, rather than copying the files from the existing extensions :-)

dsnopek: Fixed in latest push

common/src/main/cpp/include/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.h
dsnopek: Fixed in latest push

common/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
dsnopek: Fixed in latest push

m4gr3d (approved):

dsnopek:
The Godot PR was merged, so taking this out of draft

m4gr3d:
@dsnopek Can you rebase this PR for merging.

dsnopek:
Rebased!

PR #119, opened 2024-04-08
Implement XR_META_passthrough_preferences extension

devloglogan:
~~*Draft PR because this is a follow up PR to #115*~~

Adds an `is_passthrough_preferred()` function to `OpenXRFbPassthroughExtensionWrapper` that will return true if the headset's passthrough feature was enabled on application startup.

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
@@ -57,6 +57,7 @@ OpenXRFbPassthroughExtensionWrapper::OpenXRFbPassthroughExtensionWrapper() :
 
 	request_extensions[XR_FB_PASSTHROUGH_EXTENSION_NAME] = &fb_passthrough_ext;
 	request_extensions[XR_FB_TRIANGLE_MESH_EXTENSION_NAME] = &fb_triangle_mesh_ext;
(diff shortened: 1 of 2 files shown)


common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -607,6 +619,27 @@ bool OpenXRFbPassthroughExtensionWrapper::has_layer_depth_passthrough_capability
...
+
+	XrPassthroughPreferencesMETA passthrough_preferences = {
+		XR_TYPE_PASSTHROUGH_PREFERENCES_META, // type
+		nullptr, // next
+		0, // flags
+	};
+
+	XrResult result = xrGetPassthroughPreferencesMETA(SESSION, &passthrough_preferences);
+	if (XR_FAILED(result)) {
+		UtilityFunctions::print("Failed to get passthrough preferences, error code: ", result);
```
m4gr3d, line 636: Can you update this `print` to `printerr` as discussed in the other PR.

m4gr3d (approved):
Looks good!

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
devloglogan: Updated on latest push!

PR #121, opened 2024-04-15
Add support for developer-created spatial anchors via XR_FB_spatial_entity

dsnopek:
This is a follow-up to PR https://github.com/GodotVR/godot_openxr_vendors/pull/107 which added support for scene anchors (aka the data from Meta's scene understanding).

This PR adds the ability for the developer to create and manage their own spatial anchors.

Here's a screenshot from the demo:

org godotengine openxr vendors demo-20240415-165709

When you switch into passthrough mode, you can press the trigger on a surface from scene understanding (aka a scene anchor) and place a spatial anchor, which is represented by a colored cone (the color is selected randomly). If you point at an existing spatial anchor, pressing the trigger will remove it. If you close and open the app again, the spatial anchors will be restored with the same colors they had previously.

This is done via a new node called `OpenXRFbSpatialAnchorManager` which works a lot like the `OpenXRFbSceneManager` which was added in PR https://github.com/GodotVR/godot_openxr_vendors/pull/107

Here's how it works:

- You can create an anchor by calling `create_anchor(transform: Transform3D, custom_data: Dictionary)` on the node. It won't be created immediately - it's an asynchronous operation. But when it is, the `openxr_fb_spatial_anchor_tracked` signal will be emitted, which will give the developer an `XRAnchor3D` and the `OpenXRFbSpatialEntity` object (with the `custom_data` stashed in it). The `custom_data` can be used to store application-specific data about what the anchor actually is, ie what object the player is actually placing in the world (on the OpenXR side it only has a UUID and a location).
- I suspect a very common case will be creating an instance of a particular scene as a child of the `XRAnchor3D`. So, rather than connecting to the `openxr_fb_spatial_anchor_tracked` signal and instantiating the scene manually, the developer can optionally set the `scene` property, which will be automatically instantiated, and have a method called on it with the `OpenXRFbSpatialEntity`. (This is very similar to how `OpenXRFbSceneManager` works!) Here they can grab the `custom_data` and use it to initialize the scene - the demo uses this to setup the anchor with the correct color.
- ~~By using the `persist_in_local_file` and `local_file_path`, the `OpenXRFbSpatialAnchorManager` node will save the anchors to a JSON file (along with their `custom_data`), and then automatically load them from the file when the application starts up again. While it's possible to query OpenXR for all pre-existing anchors, all you get is the UUIDs and positions, you don't get any information about what the anchor is meant to be in the application. That's why we need to also store these in a JSON file along with the `custom_data`, which is where the developer would put that information.~~
- The `openxr_fb_spatial_anchor_tracked` signal is also emitted for anchors that are loaded (not just created), which allows the developer to handle new and existing anchors at the same time.
- ~~If the developer doesn't want any of the automatic file saving/loading or other automatic features, they can disable those via properties on the node, which gives them the freedom to do something more custom if they want.~~

 **UPDATE:** _The saving/loading from file functionality was removed from the node, and moved into the demo project. Developers will need to implement their own loading/saving, but once they hand the data off to the node, it should take care of everything else._

Here's the properties in the editor:

Selection_149

### DIFF
--- a/common/src/main/cpp/classes/openxr_fb_spatial_anchor_manager.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_spatial_anchor_manager.cpp
@@ -1,0 +1,407 @@
+/**************************************************************************/
+/*  openxr_fb_spatial_anchor_manager.cpp                                  */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "classes/openxr_fb_spatial_anchor_manager.h"
+
+#include <godot_cpp/classes/open_xr_interface.hpp>
+#include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/classes/packed_scene.hpp>
+#include <godot_cpp/classes/xr_origin3d.hpp>
+#include <godot_cpp/classes/xr_anchor3d.hpp>
+#include <godot_cpp/classes/xr_server.hpp>
+
+#include <godot_cpp/variant/utility_functions.hpp>
+
+#include "classes/openxr_fb_spatial_entity_query.h"
+#include "extensions/openxr_fb_spatial_entity_extension_wrapper.h"
+
+using namespace godot;
+
+void OpenXRFbSpatialAnchorManager::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("set_scene", "scene"), &OpenXRFbSpatialAnchorManager::set_scene);
+	ClassDB::bind_method(D_METHOD("get_scene"), &OpenXRFbSpatialAnchorManager::get_scene);
+
+	ClassDB::bind_method(D_METHOD("set_scene_setup_method", "method_name"), &OpenXRFbSpatialAnchorManager::set_scene_setup_method);
+	ClassDB::bind_method(D_METHOD("get_scene_setup_method"), &OpenXRFbSpatialAnchorManager::get_scene_setup_method);
+
+	ClassDB::bind_method(D_METHOD("set_visible", "visible"), &OpenXRFbSpatialAnchorManager::set_visible);
+	ClassDB::bind_method(D_METHOD("get_visible"), &OpenXRFbSpatialAnchorManager::get_visible);
+	ClassDB::bind_method(D_METHOD("show"), &OpenXRFbSpatialAnchorManager::show);
+	ClassDB::bind_method(D_METHOD("hide"), &OpenXRFbSpatialAnchorManager::hide);
+
+	ClassDB::bind_method(D_METHOD("create_anchor", "transform", "custom_data"), &OpenXRFbSpatialAnchorManager::create_anchor, DEFVAL(Dictionary()));
+	ClassDB::bind_method(D_METHOD("load_anchor", "uuid", "custom_data", "location"), &OpenXRFbSpatialAnchorManager::load_anchor, DEFVAL(Dictionary()), DEFVAL(OpenXRFbSpatialEntity::STORAGE_LOCAL));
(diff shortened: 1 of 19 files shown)


common/src/main/cpp/include/extensions/openxr_fb_spatial_entity_extension_wrapper.h
```diff
@@ -53,8 +53,12 @@ class OpenXRFbSpatialEntityExtensionWrapper : public OpenXRExtensionWrapperExten
 		return fb_spatial_entity_ext;
 	}
 
+	typedef void (*SpatialAnchorCreatedCallback)(XrResult p_result, XrSpace p_space, const XrUuidEXT *p_uuid, void *p_userdate);
```
m4gr3d, line 56: typo on `p_userdate`.

common/src/main/cpp/classes/openxr_fb_spatial_anchor_manager.cpp
```diff
@@ -0,0 +1,505 @@
...
+	} else {
+		// Otherwise, we just query the specific anchors we know about.
+		query->query_by_uuid(anchor_data.keys(), OpenXRFbSpatialEntity::STORAGE_LOCAL);
+		query->set_max_results(anchor_data.size());
+	}
+	query->connect("openxr_fb_spatial_entity_query_completed", callable_mp(this, &OpenXRFbSpatialAnchorManager::_on_anchor_load_query_completed).bind(anchor_data, OpenXRFbSpatialEntity::STORAGE_LOCAL, false));
+	query->execute();
+
+	return OK;
+}
```
m4gr3d, line 477: I don't think saving to / loading from a file are functionality that belongs to this class / plugin for several reasons:
- File access differs widely based on the platform. On desktop OSes (e.g: Windows, Linux) it has few limitations whereas on mobile platforms like Android only certain paths are available to write to (`res://` cannot be written to, `user://` can, and only certain `filesystem` paths are available). By providing the functionality in this plugin, we're giving the user the illusion we're smoothing all the platform specific underlying issues, which this PR definitely doesn't, but attempting to do so would deserve its own plugin given the complexity and edge cases involved.
- IO bound operations are typically best handled on a separate thread to avoid blocking the main thread. For a sample like this plugin's demo not concerned with performance, we can perform IO within the same thread, but this is not an approach I'd advocate for more complex apps / games.
- The proposed implementation uses `json` to parse the data; this is fine for a sample like this plugin's demo, but again this is not an approach I'd advocate for more complex apps / games, due to `json` own limitations.

m4gr3d (changes requested):

common/src/main/cpp/classes/openxr_fb_spatial_anchor_manager.cpp
dsnopek: So, are you advocating for leaving this problem entirely up to the developer?

My thinking was that we provide a default in order to reduce the boilerplate for getting started, but it is entirely optional. The developer can turn this off by disabling the `persist_in_local_file` property, and then do their own thing.

Anyway, I can certainly just move all this code to the demo. But then when developers want to add this to their own project, they'll either have to copy that same code, or write something new from scratch, making it more work to get started.

m4gr3d: Yeah, I advocate leaving this up to the developer. Here I use the same approach we use for the Godot proposals; i.e: is this something that the developer can resolve using scripting or is this something that needs to be incorporate into `core`, where `core` here stands for the vendors plugin.

And this definitely falls into the first category, i.e: this is something developers can resolve via scripting. We can provide guidance via the demo, or by putting a gist illustrating this capability.
My expectation is that this would be useful for folks testing the capability, but not when building a full app / game, since in that scenario you'd want the data layer for the app / game to be consolidated (e.g: common approach for storing anchors, user data, game state, etc...).

dsnopek: Alright, that makes sense!

In my next update, I'll move the file bits into the demo.

common/src/main/cpp/include/extensions/openxr_fb_spatial_entity_extension_wrapper.h
dsnopek: Fixed in my latest push

dsnopek:
This is updated per @m4gr3d's requested changes, and rebased on all the recent merges

common/src/main/cpp/extensions/openxr_fb_spatial_entity_extension_wrapper.cpp
```diff
@@ -155,6 +161,51 @@ bool OpenXRFbSpatialEntityExtensionWrapper::_on_event_polled(const void *event)
...
+		{ quat.x, quat.y, quat.z, quat.w }, // orientation
+		{ pos.x, pos.y, pos.z }, // position
+	};
+
+	XrSpatialAnchorCreateInfoFB info = {
+		XR_TYPE_SPATIAL_ANCHOR_CREATE_INFO_FB, // type
+		nullptr, // next
+		reinterpret_cast<XrSpace>(get_openxr_api()->get_play_space()), // space
+		pose, // poseInSpace
+		get_openxr_api()->get_next_frame_time(), // time
```
m4gr3d, line 179: This should use `get_predicted_display_time` following the merging of https://github.com/godotengine/godot/pull/89734

m4gr3d (changes requested):

common/src/main/cpp/classes/openxr_fb_spatial_anchor_manager.cpp
```diff
@@ -0,0 +1,409 @@
...
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "classes/openxr_fb_spatial_anchor_manager.h"
+
+#include <godot_cpp/classes/file_access.hpp>
+#include <godot_cpp/classes/json.hpp>
```
m4gr3d, line 33: Is this include still needed?

m4gr3d, line 33: Are these includes still needed?

common/src/main/cpp/extensions/openxr_fb_spatial_entity_extension_wrapper.cpp
dsnopek: Yes! However, we'd need to update the `extension_api.json` we're using, and there's 4 other unrelated places where we're calling `get_next_frame_time()`. That feels like something for a dedicated PR.

common/src/main/cpp/classes/openxr_fb_spatial_anchor_manager.cpp
dsnopek: Ack, nope, not anymore! Fixed in my latest push

m4gr3d (approved):
Looks great!

common/src/main/cpp/extensions/openxr_fb_spatial_entity_extension_wrapper.cpp
m4gr3d: Sounds good! I've added https://github.com/GodotVR/godot_openxr_vendors/issues/140 to track.

Issue #45, opened 2023-09-09
Add descriptions to tickboxes


BastiaanOlij:
Godot has the ability to add descriptions in the popup for tickboxes, we should make use of this to clarify what each settings is for. May require upstream changes.

![image](https://github.com/GodotVR/godot_openxr_loaders/assets/1945449/9ca373b4-bbc3-4c91-b697-ffcf77aa96e0)

PR #122, opened 2024-04-16
Add support for XR_META_spatial_entity_mesh to get global scene mesh

dsnopek:
Here's a screen shot showing a wire frame (the light white lines) of the global scene mesh, which is what's added to the demo:

org godotengine openxr vendors demo-20240416-175351

This PR also fixes a bug that I noticed in `OpenXRFbSceneManager`,  with storing the scenes based on semantic label (it was inconsistently converting upper to lower case, causing the properties to not really get set).

### DIFF
--- a/common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
@@ -93,7 +93,7 @@ bool OpenXRFbSceneManager::_get(const StringName &p_name, Variant &r_ret) const 
 	if (parts.size() == 2 && parts[0] == "scenes") {
 		const PackedStringArray &semantic_labels = OpenXRFbSceneExtensionWrapper::get_supported_semantic_labels();
 		if (semantic_labels.has(parts[1])) {
-			const Ref<PackedScene> *scene = scenes.getptr(p_name.substr(7));
+			const Ref<PackedScene> *scene = scenes.getptr(parts[1]);
 			r_ret = scene ? Variant(*scene) : Variant();
 			return true;
 		}
@@ -104,7 +104,7 @@ bool OpenXRFbSceneManager::_get(const StringName &p_name, Variant &r_ret) const 
 void OpenXRFbSceneManager::_get_property_list(List<PropertyInfo> *p_list) const {
 	const PackedStringArray &semantic_labels = OpenXRFbSceneExtensionWrapper::get_supported_semantic_labels();
 	for (int i = 0; i < semantic_labels.size(); i++) {
-		p_list->push_back(PropertyInfo(Variant::Type::OBJECT, "scenes/" + semantic_labels[i].to_lower(), PROPERTY_HINT_RESOURCE_TYPE, "PackedScene"));
+		p_list->push_back(PropertyInfo(Variant::Type::OBJECT, "scenes/" + semantic_labels[i], PROPERTY_HINT_RESOURCE_TYPE, "PackedScene"));
 	}
 }
 
@@ -300,12 +300,12 @@ void OpenXRFbSceneManager::_create_scene_anchor(const Ref<OpenXRFbSpatialEntity>
 }
 
 Ref<PackedScene> OpenXRFbSceneManager::get_scene_for_entity(const Ref<OpenXRFbSpatialEntity> &p_entity) const {
-	Array semantic_labels = p_entity->get_semantic_labels();
+	PackedStringArray semantic_labels = p_entity->get_semantic_labels();
 
 	// @todo Allow developers to override which label is selected.
-	Variant selected_label = semantic_labels.size() > 0 ? semantic_labels[0] : Variant();
+	String selected_label = semantic_labels.size() > 0 ? semantic_labels[0] : String();
 
-	if (selected_label.get_type() == Variant::STRING && scenes.has(selected_label)) {
+	if (!selected_label.is_empty() && scenes.has(selected_label)) {
 		return scenes[selected_label];
 	}
 
--- a/common/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
@@ -29,15 +29,18 @@
 
 #include "classes/openxr_fb_spatial_entity.h"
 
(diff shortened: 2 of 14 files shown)


common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
```diff
@@ -300,12 +300,12 @@ void OpenXRFbSceneManager::_create_scene_anchor(const Ref<OpenXRFbSpatialEntity>
 }
 
 Ref<PackedScene> OpenXRFbSceneManager::get_scene_for_entity(const Ref<OpenXRFbSpatialEntity> &p_entity) const {
-	Array semantic_labels = p_entity->get_semantic_labels();
+	PackedStringArray semantic_labels = p_entity->get_semantic_labels();
```
m4gr3d, line 303: Can we instead return `return String(label_data).to_lower().split(",");` in `OpenXRFbSceneExtensionWrapper::get_supported_semantic_labels();` to have the semantic labels in lower case all the time.

Switching back and forth between `to_lower` and `to_upper` in different sections of the code is likely to introduce accidental subtle bugs.

common/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
```diff
@@ -180,10 +184,47 @@ bool OpenXRFbSpatialEntity::is_tracked() const {
 	return OpenXRFbSpatialEntityExtensionWrapper::get_singleton()->is_entity_tracked(uuid);
 }
 
+Array OpenXRFbSpatialEntity::get_triangle_mesh() const {
```
m4gr3d, line 187: Shouldn't this return an `ArrayMesh` instead?

m4gr3d (changes requested):

common/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
dsnopek: No, this method is for if the developer wants the use the actual data for something. An `ArrayMesh` is immediately uploaded to the GPU, and if you want to access the data, Godot would actually have to pull it back down from the GPU again.

By returning an `Array` in the same format as `ArrayMesh` uses, it keeps the data on the CPU side, and allows the developer to inspect it, or manipulate it, possibly in order to construct their own `ArrayMesh` with it. (The demo does this - in order for the wireframe shader to work, we need to alter the mesh data.)

If the developer wanted to just use the unaltered mesh in rendering, then the idea is that they should call `create_mesh_instance()` which will give you a `MeshInstance3D` with an `ArrayMesh` on it.

common/src/main/cpp/classes/openxr_fb_scene_manager.cpp
dsnopek: Sure, we could do that.

My original plan was to keep them in their original upper-cased form, because that's what we get from OpenXR. However, I was surprised to learn that Godot automatically lower-cases properties at some point (I think when saving to the .tscn file?), and so that didn't actually work. :-)

So, I attempted to switch to (1) using lower-case for the scene property names, but (2) upper-case when returning the list of semantic labels (so they matched the spec). But I clearly missed a couple spots that needed to change!

Anyway, if we're cool with just telling devs that we always deal with the semantic labels in lower-case (despite the spec giving them in upper-case), we could do that

m4gr3d: > Anyway, if we're cool with just telling devs that we always deal with the semantic labels in lower-case (despite the spec giving them in upper-case), we could do that

That sounds like a reasonable approach to me!

On the topic of documentation, do you know if GDExtension provides similar documentation mechanism as the core nodes? It'd be handy for developers to be able to access the documentation straight from the editor.

dsnopek: > On the topic of documentation, do you know if GDExtension provides similar documentation mechanism as the core nodes?

There will be a mechanism in Godot 4.3! However, at this moment, it isn't implemented in godot-cpp yet (I have a draft PR that I need to find the time to finish :-))

m4gr3d: Sweet! 
Would that address https://github.com/GodotVR/godot_openxr_vendors/issues/45?

dsnopek: Maybe. It basically allows us to give the editor the same kind of XML docs as we can in Godot itself. But export settings are treated as properties of the `EditorExportPlatform` - I don't know if there's a way for us to pass in some XML that would be added to that? I'd need to investigate a bit

dsnopek: I've updated everywhere in the API where we expose semantic labels to return them in lower-case. This required some changes in the demo. It feels a little weird not matching the names from the spec, but it works!

common/src/main/cpp/extensions/openxr_fb_scene_extension_wrapper.cpp
```diff
@@ -98,7 +98,7 @@ void OpenXRFbSceneExtensionWrapper::_on_instance_destroyed() {
 }
 
 const PackedStringArray &OpenXRFbSceneExtensionWrapper::get_supported_semantic_labels() {
-	static PackedStringArray semantic_labels = String(SUPPORTED_SEMANTIC_LABELS).split(",");
+	static PackedStringArray semantic_labels = String(SUPPORTED_SEMANTIC_LABELS).to_lower().split(",");
```
m4gr3d, line 101: Can you add a comment describing why we're converting to lower case (e.g: to match Godot's automatically converting them to lower-case when saving properties).

m4gr3d (approved):
Looks good!

common/src/main/cpp/extensions/openxr_fb_scene_extension_wrapper.cpp
dsnopek: Sure, I added a comment at both the places where we convert the semantic labels to lower-case

m4gr3d (approved):

m4gr3d:
@dsnopek Can you rebase the PR?

dsnopek:
@m4gr3d Rebased!

PR #123, opened 2024-04-17
Implement XR_META_passthrough_color_lut extension

devloglogan:
~~_Draft PR because this is a follow up PR to #115_~~

This PR adds `XR_META_passthrough_color_lut` support to the `XR_FB_passthrough` extension wrapper. A user creates a valid `OpenXRMetaPassthroughColorLut` instance by passing in an image to `OpenXRMetaPassthroughColorLut.create_from_image()`, and then that instance may be passed to `OpenXRFbPassthroughExtensionWrapper.set_color_lut()` along with a weight. You can also interpolate between two `OpenXRMetaPassthroughColorLut` instances by using `OpenXRFbPassthroughExtensionWrapper.set_interpolated_color_lut()`. The weights on these LUT styles can be tweened to create smooth transitions, here is an example video:

https://github.com/GodotVR/godot_openxr_vendors/assets/95497005/6f80d4d2-3d6b-49cf-8c0a-4a2b4a281b49


common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -607,6 +639,137 @@ bool OpenXRFbPassthroughExtensionWrapper::has_layer_depth_passthrough_capability
...
+
+	if (color_lut_handle == handle) {
+		color_lut_handle = XR_NULL_HANDLE;
+	}
+	if (source_color_lut_handle == handle) {
+		color_lut_handle = XR_NULL_HANDLE;
+	}
+	if (target_color_lut_handle == handle) {
+		color_lut_handle = XR_NULL_HANDLE;
+	}
```
m4gr3d, line 759: Do you mean for those to be `source_color_lut_handle` and `target_color_lut_handle` instead of `color_lut_handle`?

m4gr3d (changes requested):

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
devloglogan: Yes! Fixed in latest push.

m4gr3d (approved):

devloglogan:
Made a small change to this PR; I've reordered the parameters of the `set_color_lut()` and `set_interpolated_color_lut()` such that the weight is the first parameter. This makes it easy for users to tween weights without having to use an intermediate function. Something like:

```
tween.tween_method(fb_pasthrough.set_color_lut.bind(color_lut), 0.0, 1.0, time)
```

Added a tween like this into the demo.

Tagging @m4gr3d since you already approved this PR!

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -607,6 +639,137 @@ bool OpenXRFbPassthroughExtensionWrapper::has_layer_depth_passthrough_capability
...
+
+	current_passthrough_filter = PASSTHROUGH_FILTER_COLOR_MAP_LUT;
+	color_map_lut.colorLut = color_lut_handle;
+	color_map_lut.weight = CLAMP(p_weight, 0.0, 1.0);
+	passthrough_style.next = &color_map_lut;
+
+	if (is_passthrough_started()) {
+		XrResult result = xrPassthroughLayerSetStyleFB(passthrough_layer[current_passthrough_layer], &passthrough_style);
+		if (XR_FAILED(result)) {
+			UtilityFunctions::print("Failed to set passthrough style, error code: ", result);
```
m4gr3d, line 662: nit: let's start making better use of `printerr`. I've taken a quick look at the codebase and we don't seem to use it at all, so not a blocker for this PR, but let's start making use of it in scenarios where it makes sense.

m4gr3d (approved):

common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
dsnopek: In normal engine code, I feel like we'd use `ERR_PRINT()` or `ERR_FAIL_COND()` in this sort of situations. However, at the moment, 'godot_openxr_vendors' has lots of `UtilityFunctions::print()` similar to this (and, actually, even the 'openxr' module in Godot has more `print_line()` for errors than it probably should - I think this pattern probably followed over from there).

m4gr3d: yeah, I've noticed that in the code I write myself for this plugin, I tend to default to `print` no matter the circumstances, but I also think it's because I know how to interpret a message, whether it'd be a verbose output or an error message, when I see one.

This may not be so obvious to end users though, which is where distinguishing the output comes handy. e.g: end users can use the editor's output window to isolate (and report) errors.

devloglogan: Updated to `printerr` in all the places in this commit where I believe it makes sense!

m4gr3d (approved):
Can you rebase the PR?

m4gr3d:
> Rebased!

looks like another one is needed after merging https://github.com/GodotVR/godot_openxr_vendors/pull/122 :)

devloglogan:
> looks like another one is needed after merging #122 :)

Heh, rebased again!

PR #125, opened 2024-04-21
Hand Silhouette Outline shader/material for Meta Hand-tracking Model

DanielSnd:
I created a simple Hand Silhouette Outline shader and material for the Meta Hand-tracking Model (Can be used with adjusted values for other hand models as well, but I only used and tested it on the meta hand model.)

It was inspired by the hand shader found in the meta menu.

handtrackingshader-ezgif com-optimize

The shader works in 2 passes, with the first one taking priority and a second one with lower priority being responsible for the outline. Shader properties are available to customize the fade, colors, opacity and outline width.

image

### DIFF
--- a/demo/assets/hand_outline.gdshader
+++ b/demo/assets/hand_outline.gdshader
@@ -1,0 +1,20 @@
+shader_type spatial;
+
+render_mode cull_front, unshaded;
+uniform vec3 outline_color : source_color = vec3(0.26, 0.34, 0.40);
+uniform float outline_width : hint_range(0.0, 2.5, 0.01) = 1.6;
+uniform float max_opacity : hint_range(0.0, 1.0, 0.01) = 0.95;
+uniform float fade_offset : hint_range(0.0, 0.1, 0.001) = 0.055;
+uniform float fade_sharpness : hint_range(0.0, 35.0, 0.01) = 32.00;
+
+varying vec3 vertex_pos;
+
+void vertex() {
+	VERTEX += NORMAL * (0.001 * outline_width);
+	vertex_pos = VERTEX;
+}
+
+void fragment() {
+	ALBEDO = outline_color;
+	ALPHA = (clamp((vertex_pos.y + fade_offset) * fade_sharpness, 0.00, 0.95));
+}
--- a/demo/assets/hand_silhouette.gdshader
+++ b/demo/assets/hand_silhouette.gdshader
@@ -1,0 +1,18 @@
+shader_type spatial;
+
+render_mode unshaded, cull_disabled;
+uniform vec3 silhouette_color : source_color = vec3(0.071, 0.068, 0.104);
+uniform float max_opacity : hint_range(0.0, 1.0, 0.01) = 0.95;
+uniform float fade_offset : hint_range(0.0, 0.1, 0.001) = 0.055;
+uniform float fade_sharpness : hint_range(0.0, 35.0, 0.01) = 28.00;
+
+varying vec3 vertex_pos;
+
+void vertex() {
+	vertex_pos = VERTEX;
+}
+
+void fragment() {
+	ALBEDO = silhouette_color;
+	ALPHA = (clamp((vertex_pos.y + fade_offset) * fade_sharpness, 0.00, max_opacity));
+}
--- a/demo/assets/hand_silhouette_outline_mat.tres
+++ b/demo/assets/hand_silhouette_outline_mat.tres
@@ -1,0 +1,22 @@
+[gd_resource type="ShaderMaterial" load_steps=4 format=4 uid="uid://bdwh0vc86hsdb"]
+
+[ext_resource type="Shader" path="res://assets/hand_outline.gdshader" id="1_wpjxj"]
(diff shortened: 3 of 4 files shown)


m4gr3d (approved):
Looks pretty cool!

We haven't in the past received direct contributions to the `demo` project but I don't see why that shouldn't be allowed. I'm in favor of merging this in!

cc @BastiaanOlij @dsnopek for additional feedback.

DanielSnd:
> Looks pretty cool!
> 
> We haven't in the past received direct contributions to the `demo` project but I don't see why that shouldn't be allowed. I'm in favor of merging this in!
> 
> cc @BastiaanOlij @dsnopek for additional feedback.

I see, I'm new to contributing to projects so I'm still learning how to best do things. If there's a better place for it I'd be happy to move it :)

BastiaanOlij:
This is super cool for sure. Personally I wonder if we could start putting some more support files in the plugin. This shader would actually be good to just have available. That may be worth for a follow up discussion.
For the demo I'd like to see it in action, this may be nice to combine with Logans work for the meta hand mesh API.

Just nitpicking, but I would save the material as a `.tres` file, not `.material`. I know the IDE suggest material here by default, but if you save it as a tres file it will stay in text which is more friendly for GIT. On export all tres files are converted to their binary format anyway.

DanielSnd:
> Just nitpicking, but I would save the material as a `.tres` file, not `.material`. I know the IDE suggest material here by default, but if you save it as a tres file it will stay in text which is more friendly for GIT. On export all tres files are converted to their binary format anyway.
Makes sense, I personally like .material not for the binary of it, just for the "knowing what it is by looking at the file" aspect of it. I wonder if in the future we can have .tmaterial and other .toptions in godot.
For now could rename it with _material.tres and would achieve the same effect of knowing what it is :)

dsnopek:
Wow, this is really cool! Given how this matches the hands when in the menu, I think other folks will want to use this as well.

> This is super cool for sure. Personally I wonder if we could start putting some more support files in the plugin. This shader would actually be good to just have available.

Yeah, this is tricky. I think this shader will be popular with the Meta hand-tracking mesh extension, and so would make sense to provide as part of this extension. However, it could certainly be useful on other platforms too, including ones that don't need to use 'godot_openxr_vendors' or even non-OpenXR platforms (ie WebXR). Maybe we could include it here _and_ as part of XR Tools, similar to how `viewport_2d_in_3d.tscn` has gotten copied into a couple projects?

devloglogan:
Love it! :) If we do want to merge this in, I'd say we should assign it as the material for the `OpenXRFbHandTrackingMesh` nodes present in the demo scene. Currently, this PR just adds the material/shaders to the asset folder.

DanielSnd:
> Love it! :) If we do want to merge this in, I'd say we should assign it as the material for the `OpenXRFbHandTrackingMesh` nodes present in the demo scene. Currently, this PR just adds the material/shaders to the asset folder.

Perfect, it's my first time collaborating back into a project so I didn't want to make any assumptions and touch the main scene hahaha. Touching the main scene feels important and I didn't want to cause any conflicts or anything.

Should I assign it in the main scene?

devloglogan:
@DanielSnd I would say yes, it should be assigned to the hands in the main scene! And if you change the file format to `.tres` like Bastiaan suggested, I don't think anyone is opposed to merging this?

DanielSnd:
> @DanielSnd I would say yes, it should be assigned to the hands in the main scene! And if you change the file format to `.tres` like Bastiaan suggested, I don't think anyone is opposed to merging this?

Sounds good, I did that and attempted an interactive rebase to squash the changes into one commit as suggested in the godot manual "PR workflow". It looks like it was successful from what I can tell but it's my first time doing that so let me know if there's anything else I should do :)

demo/assets/hand_outline.gdshader
```diff
@@ -0,0 +1,20 @@
+shader_type spatial;
+
+render_mode cull_front,unshaded;
+uniform vec3 outline_color:source_color = vec3(0.26,0.34,0.40);
```
dsnopek, line 4: We usually use more whitespace than this, for example:

```suggestion
render_mode cull_front, unshaded;
uniform vec3 outline_color : source_color = vec3(0.26, 0.34, 0.40);
```

However, I don't think that's super important or should necessarily hold up this PR.

dsnopek (approved):
Thanks!

I just tested this (combined locally with PR #131) and works really great. :-) I love this shader :heart: 

The code changes look fine to me as well. The one comment I would make is that we usually use more whitespace than used in the `.gdshader` files. I'd personally prefer to those to use similar coding standards to our GDScript and C++ files, however, I don't think it should necessarily hold up this PR.

dsnopek:
Unfortunately, it looks like this needs a rebase now that PR https://github.com/GodotVR/godot_openxr_vendors/pull/131 has been merge.

DanielSnd:
Rebased and fixed whitespaces :) Please let me know if there's anything else that needs changing.

dsnopek (approved):
Thanks! I haven't re-tested, but the code changes look great :-)

m4gr3d (approved):

PR #130, opened 2024-04-25
Retire use of the XRHandTracker::Hand

Malcolmnixon:
When `XRHandTracker` was initially developed it had a `Hand` enum _[Left=0, Right=1]_ identify the tracked hand.

When XR Trackers were reorganized into an inheritance-tree, `XRHandTracker` was modified to extend from `XRPositionalTracker` which already has a `hand` property. This resulted in duplicate information, and for C#/.Net a collision between the enum and property symbol names.

~~PR https://github.com/godotengine/godot/pull/91130 has been introduced to remove `XRHandTracker::Hand` from Godot, and this PR removes its use in godot_openxr_vendors.~~

UPDATE: The Godot PR has been merged removing `XRHandTracker::Hand`. This PR now includes the updated extension_api.json including that merge.

### DIFF
--- a/common/src/main/cpp/include/classes/openxr_fb_hand_tracking_mesh.h
+++ b/common/src/main/cpp/include/classes/openxr_fb_hand_tracking_mesh.h
@@ -39,7 +39,11 @@ namespace godot {
 class OpenXRFbHandTrackingMesh : public Node3D {
 	GDCLASS(OpenXRFbHandTrackingMesh, Node3D)
 public:
-	using Hand = XRHandTracker::Hand;
+	enum Hand {
+		HAND_LEFT,
+		HAND_RIGHT,
+		HAND_MAX,
+	};
 
 private:
 	Hand hand = Hand::HAND_LEFT;
@@ -74,5 +78,7 @@ public:
 	float get_scale_override() const;
 };
 } //namespace godot
+
+VARIANT_ENUM_CAST(godot::OpenXRFbHandTrackingMesh::Hand);
 
 #endif
--- a/common/src/main/cpp/include/extensions/openxr_fb_hand_tracking_capsules_extension_wrapper.h
+++ b/common/src/main/cpp/include/extensions/openxr_fb_hand_tracking_capsules_extension_wrapper.h
@@ -42,7 +42,6 @@ class OpenXRFbHandTrackingCapsulesExtensionWrapper : public OpenXRExtensionWrapp
 	GDCLASS(OpenXRFbHandTrackingCapsulesExtensionWrapper, OpenXRExtensionWrapperExtension);
 
 public:
-	using Hand = XRHandTracker::Hand;
 	using HandJoint = XRHandTracker::HandJoint;
 
 	godot::Dictionary _get_requested_extensions() override;
@@ -70,6 +69,12 @@ protected:
 	static void _bind_methods();
 
 private:
+	enum Hand {
+		HAND_LEFT,
+		HAND_RIGHT,
+		HAND_MAX,
+	};
+
 	std::map<godot::String, bool *> request_extensions;
 
 	void cleanup();
--- a/common/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
@@ -71,7 +71,7 @@ void OpenXRFbHandTrackingMesh::set_hand(Hand p_hand) {
 	hand = p_hand;
 }
 
-XRHandTracker::Hand OpenXRFbHandTrackingMesh::get_hand() const {
+OpenXRFbHandTrackingMesh::Hand OpenXRFbHandTrackingMesh::get_hand() const {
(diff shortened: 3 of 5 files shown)


common/src/main/cpp/include/classes/openxr_fb_hand_tracking_mesh.h
```diff
@@ -39,7 +39,11 @@ namespace godot {
 class OpenXRFbHandTrackingMesh : public Node3D {
 	GDCLASS(OpenXRFbHandTrackingMesh, Node3D)
 public:
-	using Hand = XRHandTracker::Hand;
+	enum Hand {
```
m4gr3d, line 42: Can't we instead use `XRPositionalTracker::TrackerHand` instead of introducing new enums?

m4gr3d (changes requested):

common/src/main/cpp/include/classes/openxr_fb_hand_tracking_mesh.h
dsnopek: No, because the point of this enum (and the reason we originally had the 2nd enum in `XRHandTracker`) is that the values line up with the indexes used for the hands. So, this exists to give a name to the magic numbers we're using, that is 0 = left and 1 = right.

But I suppose we could use the custom enum only internally, and then use `XRPositionalTracker::TrackerHand` externally?

Malcolmnixon: Indeed - the code in this repository was using `XRHandTracker::Hand` as both container-size (HAND_MAX=2) and then indexing into that container (HAND_LEFT=0, HAND_RIGHT=1). `XRPositionalTracker::TrackerHand` does not work for this usage.

The introduction of `OpenXRFbHandTrackingMesh::Hand` is an attempt to cause the least amount of code-disruption while retiring the old (soon-to-be-deleted) enum. Switching to `XRPositionalTracker::TrackerHand` would necessitate changing the implementation from container-size / container-index and cause further changes in code which is already being refactored for XRTracker scene-tree changes.

Malcolmnixon: I would suggest after merging this, that https://github.com/GodotVR/godot_openxr_vendors/pull/131 be rebased; then subsequent work on normalizing the hand mesh could be performed on that PR as desired.

dsnopek (approved):
> I would suggest after merging this, that https://github.com/GodotVR/godot_openxr_vendors/pull/131 be rebased; then subsequent work on normalizing the hand mesh could be performed on that PR as desired.

I agree! Let's move forward with this one.

common/src/main/cpp/include/classes/openxr_fb_hand_tracking_mesh.h
Malcolmnixon: @m4gr3d Actually in reviewing https://github.com/GodotVR/godot_openxr_vendors/pull/131 I'm actually thinking it might be better to keep it with a custom enum rather than using `XRPositionalTracker::TrackerHand`. Currently if a user manipulates the hand property they get:

image

If this were switched to the `XRPositionalTracker::TrackerHand` enum they would get an `Unknown` option - something not relevant for a hand-tracking-mesh.

m4gr3d: Sounds good!

common/src/main/cpp/include/extensions/openxr_fb_hand_tracking_capsules_extension_wrapper.h
```diff
@@ -70,6 +69,12 @@ class OpenXRFbHandTrackingCapsulesExtensionWrapper : public OpenXRExtensionWrapp
 	static void _bind_methods();
 
 private:
+	enum Hand {
+		HAND_LEFT,
+		HAND_RIGHT,
+		HAND_MAX,
+	};
```
m4gr3d, line 76: You're reusing `OpenXRFbHandTrackingMesh::Hand` below in `openxr_fb_hand_tracking_mesh_extension_wrapper.h`; is there a reason for not doing the same here?

Also should the `OpenXRFbHandTrackingMesh::Hand` enum be moved to `util.h`?

m4gr3d (changes requested):

Malcolmnixon:
This PR was an attempt to get a PR in ahead of Godot deleting `XRHandTracker::Hand` and with no conflicts with https://github.com/GodotVR/godot_openxr_vendors/pull/131. It wasn't going to be pretty, but would allow the repository to move forwards with Godot 4.3-dev6 compatibility, and for the hand-mesh PR to continue cleanup after rebasing.

At this point the `XRHandTracker::Hand` deletion has already occurred in Godot, and the team has indicated a desire for the code to be implemented cleanly; and so this PR will be deleted and the work will have to be done as part of https://github.com/GodotVR/godot_openxr_vendors/pull/131 rather than occurring extensive merge-conflicts.

PR #131, opened 2024-04-25
Update OpenXRFbHandTrackingMesh node and extension wrapper

devloglogan:
Updating `OpenXRFbHandTrackingMesh` to work with the new `XRTracker` PR: https://github.com/godotengine/godot/pull/90645

The `OpenXRFbHandTrackingMesh` now inherits from `Skeleton3D`. For it to work with `XRHandModifier3D`, it now constructs the skeleton on `NOTIFICATION_POSTINITIALIZE`, and then does the rest of the setup after the `openxr_fb_hand_tracking_mesh_data_fetched` signal.

### DIFF
--- a/common/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
@@ -33,6 +33,7 @@
 
 #include <godot_cpp/classes/xr_hand_modifier3d.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
+#include <godot_cpp/classes/xr_hand_tracker.hpp>
 
 using namespace godot;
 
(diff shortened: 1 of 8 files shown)


Malcolmnixon:
This code uses `XRHandTracker::Hand` which will shortly be removed. The https://github.com/GodotVR/godot_openxr_vendors/pull/130 PR resolves that issue, and I believe should merge mostly seamlessly with these changes.

Malcolmnixon:
I tested this with the latest Godot and https://github.com/GodotVR/godot_openxr_vendors/pull/130 patched on top, and it works fine.

https://github.com/GodotVR/godot_openxr_vendors/assets/1863707/5b0b033b-1f89-4a2d-8415-7ca9bcc2a7b8

Malcolmnixon (approved):
This looks good to me and works well.

m4gr3d (approved):

Malcolmnixon:
The `XRHandTracker::Hand` type has already been removed from Godot. As such the https://github.com/GodotVR/godot_openxr_vendors/pull/130 has been abandoned, as improving it to be architecturally clean would result in merge conflicts with this PR.

As such this PR should probably be enhanced to update to the latest Godot `extension_api.json` and then replace all remaining uses of `XRHandTracker::Hand` with `XRPositionalTracker::TrackerHand` (rather than introducing a new enum and potential type-incompatibility).

Using `XRPositionalTracker::TrackerHand` will result in larger changes as that type cannot be used as container-sizing and container-indexing.

devloglogan:
I'm _really_ not a fan of using `XRPositionalTracker::TrackerHand`, my vote is just for making another enum, but perhaps I'm missing the negatives to that @Malcolmnixon? In my latest updates I've added a `Hand` enum in `OpenXRFbHandTrackingMesh`. Adding the enum into `utils.h` is possible but the enum has to be passed to `VARIANT_ENUM_CAST` at some point, and I'm not sure where it makes sense to do that.

I've removed the use of the `Hand` enum from the fb hand tracking capsules extension wrapper, and just created a `HAND_MAX` constant in its place, since that's all that is needed.

And I've updated `extension_api.json`.

Malcolmnixon:
> I'm not sure where it makes sense to do that.

I'm not certain of the best place for this either - Perhaps @dsnopek or @BastiaanOlij might be able to help - they're far more familiar with GDExtension.

I personally prefer the custom enum without the `Unknown` value which makes no sense for hand-specific nodes, but I can also understand the desire to minimize the number of different enums expressing basically the same information. Perhaps having more eyes on it can result in a good alternative.

BastiaanOlij:
> I'm not certain of the best place for this either - Perhaps @dsnopek or @BastiaanOlij might be able to help - they're far more familiar with GDExtension.

I think it's a hard one, on one side it doesn't make sense to have the unknown status, on the other hand it doesn't make sense to have two enums for the same purpose. If only we had the foresight to make UNKNOWN a special value, but we're a few years to late for that.

We can also just ignore the enum for our hand mesh, and just do a normal int and only have the predefined values in the property definition.

All this just because our C# implementation is so strict... sigh :)

dsnopek:
Personally, I'd be for adding the new enum. We have type checking to prevent users from using values from the wrong enum, so I don't think having a 2nd enum is really going to cause that much confusion.

Users will mostly set this through the editor anyway, and if you do set it from code, I don't think it'll be surprising to do something like:

```
# The enum value comes from the class that I already know I'm using. Simple!
hand_mesh.hand = OpenXRFbHandTrackingMesh.HAND_LEFT
```

In fact, I think it's slightly more confusing to have to grab `XRPositionalTracker.TRACKER_HAND_LEFT` from another place altogether, and from a class that users don't usually interact with.

dsnopek (approved):
I just re-tested and the latest version is still working great for me! The code looks good to me, and I personally like where this has ended up with the `Hand` enum.

m4gr3d (approved):

m4gr3d:
The consensus is to go with the new enum, so let's use that approach.

PR #139, opened 2024-05-07
Implement hole punch functionality for `OpenXRFbPassthroughGeometry` node

devloglogan:
In a similar fashion to https://github.com/godotengine/godot/pull/91485, implements "hole punching" for fb passthrough geometry. This mode makes it so objects behind the passthrough geometry will not be seen, like [the clip in this comment](https://github.com/godotengine/godot/pull/91096#issuecomment-2074150150). The passthrough geometry in this repo's demo now has this enabled.

I believe this mode would be the more common use case for passthrough geometry, so I've set `enable_hole_punch` to `true` by default.

### DIFF
--- a/common/src/main/cpp/classes/openxr_fb_passthrough_geometry.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_passthrough_geometry.cpp
@@ -32,10 +32,20 @@
 #include "extensions/openxr_fb_passthrough_extension_wrapper.h"
 
 #include <godot_cpp/classes/engine.hpp>
+#include <godot_cpp/classes/shader.hpp>
+#include <godot_cpp/classes/shader_material.hpp>
 #include <godot_cpp/classes/standard_material3d.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
 using namespace godot;
+
+const Color PREVIEW_COLOR = Color(1.0, 0.0, 1.0);
+static const char *HOLE_PUNCH_SHADER_CODE =
+		"shader_type spatial;\n"
+		"render_mode blend_mix, depth_draw_opaque, cull_back, shadow_to_opacity, shadows_disabled;\n"
+		"void fragment() {\n"
+		"\tALBEDO = vec3(0.0, 0.0, 0.0);\n"
+		"}\n";
 
 void OpenXRFbPassthroughGeometry::set_mesh(const Ref<Mesh> &p_mesh) {
 	if (p_mesh == mesh) {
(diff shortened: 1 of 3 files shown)


common/src/main/cpp/classes/openxr_fb_passthrough_geometry.cpp
```diff
@@ -91,6 +121,36 @@ void OpenXRFbPassthroughGeometry::destroy_passthrough_geometry() {
 		OpenXRFbPassthroughExtensionWrapper::get_singleton()->destroy_geometry_instance(geometry_instance);
 		geometry_instance = XR_NULL_HANDLE;
 	}
+
+	if (opaque_mesh != nullptr) {
+		opaque_mesh->queue_free();
```
m4gr3d, line 126: Should the opaque mesh be removed from the tree when it's destroyed?

devloglogan: I don't see why not, I've added `remove_child(opaque_mesh)` in my latest push to the instances where it's destroyed.

common/src/main/cpp/classes/openxr_fb_passthrough_geometry.cpp
```diff
@@ -49,27 +59,20 @@ void OpenXRFbPassthroughGeometry::set_mesh(const Ref<Mesh> &p_mesh) {
 	mesh = p_mesh;
 
 	if (mesh.is_null()) {
-		if (preview_mesh != nullptr) {
-			preview_mesh->queue_free();
-			preview_mesh = nullptr;
+		if (opaque_mesh != nullptr) {
+			remove_child(opaque_mesh);
+			opaque_mesh->queue_free();
+			opaque_mesh = nullptr;
```
m4gr3d, line 65: This code block is repeated multiple times; can you wrap it into a `delete_opaque_mesh` helper function.

common/src/main/cpp/classes/openxr_fb_passthrough_geometry.cpp
```diff
@@ -49,27 +59,20 @@ void OpenXRFbPassthroughGeometry::set_mesh(const Ref<Mesh> &p_mesh) {
 	mesh = p_mesh;
 
 	if (mesh.is_null()) {
```
m4gr3d, line 61: There was a comment made in https://github.com/godotengine/godot/pull/91247 about how `Ref::is_null()` is different from `!Ref::is_valid()`, example being that if the reference is freed via `queue_free()`, it's no longer valid but it's not `null`.

In the above assignment, we should check if `p_mesh.is_valid()` before assigning it to `mesh`, and set `mesh` to `null` if it's not.

common/src/main/cpp/classes/openxr_fb_passthrough_geometry.cpp
```diff
@@ -91,6 +123,37 @@ void OpenXRFbPassthroughGeometry::destroy_passthrough_geometry() {
...
+
+	if (opaque_mesh != nullptr) {
+		remove_child(opaque_mesh);
+		opaque_mesh->queue_free();
+		opaque_mesh = nullptr;
+	}
+}
+
+void OpenXRFbPassthroughGeometry::instatiate_opaque_mesh() {
+	opaque_mesh = memnew(MeshInstance3D);
```
m4gr3d, line 135: Should add an error condition if `opaque_mesh` is not null, and one if `!mesh.is_valid()`.

m4gr3d (changes requested):

common/src/main/cpp/classes/openxr_fb_passthrough_geometry.cpp
dsnopek: The linked issue was specifically about `Callable`. In the case of `Ref`, there is no difference between `is_null()` and `is_valid()` other than them being boolean opposites.

See: https://github.com/godotengine/godot/blob/master/core/object/ref_counted.h#L201

m4gr3d: I'd assume that `Callable` was also a `Ref` but that's not the case.. thanks for the clarification!

common/src/main/cpp/classes/openxr_fb_passthrough_geometry.cpp
devloglogan: Added!

m4gr3d (approved):

PR #136, opened 2024-05-01
Implement XR_FB_composition_layer_settings extension wrapper

devloglogan:
Adds composition layer filtering properties to `OpenXRCompositionLayer` nodes, allowing users to adjust supersampling and sharpening settings. These settings can improve the image quality of composition layers. More info [here](https://developer.oculus.com/documentation/native/android/mobile-openxr-composition-layer-filtering/). Planning to add auto filtering soon as well, which requires the XR_META_automatic_layer_filter extension.

composition_layer_settings

PR #142, opened 2024-05-08
Implement XR_META_automatic_layer_filter extension

devloglogan:
This PR is a follow up to https://github.com/GodotVR/godot_openxr_vendors/pull/136. The previous settings have now been moved into a "manual" subcategory, while this PR adds new properties related to the "auto" subcategory. Set `enable_auto_filter` to `true`, and within `options` select the candidate filters you would like it to be able to select from.

If `enable_auto_filter` is `true` and at least one option is selected, automatic filtering will override the manual options.

composition_layer_auto_filtering

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_composition_layer_settings_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_composition_layer_settings_extension_wrapper.cpp
@@ -34,8 +34,10 @@
 
 using namespace godot;
 
-static const char *SUPERSAMPLING_MODE_PROPERTY_NAME = "XR_FB_composition_layer_settings/supersampling_mode";
-static const char *SHARPENING_MODE_PROPERTY_NAME = "XR_FB_composition_layer_settings/sharpening_mode";
+static const char *SUPERSAMPLING_MODE_PROPERTY_NAME = "XR_FB_composition_layer_settings/manual/supersampling_mode";
+static const char *SHARPENING_MODE_PROPERTY_NAME = "XR_FB_composition_layer_settings/manual/sharpening_mode";
+static const char *ENABLE_AUTO_FILTER_PROPERTY_NAME = "XR_FB_composition_layer_settings/auto/enable_auto_filter";
+static const char *AUTO_OPTIONS_PROPERTY_NAME = "XR_FB_composition_layer_settings/auto/options";
 
 OpenXRFbCompositionLayerSettingsExtensionWrapper *OpenXRFbCompositionLayerSettingsExtensionWrapper::singleton = nullptr;
 
@@ -51,6 +53,8 @@ OpenXRFbCompositionLayerSettingsExtensionWrapper::OpenXRFbCompositionLayerSettin
 	ERR_FAIL_COND_MSG(singleton != nullptr, "An OpenXRFbCompositionLayerSettingsExtensionWrapper singleton already exists.");
 
 	request_extensions[XR_FB_COMPOSITION_LAYER_SETTINGS_EXTENSION_NAME] = &fb_composition_layer_settings;
+	request_extensions[XR_META_AUTOMATIC_LAYER_FILTER_EXTENSION_NAME] = &meta_automatic_layer_filter;
+
 	singleton = this;
 }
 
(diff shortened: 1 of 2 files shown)


common/src/main/cpp/extensions/openxr_fb_composition_layer_settings_extension_wrapper.cpp
```diff
@@ -100,6 +105,13 @@ uint64_t OpenXRFbCompositionLayerSettingsExtensionWrapper::_set_viewport_composi
 
 	settings->layerFlags = 0;
 
+	if (meta_automatic_layer_filter && p_property_values.get(ENABLE_AUTO_FILTER_PROPERTY_NAME, false) && (int)p_property_values.get(AUTO_OPTIONS_PROPERTY_NAME, 0)) {
+		settings->layerFlags |= XR_COMPOSITION_LAYER_SETTINGS_AUTO_LAYER_FILTER_BIT_META;
+		settings->layerFlags |= (int)p_property_values.get(AUTO_OPTIONS_PROPERTY_NAME, 0);
+
+		return reinterpret_cast<uint64_t>(settings);
```
m4gr3d, line 112: I imagine we return here because the auto options override the manual ones; can you add a comment describing that behavior.

common/src/main/cpp/extensions/openxr_fb_composition_layer_settings_extension_wrapper.cpp
```diff
@@ -153,12 +165,31 @@ TypedArray<Dictionary> OpenXRFbCompositionLayerSettingsExtensionWrapper::_get_vi
...
+		enable_auto_filter["hint"] = PROPERTY_HINT_NONE;
+		properties.push_back(enable_auto_filter);
+	}
+
+	{
+		Dictionary auto_options;
+		auto_options["name"] = AUTO_OPTIONS_PROPERTY_NAME;
+		auto_options["type"] = Variant::INT;
+		auto_options["hint"] = PROPERTY_HINT_FLAGS;
+		auto_options["hint_string"] = "Normal Supersampling,Quality Supersampling,Normal Sharpening,Quality Sharpening";
```
m4gr3d, line 181: What happens if both the `normal` and `quality` option variants are selected?

devloglogan: This is a good question, I've been thinking that it could swap between all possibilities, but the spec isn't really explicit on this.

Main issue is I'm not able to visually distinguish when a normal filter vs. a quality one is being applied. But whether just one or both of the options are selected, I'm definitely able to notice the filters swap from disabled to enabled, and I don't see why the filter wouldn't automatically swap between normal and quality as well. I just don't have a way of knowing which of the two is being applied.

The one bit from the spec on passing options:
> A candidate pool of preferred layer filtering methods from [XrCompositionLayerSettingsFlagBitsFB](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XrCompositionLayerSettingsFlagBitsFB) must be passed along with XR_COMPOSITION_LAYER_SETTINGS_AUTO_LAYER_FILTER_BIT_META. The runtime may apply the appropriate filter when rendering the layer.

dsnopek: My reading of the spec is that if you give both `normal` and `quality` then it will include both of those in its list of potential options that auto could switch between.

But like Logan, when testing the "manual" settings, I can't tell visually when quality is enabled either.

Reading [this article](https://developer.oculus.com/documentation/native/android/mobile-openxr-composition-layer-filtering/), it makes it sound like the `quality` options are only a little better visually, but take a lot more resources.

common/src/main/cpp/extensions/openxr_fb_composition_layer_settings_extension_wrapper.cpp
devloglogan: Added a comment above the if statement in latest push!

m4gr3d (approved):

common/src/main/cpp/extensions/openxr_fb_composition_layer_settings_extension_wrapper.cpp
m4gr3d: From the [Enable Layer Filtering](https://developer.oculus.com/documentation/native/android/mobile-openxr-composition-layer-filtering/#enable-layer-filtering) section, they mention the followings:
> If both bits are set, the compositor defaults to the normal supersampling method.

> If both bits are set, the compositor defaults to the normal sharpening method.

So the current approach should be good to go!
Let me know if I should go ahead and merge the changes.

m4gr3d: @devloglogan Let me know if I can go ahead and merge this PR.

PR #144, opened 2024-05-13
Cleanup: Update hand material resource format

devloglogan:
When opening up the demo project on godot master, the hand material resource will always be updated to `format=3`.


dsnopek (approved):
Thanks!

This looks good to me. Will be nice to not have this turning up in every diff. :-)

m4gr3d (approved):

PR #147, opened 2024-05-13
Update hand mesh data fetch process

devloglogan:
@dsnopek informed me that, in the xr simulator, the demo project would fail to retrieve hand meshes and proceed to keep trying every frame. This would cause a constant flow of error messages. Now when attempting to fetch hand mesh data, it will abort the process if `xrGetHandMeshFB()` returns `XR_ERROR_FUNCTION_UNSUPPORTED` or `XR_ERROR_FEATURE_UNSUPPORTED`.

I've also updated it so after all the hand mesh data is successfully fetched, `should_fetch_hand_mesh_data` is set to `false`.

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
@@ -134,9 +134,21 @@ void OpenXRFbHandTrackingMeshExtensionWrapper::_on_process() {
 				if (fetch_hand_mesh_data(Hand(i))) {
 					emit_signal("openxr_fb_hand_tracking_mesh_data_fetched", i);
 				}
+
+				// May be turned false if fetch_hand_mesh_data encounters errors indicating hand tracking mesh is not supported.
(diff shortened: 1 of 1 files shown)


common/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -134,9 +134,20 @@ void OpenXRFbHandTrackingMeshExtensionWrapper::_on_process() {
 				if (fetch_hand_mesh_data(Hand(i))) {
 					emit_signal("openxr_fb_hand_tracking_mesh_data_fetched", i);
 				}
+
+				if (!should_fetch_hand_mesh_data) {
+					return;
+				}
```
m4gr3d, line 140: Can you add a comment explaining that `fetch_hand_mesh_data` may change the value of `should_fetch_hand_mesh_data` hence why we're checking here again.

devloglogan: Added!

m4gr3d (approved):

PR #148, opened 2024-05-15
Add settings for HTC

BastiaanOlij:
This adds a number of setting for HTC to the export profiles:

image

I've added a subsection for HTC here under a header for Khronos as we'll be adding settings for other platforms in a similar fashion as more and more headsets support the Khronos loader.

Also in the case of HTC these settings are simply `No/Yes`, not `None,Optional,Required`. Unlike with Meta, HTC doesn't gate functionality on these feature flags, these feature flags are purely used by the store to inform the user what features the game makes use of.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 - Add XR_FB_hand_tracking_capsules extension wrapper
 - Add OpenXRFbPassthroughGeometry node
 - Add OpenXRMetaPassthroughColorLut
+- Add feature flags to Khronos loader for HTC
 
 ## 2.0.3
 - Migrate the export scripts from gdscript to C++ via gdextension
--- a/common/src/main/cpp/export/khronos_export_plugin.cpp
+++ b/common/src/main/cpp/export/khronos_export_plugin.cpp
@@ -48,9 +48,102 @@ void KhronosEditorPlugin::_exit_tree() {
 
 KhronosEditorExportPlugin::KhronosEditorExportPlugin() {
 	set_vendor_name(KHRONOS_VENDOR_NAME);
+
+	_hand_tracking_option = _generate_export_option(
+			"khronos_xr_features/htc/hand_tracking",
+			"",
+			Variant::Type::INT,
+			PROPERTY_HINT_ENUM,
+			"No,Yes",
+			PROPERTY_USAGE_DEFAULT,
+			MANIFEST_FALSE_VALUE,
+			false);
+	_tracker_option = _generate_export_option(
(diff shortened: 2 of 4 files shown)


BastiaanOlij:
**We** do gate the eye tracking feature on Android and don't enable it unless the feature flag is there and permission is given.
I've added the feature flag support here, but I'm missing the permission part.

@m4gr3d is this a matter of implementing `supportsFeature` for Khronos and just returning true on the feature tag and not checking permission? And how are we going to deal with this long term as multiple headsets use the Khronos loader but only **some** require permission, and each of those have their own permission string?

@Malcolmnixon can you do some more testing here seeing you have the eye tracker?

m4gr3d (approved):
The logic looks good.

m4gr3d:
> **We** do gate the eye tracking feature on Android and don't enable it unless the feature flag is there and permission is given. I've added the feature flag support here, but I'm missing the permission part.
> 
> @m4gr3d is this a matter of implementing `supportsFeature` for Khronos and just returning true on the feature tag and not checking permission? 

Yes that's correct. 

> And how are we going to deal with this long term as multiple headsets use the Khronos loader but only **some** require permission, and each of those have their own permission string?
> 

We'll need to check the Android build configs at runtime which provide information on the headset brand and manufacturer. Depending on these values, we can select the correct permission string or omit permission checking as done here.

dsnopek (approved):
I did some limited testing: mainly, that the editor plugin part works in how it affects the `AndroidManifest.xml`. I did export and run on the HTC Vive XR Elite, and everything seemed to be working fine, but Bastiaan says the manifest changes only affect the store, not the runtime operation of the headset. The one exception being the permission for eye tracking, however, I don't have the eye tracking accessory, so I can't test that.

The code looks good to me! The bits added to the `AndroidManifest.xml` match the information from HTC.

BastiaanOlij:
> The one exception being the permission for eye tracking, however, I don't have the eye tracking accessory, so I can't test that.

Malcolm should be able to test that, but important here is that on HTC there is no permission, but we require one because other platforms require one. That is still an issue to solve long term as I think Magic Leap requires a permission here, and definately Meta does and I think they now officially support the Khronos loader.

So long term we're going to need to have a discussion on how we're going to enhance this code so we somehow magically know if the device our app gets installed on, requires a permission check, and which permission check that is. Especially in the scenario where our users deploy on say, itch.io, and we have no idea what devices our app end up being run on (possibly a device that didn't even exist when our app was written).

m4gr3d (approved):

PR #151, opened 2024-05-16
Add support for HTC passthrough

BastiaanOlij:
This PR adds support for HTCs version of passthrough in a way that is transparent to the user.
Here we also react to our blend mode.

I did have to add some extra checks to the Meta implementation because while the HTC implementation became active, Metas implementation wasn't checking whether it was actually active and just started error spamming as a result :P

Note that there is an interesting difference here as well, while Meta explicitly requires us to start and stop passthrough, HTC doesn't. We do need to create a passthrough handle once the session is setup, but from then on passthrough is purely switched on whether the passthrough composition layer is added.

I am making the assumption here, and I think that's a fairly safe assumption, that no device is going to support both Metas and HTCs passthrough extensions. If this does become an issue we may need to do something like `set_emulate_environment_blend_mode_alpha_blend` returning an error when another extension has already claimed it.

Tested this on my HTC Elite XR and seems to work fine. 

Note, thanks to @seichter (PR https://github.com/godotengine/godot/pull/86407) for doing much of the ground work. Made it really easy to implement this in the vendor plugin.

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
@@ -201,6 +201,10 @@ void OpenXRFbPassthroughExtensionWrapper::_on_session_created(uint64_t p_session
 }
 
 void OpenXRFbPassthroughExtensionWrapper::_on_state_ready() {
+	if (!fb_passthrough_ext) {
+		return;
+	}
+
 	XRInterface::EnvironmentBlendMode blend_mode = get_blend_mode();
 	if (blend_mode == XRInterface::XR_ENV_BLEND_MODE_ALPHA_BLEND) {
 		start_passthrough_layer(LAYER_PURPOSE_RECONSTRUCTION);
@@ -208,6 +212,10 @@ void OpenXRFbPassthroughExtensionWrapper::_on_state_ready() {
 }
 
 void OpenXRFbPassthroughExtensionWrapper::_on_process() {
+	if (!fb_passthrough_ext) {
+		return;
+	}
+
 	XRInterface::EnvironmentBlendMode blend_mode = get_blend_mode();
 
 	// Reconstruction layer will always take priority
@@ -221,6 +229,10 @@ void OpenXRFbPassthroughExtensionWrapper::_on_process() {
 }
 
 bool OpenXRFbPassthroughExtensionWrapper::_on_event_polled(const void *p_event) {
+	if (!fb_passthrough_ext) {
+		return false;
+	}
+
 	if (static_cast<const XrEventDataBuffer *>(p_event)->type == XR_TYPE_EVENT_DATA_PASSTHROUGH_STATE_CHANGED_FB) {
 		XrEventDataPassthroughStateChangedFB *passthrough_event = (XrEventDataPassthroughStateChangedFB *)p_event;
 		switch (passthrough_event->flags) {
--- a/common/src/main/cpp/extensions/openxr_htc_passthrough_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_htc_passthrough_extension_wrapper.cpp
@@ -1,0 +1,160 @@
+/**************************************************************************/
+/*  openxr_htc_passthrough_extension_wrapper.cpp                          */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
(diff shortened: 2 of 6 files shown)


common/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -202,43 +202,49 @@ void OpenXRFbPassthroughExtensionWrapper::_on_session_created(uint64_t p_session
 }
 
 void OpenXRFbPassthroughExtensionWrapper::_on_state_ready() {
-	XRInterface::EnvironmentBlendMode blend_mode = get_blend_mode();
-	if (blend_mode == XRInterface::XR_ENV_BLEND_MODE_ALPHA_BLEND) {
-		start_passthrough_layer(LAYER_PURPOSE_RECONSTRUCTION);
+	if (fb_passthrough_ext) {
```
m4gr3d, line 205: nit - to minimize visual noise due to the added indentation, we can instead do:

```
if (!fb_passthrough_ext) { 
    return;
}
```

BastiaanOlij: yeah that makes sense

common/src/main/cpp/include/extensions/openxr_htc_passthrough_extension_wrapper.h
```diff
@@ -0,0 +1,123 @@
+/**************************************************************************/
+/*  openxr_htc_passthrough_extension_wrapper.h                            */
+/**************************************************************************/
+/*                         This file is part of:                          */
+/*                             GODOT ENGINE                               */
+/*                        https://godotengine.org                         */
+/**************************************************************************/
+/* Copyright (c) 2014-present Godot Engine contributors (see AUTHORS.md). */
+/* Copyright (c) 2007-2014 Juan Linietsky, Ariel Manzur.                  */
```
m4gr3d, line 9: This section needs to be updated to reflect the style used in this repo.
Or are you leaving it like this because it originated from the Godot repo?

common/src/main/cpp/extensions/openxr_htc_passthrough_extension_wrapper.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+	return result;
+}
+
+void OpenXRHtcPassthroughExtensionWrapper::_on_instance_created(uint64_t p_instance) {
+	XrInstance instance = (XrInstance)p_instance;
+	if (htc_passthrough_ext) {
+		bool result = initialize_htc_passthrough_extension(instance);
+		if (!result) {
+			UtilityFunctions::print("Failed to initialize htc_passthrough extension");
```
m4gr3d, line 85: Should use `printerr`

common/src/main/cpp/extensions/openxr_htc_passthrough_extension_wrapper.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+		// Note: With HTC Passthrough, we create our handle and just switch passthrough
+		// by providing, or not providing, a passthrough composition layer.
+		XrSession session = (XrSession)p_session;
+
+		const XrPassthroughCreateInfoHTC create_info = {
+			XR_TYPE_PASSTHROUGH_CREATE_INFO_HTC,
+			nullptr,
+			XR_PASSTHROUGH_FORM_PLANAR_HTC
+		};
```
m4gr3d, line 110: Can you add the struct field names as comments on the line.

common/src/main/cpp/extensions/openxr_htc_passthrough_extension_wrapper.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+		XrResult result = xrCreatePassthroughHTC(session, &create_info, &passthrough_handle);
+		if (XR_FAILED(result)) {
+			UtilityFunctions::print("Failed to create passthrough");
```
m4gr3d, line 114: Should use `printerr`

common/src/main/cpp/extensions/openxr_htc_passthrough_extension_wrapper.cpp
```diff
@@ -0,0 +1,161 @@
...
+		get_openxr_api()->register_composition_layer_provider(this);
+		get_openxr_api()->set_emulate_environment_blend_mode_alpha_blend(true);
+	}
+}
+
+void OpenXRHtcPassthroughExtensionWrapper::_on_session_destroyed() {
+	if (htc_passthrough_ext && passthrough_handle != XR_NULL_HANDLE) {
+		XrResult result = xrDestroyPassthroughHTC(passthrough_handle);
+		if (XR_FAILED(result)) {
+			UtilityFunctions::print("Unable to destroy passthrough feature");
```
m4gr3d, line 128: Should use `printerr`

common/src/main/cpp/include/extensions/openxr_htc_passthrough_extension_wrapper.h
BastiaanOlij: I copied it from the FB one, so I guess we haven't changed it in all classes?

m4gr3d (approved):
Looks good!

common/src/main/cpp/include/extensions/openxr_htc_passthrough_extension_wrapper.h
m4gr3d: We forgot to do so when migrating the class over. Can you update the fb one as well.

dsnopek (approved):
I tested on the HTC Vive XR Elite and it worked fine! The code changes look good to me as well. Great work!

dsnopek:
Looks like this needs a rebase on conflicts in `CHANGES.md`

PR #154, opened 2024-05-21
Generate normals and tangets for Meta's global scene mesh

dsnopek:
@Nitwel brought up on Discord that `OpenXRFbSpatialEntity.create_mesh_instance()` for the global scene mesh won't have normals or tangents and hence can't be correctly lit.

Since developers can access the raw data we get from the API via `OpenXRFbSpatialEntity.get_triangle_mesh()`, I think doing this extra bit of processing in `create_mesh_instance()` makes sense.

(I had thought this would maybe allow us to remove the "deindexing" that we're doing in `scene_global_mesh.gd`, because `SurfaceTool` has to deindex the mesh in order to generate the normals, however, it will automatically re-index it afterwards. I had considered purposefully deindexing it, but that increases the size of the mesh data by a lot, and I don't know if most folks will end up needing that anyway. So, I just stuck with what we previously had in `scene_global_mesh.gd` for now.)

### DIFF
--- a/common/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
+++ b/common/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
@@ -36,6 +36,7 @@
 #include <godot_cpp/classes/concave_polygon_shape3d.hpp>
 #include <godot_cpp/classes/mesh_instance3d.hpp>
 #include <godot_cpp/classes/plane_mesh.hpp>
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):
LGTM, had to double check where `get_triangle_mesh` gets its data from but it's coming directly from the extension which is good.

Generating normals and tangents early here, considering the simplicity of the geometry, should have very little overhead.

m4gr3d (approved):

PR #155, opened 2024-05-21
Create fb passthrough sample project

devloglogan:
Removes the fb passthrough extension material from the repos demo project and adds a new `samples` directory, along with a new standalone project demonstrating the passthrough functionality in both a simpler and more complete manner.

Marking this as a draft for the moment because the gradle build process isn't working perfectly yet.


SConstruct
```diff
@@ -19,32 +19,35 @@ sources += Glob("#common/src/main/cpp/export/*.cpp")
 sources += Glob("#common/src/main/cpp/extensions/*.cpp")
 sources += Glob("#common/src/main/cpp/classes/*.cpp")
 
-binary_path = '#demo/addons/godotopenxrvendors/.bin'
+binary_paths = ['#demo/addons/godotopenxrvendors/.bin', '#samples/fb-passthrough-sample/addons/godotopenxrvendors/.bin']
```
dsnopek, line 22: Rather than looping like this, it probably makes sense to have scons copy the resulting library. Otherwise, you're basically asking scons to do the linking process multiple times (and linking is the slowest part of the build, at least for me).

So, I'd imagine at the end of the `SConstruct` having something like:

```
samples  = ['samples/fb-passthrough-sample']
for sample in samples:
  sample_library = env.Install(target=f'#{sample}/addons/godotopenxrvendors/.bin', source=library)
  Default(sample_library)
```

Then it's only linked once, and copied for each sample.

dsnopek (commented):
Thanks!

@m4gr3d would know better about the Gradle changes, but I do have some notes regarding the SCons changes.

For some reason, geometry passthrough isn't working for me with this version (whereas it was working on an earlier version I tested before the PR). When using plain "Geometry", I see nothing (the box is completely invisible). And when using "Geometry (Hole Punch)", I see a block box. I'm not sure if I've messed something on my end, but I didn't really change anything from when I tested it before.

Also, I still think there should be a "buffer" on the left/right side of the slider UI. Like the `0.1`m on each side should register as fully to the left or right, and the ball indicating the position should remain within `0.1`m from the edge. I think this would make it easier to set the sliders fully to one side, where it's really tricky to do presently.

SConstruct
devloglogan: Makes sense, I've applied this change!

devloglogan:
> For some reason, geometry passthrough isn't working for me with this version (whereas it was working on an earlier version I tested before the PR). When using plain "Geometry", I see nothing (the box is completely invisible). And when using "Geometry (Hole Punch)", I see a block box. I'm not sure if I've messed something on my end, but I didn't really change anything from when I tested it before.

Strange, for me the plain geometry works just fine. But now I'm noticing that hole punched geometry is a bit more noticeably "out of sync" (the opaque mesh is moving ahead of the passthrough geometry instance, so they don't line up). Zero clue what could be causing these problems right now but I'll think on it further.

> Also, I still think there should be a "buffer" on the left/right side of the slider UI.

Ah, I'd added that "buffer" suggestion, but in the form that it would instead snap to the extreme ends of the slider. I've adjusted it to be more in line with your suggestion now.

SConstruct
devloglogan: Looks like the `Install` method wasn't copying over files with their complete filepaths, in latest push I've amended `target` to fix that.

devloglogan:
I've updated the gradle build files and everything should now be in working condition, so I'm going to open this PR!

m4gr3d (changes requested):
I don't think we need to be this granular when updating the addon for the samples.

The current build process run the `scons` build to generate the editor binaries, and run the `gradle` build to generate the AAR binaries used on device.
To that flow, I'd add an additional `gradle` (we don't need to modify the scons build) task to copy the full `demo/addons/godotopenxrvendors` directory to the sub directories under the `samples` folder.
To make that change we should only need to update the top `build.gradle` file.

devloglogan:
@m4gr3d That makes sense to me! I've swapped to your suggestion, adding a `copyBuildToSamples` task and reverting the other changes.

build.gradle
```diff
@@ -125,4 +125,15 @@ task buildSconsArtifacts {
 
     dependsOn 'buildGodotOpenXRVendorsDebugGDExtension'
     dependsOn 'buildGodotOpenXRVendorsReleaseGDExtension'
+}
+
+task copyBuildToSamples {
```
m4gr3d, line 130: Looks good! 

Let's have this task called by the `buildPlugin` task when it completes. This can be done by using the `finalizedBy 'copyBuildToSamples'` at the end of the `buildPlugin` task definition.

Let's also include a `cleanBuildFromSamples` task that can be called by the `clean` task; i.e: the `clean` task should include a `dependsOn 'cleanBuildFromSamples'` in its definition.

samples/fb-passthrough-sample/.gitignore
m4gr3d, line 1: Can we add a common `gitignore` and `gitattributes` file within the `sanples` directory since the content is likely to be the same for each added sample.

samples/fb-passthrough-sample/export_presets.cfg
```diff
@@ -0,0 +1,227 @@
...
+name="Quest"
+platform="Android"
+runnable=true
+advanced_options=true
+dedicated_server=false
+custom_features=""
+export_filter="all_resources"
+include_filter=""
+exclude_filter=""
+export_path="build/quest/XRPassthroughSample.apk"
```
m4gr3d, line 12: The `export_path` can be made relative to the `samples` directory so all samples output to the same location for easy access.

m4gr3d (changes requested):

devloglogan:
Added requested changes in latest push! Gradle confuses me a bit with the "configuration" vs "execution" phases but I was able to get the tasks working by adding `doLast` in both tasks, feel free to let me know if there's a better way to write that.

samples/fb-passthrough-sample/export_presets.cfg
```diff
@@ -0,0 +1,227 @@
...
+name="Quest"
+platform="Android"
+runnable=true
+advanced_options=true
+dedicated_server=false
+custom_features=""
+export_filter="all_resources"
+include_filter=""
+exclude_filter=""
+export_path="../builds/FbPassthroughSample.apk"
```
m4gr3d, line 12: We'll need to exclude this directory from the added copy and clean tasks.

samples/fb-passthrough-sample/main.gd
```diff
@@ -0,0 +1,210 @@
...
+@onready var interface_color_map_lut: Node3D = $Interfaces/InterfaceColorMapLut
+@onready var interface_color_map_interpolated_lut: Node3D = $Interfaces/InterfaceColorMapInterpolatedLut
+@onready var interface_disabled: Node3D = $Interfaces/InterfaceDisabled
+@onready var color_map_mesh: MeshInstance3D = $Interfaces/InterfaceColorMap/ColorMapMesh
+@onready var mono_map_mesh: MeshInstance3D = $Interfaces/InterfaceMonoMap/MonoMapMesh
+
+func _ready() -> void:
+	openxr_interface = XRServer.find_interface("OpenXR")
+	if openxr_interface and openxr_interface.initialize():
+		get_viewport().use_xr = true
```
m4gr3d, line 33: We should also include the logic to disable vsync.

m4gr3d (commented):
Looks good overall! Just some minor nits to address.

samples/fb-passthrough-sample/main.gd
devloglogan: Added!

m4gr3d (approved):

PR #165, opened 2024-06-17
Finish filling in all of the API docs

dsnopek:
This PR finishes filling in all the missing API documentation. This will appear both in the editor and in the online documentation.


dsnopek:
@devloglogan Can you review the docs I added for the extensions that you worked on?

Related to that:

- I unbound `OpenXRFbPassthroughExtensionWrapper::destroy_color_lut()` because I don't think developers  are meant to call that directly, but instead just free `OpenXRMetaPassthroughColorLut` objects.
- We have both `is_passthrough_supported()` and `has_passthrough_capability()` on `OpenXRFbPassthrough`. I'm not sure how to explain the difference.

Thanks!

devloglogan:
@dsnopek I agree with unbinding `OpenXRFbPassthroughExtensionWrapper::destroy_color_lut()`. As for the other point, I believe I meant to rename `is_passthrough_supported()` to `is_passthrough_started()`, but I never actually removed the first function. So those two functions just do the same thing. So, my vote is to remove the `is_passthrough_supported()` function. Or we could rename `has_passthrough_capability()` to `is_passthrough_supported()`. Whatever is preferable!

doc_classes/OpenXRFbBodyTrackingExtensionWrapper.xml
```diff
@@ -1,8 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="OpenXRFbBodyTrackingExtensionWrapper" inherits="OpenXRExtensionWrapperExtension" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
 	<brief_description>
+		Wraps the [code]XR_FB_body_tracking[/code] extension.
```
m4gr3d, line 4: Should we unregister the singleton for the extension wrapper classes that don't provide any methods?

dsnopek: The singleton isn't being registered, only the class, but that's enough for it to appear in the editor's help docs so we need to provide something here. And, unfortunately, we have to register the class, because unlike in Godot itself, GDExtension classes won't work without being registered.

doc_classes/OpenXRFbPassthroughGeometry.xml
```diff
@@ -1,15 +1,21 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <class name="OpenXRFbPassthroughGeometry" inherits="Node3D" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
 	<brief_description>
+		A mesh that passthrough imagery will be rendered upon (via the [code]XR_FB_passthrough[/code] extension).
 	</brief_description>
 	<description>
+		A mesh that passthrough imagery will be rendered upon (via the [code]XR_FB_passthrough[/code] extension).
+		Even though the node can be placed anywhere in 3D space, everything rendered by Godot is always shown on top of passthrough. If you want to give the impression that objects rendered by Godot can move behind the passthrough, then you must set [member enable_hole_punch] to [code]true[/code].
```
devloglogan, line 8: Hole punch is enabled by default for this node, since I think it's the more common use case. Personally, this line gives me the impression that it's not. Maybe the wording could be improved here, saying that it's enabled by default.

dsnopek:
> As for the other point, I believe I meant to rename `is_passthrough_supported()` to `is_passthrough_started()`, but I never actually removed the first function.

We already have an `is_passthrough_started()` also, but that's something else.

> So those two functions just do the same thing.

Well, looking at the code, they don't do exactly the same thing.

Here's what all the methods mentioned here appear to do:

- `is_passthrough_supported()`: checks if the `XR_FB_passthrough` module is enabled
- `has_passthrough_capability()`: checks if the `XR_PASSTHROUGH_CAPABILITY_BIT_FB` is set on `XrSystemPassthroughProperties2FB.capabilities`
- `is_passthrough_started()`: checks if we've been able to successfully execute `xrPassthroughStartFB()`

Looking at the spec, it seems like not all runtimes that support `XR_FB_passthrough` will support `XrSystemPassthroughProperties2FB`:

Selection_165

So, thinking about this a bit more, I think all the `has_passthrough_capability()` tells us, is if the developer can trust the result of `has_color_passthrough_capability()` or `has_layer_depth_passthrough_capability()`. Because, if the `XR_PASSTHROUGH_CAPABILITY_BIT_FB` wasn't set, then it probably means the runtime doesn't support `XrSystemPassthroughProperties2FB` because I think it's safe to assume that if the `XR_FB_passthrough` extension is enabled, that we have support for basic passthrough.

Does that sound right?

I'll update the docs to say that for now :-)

devloglogan (commented):
Just a couple small comments. Thanks for writing these out!

doc_classes/OpenXRFbPassthroughExtensionWrapper.xml
```diff
@@ -1,71 +1,87 @@
...
+		Wraps [code]XR_FB_passthrough[/code] and related extensions.
 	</brief_description>
 	<description>
+		Wraps the following OpenXR extensions:
+		- [code]XR_FB_passthrough[/code]
+		- [code]XR_FB_triangle_mesh[/code]
+		- [code]XR_META_passthrough_preferences[/code]
+		- [code]XR_META_passthrough_color_lut[/code]
+		Only the main [code]XR_FB_passthrough[/code] extension is required for basic passthrough functionality. The others allow for more complex features, such as filtering the passthrough output or using an [OpenXRFbPassthroughGeometry] node to control where passthrough is rendered. Settings for features unsupported by the current OpenXR runtime or headset will be ignored.
+		Passthrough will start automatically when [member XRInterface.environment_blend_mode] is set to [constant XRInterface.XR_ENV_BLEND_MODE_ALPHA_BLEND], and will automatically stop when set to any other value.
```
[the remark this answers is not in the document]
dsnopek: I'd personally save it for the tutorial. We'll be able to link to the tutorial in the section just below the description (`<tutorials></tutorials>`) anyway.

devloglogan:
@dsnopek Ah, I see, had a hasty misread when I looked at the functions in the header file. That makes sense to me. I'm still not a huge fan of the somewhat colliding function names, but maybe there isn't a perfect way around it in this case.

doc_classes/OpenXRFbPassthroughGeometry.xml
dsnopek: Ah, I forgot that it's enabled by default. I think I'll just remove this bit - there's another description of hole punching on the `enable_hole_punch` property, which is probably sufficient.

doc_classes/OpenXRFbPassthroughExtensionWrapper.xml
```diff
@@ -101,66 +129,94 @@
...
-			</description>
+				Sets the current passthrough filter to [constant PASSTHROUGH_FILTER_MONO_MAP], which will use the given curve to replace each input luminance value in the passthrough imagery with a grayscale color value.
+				[b]Note:[/b] Only one passthrough filter can be enabled at a time.
+				</description>
 		</method>
 		<method name="set_passthrough_filter">
 			<return type="void" />
 			<param index="0" name="filter" type="int" enum="OpenXRFbPassthroughExtensionWrapper.PassthroughFilter" />
 			<description>
+				Sets the current passthrough filter.
```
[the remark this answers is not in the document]
dsnopek: Great idea! Added in my latest push

doc_classes/OpenXRFbPassthroughExtensionWrapper.xml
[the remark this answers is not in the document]
devloglogan: Sounds good!

devloglogan (approved):

m4gr3d (approved):

Issue #163, opened 2024-06-10
Statically link libstdc++ on Linux for portability


goatchurchprime:
This is a portability issue identified first in the webrtc libraries (since fixed) https://github.com/godotengine/webrtc-native/issues/110

![image](https://github.com/GodotVR/godot_openxr_vendors/assets/677254/2533e71c-b91d-45bb-8288-39b61e51942c)

I think all it needs are some link flags in the SConstruct file
https://github.com/godotengine/webrtc-native/pull/109/commits/1de32aa9ea1f3f4e1fd3f5f9b283ac4dc2befae1

goatchurchprime:
Works now if you build it from source and push the godot-cpp module up by two weeks.

goatchurchprime:
This issue wasn't actually fixed.

![image](https://github.com/GodotVR/godot_openxr_vendors/assets/677254/584356ca-816b-45ba-8440-c4519c448fb8)

It just so happened that when I built from sources the shared object was linked to a library that did exist on my disk:

![image](https://github.com/GodotVR/godot_openxr_vendors/assets/677254/15e9c6fa-4d52-40f1-9def-0c5b3db3fc3f)

The correct result would have been for libstdc++ not to have been an external dependency to the library, as is the case with libwebrtc:

![image](https://github.com/GodotVR/godot_openxr_vendors/assets/677254/9dc66cbd-7da8-4c2a-a744-48e2adeb114f)

dsnopek:
Since this only affects desktop Linux (and not Android), I think it's probably fine to do despite the increased binary size. See PR https://github.com/GodotVR/godot_openxr_vendors/pull/166

PR #166, opened 2024-06-17
Statically link with libgcc and libstdc++ for more Linux compatibility.

dsnopek:
Fixes https://github.com/GodotVR/godot_openxr_vendors/issues/163

This will make the binary larger, but it'll allow the extension to be compatible with more Linux distros


dsnopek:
To add a little more info about the binary size increase when compiled on my system:

| Bulid | Before | After|
| --- | --- | --- |
| debug | 2.4M | 3.7M |
| release | 2.3M | 3.7M |

So, it's an increase of ~1.3M, which is pretty negligible on desktop.

m4gr3d (approved):

PR #168, opened 2024-06-18
Rename fb hand tracking aim gesture action names

devloglogan:
When working on the FB hand tracking sample, it came to light that the gesture actions could be named in a better way. Previously both the left/right hand gestures were called "System Gesture", but now they are called "Menu Gesture" and "System Gesture" per their different functions. What was previously called "Menu Gesture" is now called "Menu Pressed" indicating that the menu gesture has been pinched. "Menu Pressed" has no right hand / system gesture equivalent.


m4gr3d (approved):

dsnopek (approved):
Thanks! Looks good to me :-)

PR #170, opened 2024-06-19
Record XrSim tests for all the sample projects

dsnopek:
Now that we have tests via the Meta XR Simulator, this PR is adding tests for the two sample projects that are currently in the repo.

Some notes:

- Cylinder composition layers don't work quite right in the simulator - it's like the faces are wound in reverse, such that it gets back face culled when looking in the correct direction (but the layer is visible when looking in the opposite direction)
- Only basic passthrough is supported by the simulator - geometry passthrough and all the styles don't work
- ~~I had to downgrade to v64 of the simulator because passthrough wasn't working~~ **UPDATE:** These issues are fixed in v66, but it has it's own issues. I'm just going to use v65, even though passthrough isn't quite right - when there is a fully fixed version, we can re-record the VRS file.
- Since we have to update the simulator version in the CI manually, we'll be able to update the VRS files if/when the above bugs are fixed (and the current VRS files can help us notice when those bugs are fixed, since the tests will fail)
- **UPDATE:** I switched to using Git LFS to store the VRS files per the discussion below

This is a draft since I don't think the CI is going to work on the first try because I reorganized some things. :-) Once it's working, I'll take it out of draft. (Also, I'm temporarily basing this on PR https://github.com/GodotVR/godot_openxr_vendors/pull/149 to make iteration faster.)


dsnopek:
Question: Should we use Git LFS for the *.vrs files?

They don't get _that_ big, but if you make one that takes more than ~2 screenshots, they can get big-ish. The one for the passthrough sample on this PR is 57mb (I think I took 5 screenshots?). If we make a couple of revisions to that, it'll start to add up in the overall repo size when doing a full clone. The other *.vrs files are smaller: 5.9mb and 16mb (I only took 1 or 2 screenshots)

I'm personally leaning towards using Git LFS, but since we aren't using it at all in this repo yet, I wanted to run it by other folks first!

/cc @m4gr3d @devloglogan @BastiaanOlij

devloglogan:
Seeing that any .vrs files we add potentially need to be updated if the related sample/feature is changed, and that "more than ~2 screenshots" feels like an extremely low bar to hit for having a large file, I'd agree that git LFS seems like the right choice.

dsnopek:
Finally taking this one out of draft! I've got simple tests recorded for all the sample projects, except for the Scene API one: it works great when running it manually, but has some issue when running automatically - I suspect an XR Simulator bug, but I'm not really sure. In any case, it's something we can solve in a follow-up, rather than delaying on getting all the rest in.

This PR also includes some good refactoring and improvements of the CI for running these tests.

And I did end up moving the VRS files into Git LFS.

samples/meta-passthrough-sample/geometry.gd
```diff
@@ -2,4 +2,7 @@ extends OpenXRFbPassthroughGeometry
 
 
 func _process(delta: float) -> void:
-	position.x = 3 * sin(Time.get_ticks_msec() / 1000.0)
+	if not "--xrsim-automated-tests" in OS.get_cmdline_user_args():
+		# When we're running tests via the XR Simulator, we don't want the text
+		# to be animated, which can lead to differences in the screenshots.
```
devloglogan, line 7: Assuming this comment was missed when copied over from the composition layer sample. :)

dsnopek: Ah, thanks! Updated

devloglogan (approved):
LGTM!

m4gr3d (approved):

PR #172, opened 2024-06-21
Add manifest entries for PICO

BastiaanOlij:
Adds manifest items for Pico, these are all related to the eye tracker on the Pico 4 Pro so I can't fully test, but in theory these should work.

One thing worth checking is if the lipsync function works on the Pico 4. In theory it should do face tracking for the mouth based on audio. I do not (yet) know what face tracking API Pico is implementing, whether they are adding support for Metas one, or adding their own.

### DIFF
--- a/godotopenxrpico/src/main/java/org/godotengine/openxr/vendors/pico/GodotOpenXRPico.kt
+++ b/godotopenxrpico/src/main/java/org/godotengine/openxr/vendors/pico/GodotOpenXRPico.kt
@@ -29,9 +29,21 @@
 
 package org.godotengine.openxr.vendors.pico
 
+import android.Manifest
+import android.app.Activity
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.content.pm.PermissionInfo
+import android.net.Uri
+import android.os.Build
+import android.os.Environment
+import android.provider.Settings
 import android.util.Log
+import android.view.View
 import org.godotengine.godot.Godot
 import org.godotengine.godot.plugin.GodotPlugin
+import org.godotengine.godot.utils.PermissionsUtil
 
 /**
  * Godot plugin for the Pico OpenXR loader.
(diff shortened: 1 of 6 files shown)


godotopenxrpico/src/main/java/org/godotengine/openxr/vendors/pico/GodotOpenXRPico.kt
```diff
@@ -48,11 +63,116 @@ class GodotOpenXRPico(godot: Godot?) : GodotPlugin(godot) {
...
+                    Log.w(TAG, "Unable to identify permission $permission", e)
+                }
+            }
+            if (requestedPermissions.isEmpty()) {
+                // If list is empty, all of dangerous permissions were granted.
+                return true
+            }
+            activity.requestPermissions(requestedPermissions.toTypedArray<String>(), PermissionsUtil.REQUEST_ALL_PERMISSION_REQ_CODE)
+            return true
+        }
```
m4gr3d, line 136: Once you rebase on top of https://github.com/GodotVR/godot_openxr_vendors/pull/176, you can remove these methods; the updated version of Godot 4.3 we're targeting already includes them.

godotopenxrpico/src/main/java/org/godotengine/openxr/vendors/pico/GodotOpenXRPico.kt
```diff
@@ -48,11 +63,116 @@ class GodotOpenXRPico(godot: Godot?) : GodotPlugin(godot) {
...
+
+        /**
+         * Dispatch the necessary requests for all plugin's permissions in the app's manifest.
+         */
+        @JvmStatic
+        fun requestAllPluginPermissions(activity: Activity) {
```
m4gr3d, line 142: This method is not being called. Are you missing [this logic](https://github.com/GodotVR/godot_openxr_vendors/blob/10cf4012e202daa3f5734cd4c99b8acc9fcea38b/godotopenxrmeta/src/main/java/org/godotengine/openxr/vendors/meta/GodotOpenXRMeta.kt#L175-L178)?

m4gr3d (changes requested):

m4gr3d (approved):

PR #174, opened 2024-06-21
Add fb-body-tracking-sample project

devloglogan:
Moves the body tracking examples of the demo project into its own sample project.

This branch is based off of https://github.com/GodotVR/godot_openxr_vendors/pull/173 to simplify the stripping-down of the demo project. This PR will be marked as a draft until the predecessor is merged.


samples/fb-body-tracking-sample/main.gd
```diff
@@ -0,0 +1,16 @@
+extends Node3D
+
+# Called when the node enters the scene tree for the first time.
+func _ready():
+	var xr_interface = XRServer.find_interface("OpenXR")
+	if xr_interface and xr_interface.initialize():
+		xr_interface.session_stopping.connect(self._on_session_stopping)
+		var vp: Viewport = get_viewport()
+		vp.use_xr = true
```
m4gr3d, line 9: We should align with the best practices when initializing XR; this provides additional guidance for users referencing those sample projects.

See https://docs.godotengine.org/en/latest/tutorials/xr/a_better_xr_start_script.html for reference.

devloglogan: @m4gr3d after a talking about some of the startup script stuff with @dsnopek I'm opting for a slightly different approach to this than before. All of the samples now have the same `start_xr.gd` script that draw from the "Better XR Start Script" doc, and a sample's `main.gd` extends from that script (excluding this body tracking sample, which only uses `start_xr.gd` since no other code is needed). Now the sample-specific gdscript lives in a distinct place from the xr startup gdscript, which will make it easier to update these samples if that doc is changed in the future.

Let me know if you think this is still good!

devloglogan:
Opening this PR!

m4gr3d (approved):
Looks great!

PR #177, opened 2024-06-26
Create Meta Passthrough tutorial doc

devloglogan:
Adds a tutorial document to introduce users to the Meta passthrough functionality. 

I also restructured image storage a tiny bit concerning the scene manager tutorial, since dropping all images in one directory might get a little messy.


docs/manual/meta/passthrough.rst
```diff
@@ -0,0 +1,243 @@
...
+The ``OpenXRInterface`` parent class `XRInterface <https://docs.godotengine.org/en/stable/classes/class_xrinterface.html>`_ has a method ``get_supported_environment_blend_modes()``
+that will list ``XRInterface.XR_ENV_BLEND_MODE_ALPHA_BLEND`` if it is supported. To enable reconstruction passthrough, you might add the following to the script attached to the root node of your main scene:
+
+.. code-block:: gdscript
+
+    extends Node3D
+
+    @onready var world_environment: WorldEnvironment = $WorldEnvironment
+
+    func _ready():
```
m4gr3d, line 41: Let's use another method instead of `_ready`, since the user may want to enable passthrough at other times besides initialization.
Something like `_enable_passthrough(enable: bool) -> bool`

m4gr3d (approved):

m4gr3d:
@devloglogan Can you update the changelog in`CHANGES.md` as well.

docs/manual/meta/passthrough.rst
```diff
@@ -0,0 +1,248 @@
...
+The ``OpenXRInterface`` parent class `XRInterface <https://docs.godotengine.org/en/stable/classes/class_xrinterface.html>`_ has a method ``get_supported_environment_blend_modes()``
+that will list ``XRInterface.XR_ENV_BLEND_MODE_ALPHA_BLEND`` if it is supported. To enable reconstruction passthrough, you might add the following to the script attached to the root node of your main scene:
+
+.. code-block:: gdscript
+
+    extends Node3D
+
+    @onready var world_environment: WorldEnvironment = $WorldEnvironment
+
+    func _enable_passthrough(enable: bool) -> bool():
```
m4gr3d, line 41: The return type is `bool` but no return value is provided. Let's just change it to `void` instead.
Also, I don't think there's a parenthese `()` at the end of the return type.

devloglogan: Whoops! Should have caught that, fixed in latest push. :)

PR #180, opened 2024-06-29
Use Khronos loader for Pico and add store required manifest entries

BastiaanOlij:
As of PICO OS 5.9.0, PICO now uses the Khronos loader and has deprecated their own SDK.

After investigating a support query through [our forums for deploying a Godot app to the Pico store](https://forum.godotengine.org/t/openxr-apk-is-missing-version-code-fields-pico-store-submission/68865/3), we were pointed to the [Pico Quickstart page](https://developer.picoxr.com/document/native/openxr-sdk-quickstart/).

The key missing ingredient is likely the `pvr.sdk.version` manifest entry that now simply states "OpenXR" to indicate the Khronos loader is used.

This PR alters the PICO entry to use the Khronos loader and adds this new manifest entry. I have not merged the PICO settings into the Khronos loader section as the PICO permission requirements clash with HTCs permission requirements. Our only solution at the moment is to keep these as separate entries.

But everything seems to work otherwise.


BastiaanOlij:
Changed this the same way as #182 so we no longer need to keep the khronos binaries in our repo but they are pulled in as dependencies by gradle

m4gr3d (approved):

PR #189, opened 2024-07-11
Use SCons and Gradle cache on CI to improve build times

dsnopek:
This should improve build times on all PRs, but it will have the biggest impact on PRs that only change docs or CI configuration or anything not directly code related, because it'll re-use all the build artifacts from the cache.

Here's a build with an empty cache:

Selection_181

And here's one with no changes to the code with the cache fully populated:

Selection_182

I'm not entirely sure why the Android builds still take ~6 minutes, but it's a big improvement from ~15 minutes in any case! I think it may be because Gradle is running cmake and we aren't getting any of the cmake stuff cached?

**NOTE:** Marking this as a draft because it includes PR https://github.com/GodotVR/godot_openxr_vendors/pull/170, which is currently necessary for the CI to pass. I'll take out of draft and rebase once that one is merged.

### DIFF
--- a/.github/workflows/build-addon-on-push.yml
+++ b/.github/workflows/build-addon-on-push.yml
@@ -3,6 +3,11 @@ name: Build on push
 
 on: [push, pull_request]
 
+env:
+  # Only used for the cache key. Increment version to force clean build.
+  GODOT_BASE_BRANCH: master
+  SCONS_CACHE: ${{ github.workspace }}/.scons-cache/
+
 jobs:
   build:
     name: Building for ${{ matrix.name }} (${{ matrix.os }})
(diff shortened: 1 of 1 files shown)


devloglogan (approved):
Looks good at a glance!

dsnopek:
I've integrated a Gradle cache into this PR too, hoping that it would improve the Android build times, however, it only did so marginally:

Selection_186

So, only like ~30s faster. :-) I think we aren't getting the gains we should because Gradle is running cmake, and we aren't getting any of the cmake stuff cached? That's something that can probably be improved later, either by adding some cmake caching, or maybe telling Gradle about its intermediate artifacts so it can do the caching? Anyway, I think that can be saved for later.

I also noticed that we have two identical Android builds, that only capture different sets of artifacts. So, I've reduced that to one build that captures both sets of artifacts. The `GodotOpenXRVendorsAddon.zip` that is created still has all the same files and is the same size.

.github/workflows/build-addon-on-push.yml
```diff
@@ -3,6 +3,11 @@ name: Build on push
 
 on: [push, pull_request]
 
+env:
+  # Only used for the cache key. Increment version to force clean build.
+  GODOT_BASE_BRANCH: master
```
m4gr3d, line 8: Which version is the comment referring to?

m4gr3d (approved):
Looks good!

.github/workflows/build-addon-on-push.yml
dsnopek: It's the version of the plugin really,  but since I'm reusing the action from godot-cpp (which copied it from Godot), it has that name.

Anyway, if we use this on another branch (other than `master`) we'll want to change that to the name of the branch

dsnopek:
Rebased after PR https://github.com/GodotVR/godot_openxr_vendors/pull/170 has been merged, and so taking out of draft.

PR #193, opened 2024-07-17
[CI] Update to actions/deploy-pages@v4

dsnopek:
I don't know if this is related to the recent documentation deployment failures, but I just noticed we were using `v1` of the `actions/deploy-pages` action, when the latest is `v4`.

Unfortunately, this job will only run on the `master` branch, so the tests on this PR won't tell us if it works or not.


dsnopek:
Ah, it's definitely not related. I just re-ran the docs deployment job that failed and it succeeded. I'm guess it was a momentary hiccup from GitHub Pages.

We should still do this update, though, because v1 is deprecated, and will eventually no longer work.

m4gr3d (approved):

PR #195, opened 2024-07-20
Update scene_manager.rst

The-Cyber-Captain:
fixes typos referring to "open_xr_" - instead of "openxr_" signals in example snippet; now if no spacial env is found on the headset, one will be requested by this code.


dsnopek (approved):
Thanks! Looks good to me :-)

m4gr3d:
Thanks for the fix!

PR #196, opened 2024-07-20
Update scene_manager.rst

The-Cyber-Captain:
Assumes the intended / preferred logic of the scene scan example snippet - post capture - to be deletion of any existing scene_anchors and creation of new anchors in any case.

This way the anchors populate without user having to restart the app. (Tested good on Q2 v67 and Q3 v68)


dsnopek (approved):
Thanks! This looks good to me, and actually matches the code in the 'meta-scene-sample' project, which this may have been based on.

m4gr3d (approved):

PR #198, opened 2024-08-15
Update to Godot 4.3-stable

devloglogan:
Update `config.grade` and `godot-cpp` to Godot 4.3-stable!

### DIFF
--- a/.github/workflows/build-addon-on-push.yml
+++ b/.github/workflows/build-addon-on-push.yml
@@ -199,7 +199,7 @@ jobs:
    needs: build
 
    env:
-     GODOT_VERSION: "4.3-beta2"
+     GODOT_VERSION: "4.3-stable"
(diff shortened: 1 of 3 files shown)


dsnopek (approved):
Looks good to me!

The tests appear to be failed due to some issue with Meta's NPM server? That's where we download the XR Simulator from and it's just stalling forever, even when I try the link locally:

https://npm.developer.oculus.com/com.meta.xr.simulator/-/com.meta.xr.simulator-65.0.0.tgz

Hopefully they'll fix it soon, but if not, we may need to release without that bit of the tests passing.

devloglogan:
@m4gr3d @dsnopek Also just updated `GODOT_VERSION` in `build-addon-on-push.yml`!

BastiaanOlij:
I'm assuming we can ignore the XR simulator check for now?

PR #199, opened 2024-08-22
Fix issues when hand-tracking is disabled in project settings

dsnopek:
Currently, the hand-tracking and body-tracking samples are crashing with Godot v4.3-stable, because the hand-tracking project setting isn't enabled (it was enabled, but the default was changed just before Godot v4.3-stable, causing it to become disabled).

Godot PR https://github.com/godotengine/godot/pull/95959 will fix the crash too, because it's caused by us calling `OpenXRAPI::get_hand_tracker()`.

However, we really shouldn't even be trying to call it at all if the hand-tracking project setting is disabled! So, this PR checks the project setting (unfortunately, we can't check it early enough to prevent loading the extensions - this is due to a GDExtension limitation) and prevents the hand-tracking related extension wrappers from doing much if hand-tracking is disabled.

It also enables hand-tracking in the hand-tracking and body-tracking samples so that they actually work too :-)

### DIFF
--- a/common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
+++ b/common/src/main/cpp/extensions/openxr_fb_hand_tracking_aim_extension_wrapper.cpp
@@ -90,8 +90,14 @@ PackedStringArray OpenXRFbHandTrackingAimExtensionWrapper::_get_suggested_tracke
 }
 
 void OpenXRFbHandTrackingAimExtensionWrapper::_on_state_ready() {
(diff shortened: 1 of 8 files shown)


BastiaanOlij (approved):
LGTM

BastiaanOlij:
XR Meeting 23/8/24

Should be good to merge, this is something we need to keep in the back of our heads, never trust that things will be there because they can be disabled or fail to be created :)

devloglogan (approved):
LGTM, built and tested the relevant samples and everything looks good there.

PR #200, opened 2024-08-23
Remove unneeded Meta `fb_scene.h` header

dsnopek:
Seems to build fine for me locally!


m4gr3d (approved):

common/src/main/cpp/extensions/openxr_fb_scene_extension_wrapper.cpp
```diff
@@ -34,10 +34,6 @@
 #include <godot_cpp/templates/local_vector.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
-#ifdef META_VENDOR_ENABLED
-#include <openxr/fb_scene.h>
-#endif
```
m4gr3d, line 39: You can also remove the `#ifdef META_VENDOR_ENABLED` guard on line 110.

m4gr3d (changes requested):

common/src/main/cpp/extensions/openxr_fb_scene_extension_wrapper.cpp
dsnopek: Done!

m4gr3d (approved):

PR #207, opened 2024-09-19
Update switch statements to prevent build warnings

devloglogan:
These switch statements throw a lot of warnings every time I build the plugin because not every enum value is explicitly handled. Changing them to if statements fixes that.


m4gr3d:
I think the warnings are useful for that scenario; could we instead include all the enum, with a no-op comment for the ones where nothing needs to be done?
Or is the build complaining about `*_MAX` enum values?

devloglogan:
> Or is the build complaining about `*_MAX` enum values?

It does complain bout `*_MAX` enum values, but I see why keeping them switch statements could be valuable. No reason we can't just drop the `*_MAX` case in there as well I guess? I've updated the code to this.

m4gr3d (approved):
Looks good to me!

I'll wait until @BastiaanOlij and @dsnopek chime in as well before merging.

dsnopek (approved):
Looks good to me!

BastiaanOlij (approved):
lgtm

PR #208, opened 2024-09-20
Add support for `XR_FB_android_surface_swapchain_create`

dsnopek:
This adds support for `XR_FB_android_surface_swapchain_create` which gives two new flags that can be used to customize the creation of Android surface swapchains.

It depends on Godot PR https://github.com/godotengine/godot/pull/97252 (so setting as a draft for now)

That means that this will require Godot 4.4, which is why it is also updating the `extension_api.json` and godot-cpp. After this is merged, then the `master` branch will only work with Godot 4.4 (relating to our discussion from the last XR team meeting).


m4gr3d (approved):

dsnopek:
Taking this out of draft now that the Godot PR has been merged. However, this actually won't pass tests on the XR Simulator until after 4.4-dev3 is released and I change the CI to point to that, so I guess this will still need to wait a bit before we can merge it.

plugin/src/main/cpp/extensions/openxr_fb_android_surface_swapchain_create_extension_wrapper.cpp
```diff
@@ -0,0 +1,127 @@
...
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "extensions/openxr_fb_android_surface_swapchain_create_extension_wrapper.h"
+
+#include <godot_cpp/classes/object.hpp>
+#include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
```
devloglogan, line 34: I think this include is unused. Might as well remove it?

devloglogan (approved):
LGTM! Minus my one little nit.

plugin/src/main/cpp/extensions/openxr_fb_android_surface_swapchain_create_extension_wrapper.cpp
dsnopek: Ah, probably left over from some debugging :-)

I've removed it in my latest push!

dsnopek:
FYI, this one is failing testing because of a bug in Godot 4.4-dev3 - see https://github.com/godotengine/godot/issues/98102

It was fixed in PR https://github.com/godotengine/godot/pull/98187 which is already merged and should be included in Godot 4.4-dev4.

I'll update the CI to use 4.4-dev4 once it's released!

m4gr3d (approved):

dsnopek:
I've been trying to figure out why the XR Simulator tests on CI are crashing - here's some notes on it so far:

- It appears to only be the composition layer sample that is currently crashing (even without the changes from this PR - using the latest stable release from the asset library crashes)
- The latest Godot dev release (dev6) is crashing too
- The last working dev release was dev1
- Dev3 crashed due to a Vulkan issue that is fixed
- The crash in dev2 doesn't look _exactly_ like dev4 through dev6: it doesn't give a backtrace, the window just immediately closes. But the issue could potentially go back as far as dev2
- I tried to grab a recent build from GitHub Actions and it didn't crash, although, I'm not 100% confident it was from `master` and not a PR that hadn't been rebased or something like that. I need to try latest `master` for real, ideally, with a dev build so that I can get more information from the backtrace

Anyway, it doesn't seem like the CI issues really have anything to do with this PR, but do seem to represent an issue with upstream Godot and the XR Simulator

dsnopek:
Unfortunately, it looks like a custom compiled `master` and 4.4-dev6 work fine. It's only the official build of 4.4-dev6 that is crashing... :-(

dsnopek:
Once PR https://github.com/godotengine/godot/pull/100471 is merged, and there's a new 4.4-dev release, and the CI is updated to use it - then this should finally pass CI :-)

dsnopek:
This is FINALLY passing tests!

It already has approvals (based on the non-CI code) so I suppose I could merge it. :-) However, this will mean the `master` branch will become Godot 4.4+ (whereas right now it's 4.3+). Does anyone object to this happening now? (It would happen inevitably at some point)

PR #209, opened 2024-09-22
Reworked geometric algebra, plus circumvented Basis row:column glitch…

### LINKED ISSUE #205: XRBodyTracker::JOINT_<XXXX>_SHOULDER is broken ('Adjustment' mapping bugged)
**Tested versions**
plugin v3.0.1
Editor 4.4-dev1, 4.3-stable

**Issue description**

Was having some trouble getting good results on a humanoid rig, so stripped things back today and hacked together a scene which just spits out marker meshes for all the (75) joints found in XRServer.get_trackers(32).get("/user/body_tracker"). It showed JOINT_RIGHT_SHOULDER and JOINT_LEFT_SHOULDER behaving very weirdly... as if their order of rotation-translation was messed up or something?

Digging into the specs* and openxr_fb_body_tracking_extension_wrapper.cpp, I see we don't map the scapula - which makes sense enough - but then ALSO that there's an adjustment made to those very same two joints: JOINT_<XXXX>_SHOULDER? Suspicious. Then came across a magic number (-0.07) and couldn't get my head around how the adjustment is actually meant to work beyond that, I'm afraid :-/

I'll be first to admit... issue could easily be on Meta's side, my headset, even my big bushy beard getting in the way! Or just my imagination. However, don't seem to have any tracking issues of the ilk on any other platforms, and all signs point to this as the biggest anomaly on the rigging I was attempting.

Are we confident the 'adjustment' of shoulder_offset, along with the mapped starting quaternions, are working as intended?

*https://developers.meta.com/horizon/documentation/native/android/move-ref-body-joints
https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#_conventions_of_body_joints

The-Cyber-Captain:
…, to 'correct' root and shoulders trackers.

Root tracker does, indeed, now point 'forward'. IE +z aligns with hips & user's real-world 'forward' (but remaining along, and on, the XRorigin / Global XZ plane). Shoulder trackers also properly pulled back (the previously indicated 7cm) towards upper chest, more aligned with T-pose modelling.

Primarily fixes #205 , though leaves open discussion still re the approach of merely omitting / dropping XR_BODY_JOINT_<#####>_SCAPULA_FB data, then correction of XR_BODY_JOINT_<#####>_SHOULDER_FB, and techniques employed therein.

Appears also to have uncovered a row:column glitch when accessing Basis vectors.
"(we store transposed as rows for performance)"
"(remember that basis vectors are flipped!)" 
:-/

Can't seem to reproduce that in core engine - unsurprisingly. Likely something to do with either setting up this particular plugin, or GDExtension's parsing-binding of Basis. Yay! Oh joy. (Very much outside my wheelhouse here, but will see if I can learn enough about GDE to pinpoint, repro, and log an issue in the appropriate place.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ local.properties
 # Misc
 .DS_Store
 /plugin/src/gen/
+/zippedSamples
--- a/build.gradle
+++ b/build.gradle
@@ -30,6 +30,7 @@ task clean(type: Delete) {
 
     dependsOn 'cleanAddon'
     dependsOn 'cleanBuildFromSamples'
+    dependsOn 'cleanZippedSamples'
 
     dependsOn ':plugin:clean'
 }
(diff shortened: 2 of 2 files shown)


The-Cyber-Captain:
Tested as working here for 4.3-stable and 4.4-dev1 on both Quest2 and Quest3

PR #213, opened 2024-10-01
Update deprecated github actions.

Malcolmnixon:
This PR fixes some issues with the CI build pipelines:
1. Updates build actions to remove some of the deprecated warnings.
2. Fixes archiving the addon by including hidden files such as the .bin folders
3. Fixes godot build-cache restoring
4. Adds godot build-cache saving
5. Adds distributionSha256Sum to silence gradle warnings

The builds are much faster now and the build artifact is now packed and runs correctly.

The remaining deprecation issues are with the cache-script provided with godot-cpp so Godot will need to update it in the future.

### DIFF
--- a/.github/workflows/build-addon-on-push.yml
+++ b/.github/workflows/build-addon-on-push.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          - name: Linux (x86_64)
+          - name: 🐧 Linux (x86_64, GCC)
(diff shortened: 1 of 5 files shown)


BastiaanOlij (approved):
Looks good to me, all green flags on the CI so :)

dsnopek (approved):
Looks great to me!

Thanks for catching the issue with the renamed godot-cpp actions

Issue #212, opened 2024-10-01
Formatting checks report errors in CI


Malcolmnixon:
The Static Checks action for checking format is generating a stream of clang-format parse errors:
```
YAML:15:18: error: invalid boolean
AlignOperands:   DontAlign
                 ^~~~~~~~~
Error reading /home/runner/work/godot_openxr_vendors/godot_openxr_vendors/.clang-format: Invalid argument
YAML:1[5](https://github.com/Malcolmnixon/godot_openxr_vendors/actions/runs/11118011256/job/30890912770#step:5:6):18: error: invalid boolean
AlignOperands:   DontAlign
                 ^~~~~~~~~
Error reading /home/runner/work/godot_openxr_vendors/godot_openxr_vendors/.clang-format: Invalid argument
YAML:15:1[8](https://github.com/Malcolmnixon/godot_openxr_vendors/actions/runs/11118011256/job/30890912770#step:5:9): error: invalid boolean
AlignOperands:   DontAlign
                 ^~~~~~~~~
```

The reason appears to be the .clang-format `AlignOperands:   DontAlign` which I believe should be `AlignOperands:   false`

Malcolmnixon:
@devloglogan Could you look into this as I believe you checked in the initial clang-format file with this option.

Malcolmnixon:
Found it - the clang-format file is virtually identical to the godot main repo (so we can follow a similar coding convention) however the godot static-checks job runs on Ubuntu 22.04 and we're running on Ubuntu 20.04. These releases of Ubuntu have wildly different versions of clang-format and the configuration files are incompatible.

We need to update the static-checks job to 22.04 and then fix coding violations that have crept in.

devloglogan:
Fixed by https://github.com/GodotVR/godot_openxr_vendors/pull/214!

PR #214, opened 2024-10-02
Fix static-checks CI job

Malcolmnixon:
This PR fixes issue #212. Specifically it:
- Updates to ubuntu-22.04 (the same version as godotengine) to support the .clang-format settings
- Fixes the clang-format violations

Unfortunately while clang-format hasn't been correctly enforced we've had 21 source files committed with minor clang-format violations.

### DIFF
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 jobs:
   static-checks:
     name: Formatting (clang-format, file format)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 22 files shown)


BastiaanOlij:
I'm going to let others see if they are happy with this but I think this is good to merge. Big fan of clang format, but timing is always a pain for these types of changes.

dsnopek (approved):
Thanks!

This looks good to me. We don't have that many open PRs, so I think this is an OK time to merge this.

devloglogan (approved):
LGTM! Thanks for catching this. :)
Now seems as good a time as any to merge this.

PR #215, opened 2024-10-03
Add gradle tasks related to zipping sample projects

devloglogan:
@m4gr3d mentioned that having the vendor repo's sample projects available on the asset library would be super handy for users of the XR Editor. I think it makes most sense to host the zips of these in the vendor repo releases. This PR adds some gradle tasks to generate/clean those.

I'll add the zipped projects to the 3.0.1 release if we agree on this approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ local.properties
 # Misc
 .DS_Store
 /plugin/src/gen/
+/zippedSamples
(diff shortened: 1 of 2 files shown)


devloglogan:
Fixed the `zipSamples` task to also include the root directory in the zip, rather than just the contents. I've attached zips of the samples to the 3.0.1 release.

m4gr3d (approved):
Looks good!

dsnopek (approved):
Thanks, looks good!

Issue #162, opened 2024-06-09
Hand tracking on the Pico 4 seems to be broken


talpadk:
I believe that the hand tracking on pico 4 has a bug or two, or possibly I'm using OpenXR the wrong way
I'm using:
- Godot 4.3 beta 1
- Godot XR tools for Godot 4 (4.3.1) (for the hands and there skeletons)
- Godot OpenXR Vendors plugin v3.0.0 beta 1

I have tied both the depreciated OpenXRHand and XRNode3D + XRHandModifier3D

![20240609_18h51m17s_grim](https://github.com/GodotVR/godot_openxr_vendors/assets/3907657/c41e2bf5-d3f4-4af2-8388-8c5bb969d64c)

1. With both when I hold my hands horizontal the hands point up inside VR
2. Only the OpenXRHand modifies the bones, the XRHandModifier3D does nothing
3. The OpenXRHand with bone update = "rotation only" bends my fingers in the direction of my pinky finger
4. The Pico 4 insists that the program doesn't support running with hand tracking

In the export settings I have tired bot optional and required for hand tracking
![20240609_19h05m56s_grim](https://github.com/GodotVR/godot_openxr_vendors/assets/3907657/c5500d84-0f37-401e-bcc6-d22c010e23d2)

In the project settings i have enabled hand tracking and tried both with and without "Hand Interaction Profile" enabled 
![20240609_19h08m09s_grim](https://github.com/GodotVR/godot_openxr_vendors/assets/3907657/4dffb983-97f1-4172-b5f2-305e475c3829)

Some of the rotation and bending issues might be solvable by applying a rotation matrix to the data from the Pico 4 or perhaps even just swapping an axis or two.
It may of cause simply be a bug in the Pico firmware, but it would be nice if it could be worked around until they fix it (if ever).

As to why the XRHandModifier3D doesn't modify the bones I wouldn't rule out a misconfiguration on my part

Oh and last but not least: Thanks for the plugin quite amazing how relatively easy it is to code VR applications using it

talpadk:
If I apply the following Euler rotations to the Armature nodes:
Left hand: [0, 90, -90]
Right hand: [90,90,0]

The hands are oriented correctly inside VR.
(The fingers still bends the wrong way)

Which gives the rotation matrices
Left:
[  0.0000000,  0.0000000,  1.0000000;
  -1.0000000,  0.0000000,  0.0000000;
[...]

PR #216, opened 2024-10-08
Add two modes of pico hand tracking to selectable export features

ckschim:
This should resolve part of #162.

With this, hand tracking for PICO can be explicitly enabled using `<meta-data android:name="handtracking" android:value="1"/>`
I also added the high frequency mode with `<meta-data android:name="Hand_Tracking_HighFrequency" android:value="1"/>`

Both options I found in the official PICO SDK for Unity, oddly enough no other documentation anywhere: https://github.com/Pico-Developer/PICO-Unity-Integration-SDK/blob/f07f05f433a9671344c722a13ccc6f259e7e390d/Editor/PXR_BuildProcessor.cs#L229

image

### DIFF
--- a/plugin/src/main/cpp/export/pico_export_plugin.cpp
+++ b/plugin/src/main/cpp/export/pico_export_plugin.cpp
@@ -70,6 +70,16 @@ PicoEditorExportPlugin::PicoEditorExportPlugin() {
 			PROPERTY_USAGE_DEFAULT,
 			pico::FACE_TRACKING_NONE_VALUE,
 			false);
+
+	_hand_tracking_option = _generate_export_option(
+			"pico_xr_features/hand_tracking",
+			"",
+			Variant::Type::INT,
+			PROPERTY_HINT_ENUM,
+			"No,Low frequency,High frequency (60Hz)",
+			PROPERTY_USAGE_DEFAULT,
+			pico::HAND_TRACKING_NONE_VALUE,
+			false);
 }
 
 void PicoEditorExportPlugin::_bind_methods() {}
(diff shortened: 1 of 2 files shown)


plugin/src/main/cpp/export/pico_export_plugin.cpp
```diff
@@ -61,14 +71,14 @@ PicoEditorExportPlugin::PicoEditorExportPlugin() {
 			pico::MANIFEST_FALSE_VALUE,
 			false);
 
-	_face_tracking_option = _generate_export_option(
```
m4gr3d, line 64: Can you leave the `_face_tracking_option` here to keep the history clean and add the new hand tracking option below.

plugin/src/main/cpp/include/export/pico_export_plugin.h
```diff
@@ -65,9 +69,11 @@ class PicoEditorExportPlugin : public OpenXREditorExportPlugin {
 
 	Dictionary _eye_tracking_option;
 	Dictionary _face_tracking_option;
+	Dictionary _hand_tracking_option;
 
 private:
 	bool _is_eye_tracking_enabled() const;
+	bool _is_hand_tracking_enabled() const;
```
m4gr3d, line 76: I don't see an implementation for this method.

m4gr3d (changes requested):

plugin/src/main/cpp/include/export/pico_export_plugin.h
ckschim: Oops, I forgot about that, thank you! I removed it

m4gr3d (approved):
Looks good!

dsnopek (approved):
LGTM!

m4gr3d:
Thanks!

PR #219, opened 2024-10-12
Reworked geometric algebra. Accesses Basis via get_column.

### LINKED ISSUE #205: XRBodyTracker::JOINT_<XXXX>_SHOULDER is broken ('Adjustment' mapping bugged)

### LINKED ISSUE #218: Meta-body-tracking sample shoulders are broken.
**Tested versions**
plugin v3.0.1
Editor 4.4-dev1, 4.3-stable [Win].
Build target: Quest 2, Quest 3.

**Issue Description**
Shoulders are tracked - and therefore ultimately mapped by XRBodyModifier3D - incorrectly.

meta_sample_broken

Somewhat masked by the default camera location, and layout of the Test-Kun model.... this can, however, become nightmarishly apparent simply by switching bone_update mode to Full and / or repositioning of the view point.

The-Cyber-Captain:
… to 'correct' root and shoulder-tracking bugs respectively.

Root tracker does now point 'forward'. IE +z aligns with hips & user's real-world 'forward' (whilst remaining along, and on, the XRorigin / Global XZ plane). Shoulder trackers also properly pulled back (the previously indicated 7cm) towards upper chest, more aligned with T-pose modelling.

Having now delved a bit deeper into GDE wrappers, variants exposure, etc, I'm going to speculate that basis' unit vectors are, indeed, intended to be accessed via get_column out here - as we're firmly in C++ land still - and *that's* what initially tripped @m4gr3d up way back when? I've done a quick search of the files in GodotVR repo and couldn't find any others which access 'basis[]' beyond this file which would need addressing.

Primarily fixes #205, & fixes #218, though leaves open discussion still re:
* the approach of omitting / dropping XR_BODY_JOINT_<#####>SCAPULA_FB data, then correction of XR_BODY_JOINT<#####>_SHOULDER_FB, and techniques employed therein.
* modifying root at all!

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
@@ -314,17 +314,22 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
 	//   with how models are designed, rather than the Meta positions of
 	//   the tips of the clavicles.
 	if (locations.isActive) {
+		// Construct a root under the hips pointing forwards
+		// IE, +z aligns with hips & user's real-world forward.
+		// (Remaining, however, on the XRorigin / Global XZ plane; root's basis rotated around Y to fit)
+
 		// Get the hips transform
 		Transform3D hips = xr_body_tracker->get_joint_transform(XRBodyTracker::JOINT_HIPS);
-
-		// Construct the root under the hips pointing forwards
 		Vector3 root_y = Vector3(0.0, 1.0, 0.0);
-		Vector3 root_z = -hips.basis[Vector3::AXIS_X].cross(root_y);
-		Vector3 root_x = root_y.cross(root_z);
+		Vector3 hips_left = hips.basis.get_column(Vector3::AXIS_X);
+		// flatten hips' left (x basis) on to real-world / global ground plane. Normalize.
+		Vector3 root_x = (hips_left.slide(Vector3(0.0, 1.0, 0.0))).normalized();
+		Vector3 root_z = root_x.cross(root_y);
+		// flatten hips' origin on to ground plane.
 		Vector3 root_o = hips.origin.slide(Vector3(0.0, 1.0, 0.0));
+		// set the actual root transform
 		Transform3D root = Transform3D(root_x, root_y, root_z, root_o).orthonormalized();
 		xr_body_tracker->set_joint_transform(XRBodyTracker::JOINT_ROOT, root);
-		xr_body_tracker->set_pose("default", root, Vector3(), Vector3(), XRPose::XR_TRACKING_CONFIDENCE_HIGH);
 
 		// Distance in meters to push the shoulder joints back from the
 		// clavicle-position to be in-line with the upper arm joints as
@@ -333,7 +338,7 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
 
 		// Deduce the shoulder offset from the upper chest transform
 		Transform3D upper_chest = xr_body_tracker->get_joint_transform(XRBodyTracker::JOINT_UPPER_CHEST);
-		Vector3 shoulder_offset = upper_chest.basis[2] * shoulder_z_offset;
+		Vector3 shoulder_offset = upper_chest.basis.get_column(Vector3::AXIS_Z) * shoulder_z_offset;
 
 		// Correct the left shoulder
 		Transform3D left_shoulder = xr_body_tracker->get_joint_transform(XRBodyTracker::JOINT_LEFT_SHOULDER);
@@ -344,6 +349,10 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
 		Transform3D right_shoulder = xr_body_tracker->get_joint_transform(XRBodyTracker::JOINT_RIGHT_SHOULDER);
 		right_shoulder.origin += shoulder_offset;
 		xr_body_tracker->set_joint_transform(XRBodyTracker::JOINT_RIGHT_SHOULDER, right_shoulder);
+
+		// Set tracker pose, velocities, confidence.
+		// Good to go.
+		xr_body_tracker->set_pose("default", root, Vector3(), Vector3(), XRPose::XR_TRACKING_CONFIDENCE_HIGH);
 	}
 
 	// Register the XRBodyTracker if necessary
--- a/plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
@@ -307,29 +307,28 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
 	// If the location data is good then we need to apply some corrections
 	// before handing the data back to Godot. These include:
 	//
-	// - The Root joint carries no data, so instead place it on the ground
+	// - The Root joint carries no useful orientation data, so instead align it
 	//   under the hips pointing forwards.
 	//
 	// - Adjusting the left and right shoulder back so they are aligned
 	//   with how models are designed, rather than the Meta positions of
 	//   the tips of the clavicles.
 	if (locations.isActive) {
-		// Construct a root under the hips pointing forwards
-		// IE, +z aligns with hips & user's real-world forward.
-		// (Remaining, however, on the XRorigin / Global XZ plane; root's basis rotated around Y to fit)
+		// Align root under the hips pointing 'forwards'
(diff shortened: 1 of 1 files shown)


The-Cyber-Captain:
Tested as working here for 4.3-stable and 4.4-dev1 on both Quest2 and Quest3.

Neater code - from a slightly more knowledgeable position and tidied up dev env on my side - than the now discarded #209. Should pass clang* etc too this time around.

*Puts the magic into autoMAGIC that does, as far as my wee storyteller [Film / XR] brain is concerned.

plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -314,17 +314,22 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
 	//   with how models are designed, rather than the Meta positions of
 	//   the tips of the clavicles.
 	if (locations.isActive) {
+		// Construct a root under the hips pointing forwards
+		// IE, +z aligns with hips & user's real-world forward.
+		// (Remaining, however, on the XRorigin / Global XZ plane; root's basis rotated around Y to fit)
```
dsnopek, line 319: This comment is great! However, I think we could remove all the other comments added in this PR - they don't really add much that isn't already in this comment, or clear from the line of code they are commenting.

plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -344,6 +349,10 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
 		Transform3D right_shoulder = xr_body_tracker->get_joint_transform(XRBodyTracker::JOINT_RIGHT_SHOULDER);
 		right_shoulder.origin += shoulder_offset;
 		xr_body_tracker->set_joint_transform(XRBodyTracker::JOINT_RIGHT_SHOULDER, right_shoulder);
+
+		// Set tracker pose, velocities, confidence.
+		// Good to go.
+		xr_body_tracker->set_pose("default", root, Vector3(), Vector3(), XRPose::XR_TRACKING_CONFIDENCE_HIGH);
```
dsnopek, line 355: Personally, I don't like this line being moved to the end. I think it was better back up where the code calculates `root`. Also, like I said above, these comments don't really say anything that isn't clear from this line of code.

dsnopek (commented):
Thanks!

The math changes look correct to me. The `basis[i]` referring to rows in C++ (versus columns in GDScript) is something that always gets me too. :-) Otherwise, the only other change is constructing the Z-axis of the root basis from the X and Y, as opposed to constructing the X-axis from Y and Z -- this also makes sense to me.

I tested on the Quest 3, and it seemed to be working great!

I only have a couple of notes on the comments and the moving of one line explained below.

plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
The-Cyber-Captain: Copy that. I was kinda having to figure everything out from first principles - purely to rectify the broken SHOULDER joints I was experiencing - so things are / were a bit verbose, yes. Agreed.

plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
The-Cyber-Captain: Thanks @dsnopek . But would you believe it?... Was aiming to update the PR this evening (UK time) after an afternoon working on my project itself...  only to discover during that afternoon session that - as I see it - the initial assertion "The Root joint carries no data, so instead place it on the ground under the hips pointing forwards" is erroneous. Or, more accurately, zeroing out the position of root seems inconsistent / illogical when the reference play area is set to LOCAL (aka XR_PLAY_AREA_SITTING).

The new Root orientation (pointing real-world forward, and parallel to real-world ground) is good, but I *believe* the original position should actually be maintained. This way, when switched to LOCAL, Root doesn't magically teleport up to Head, but maintains its location in tracker-space: being found where the ground would be, between the two feet. [Exactly where all the good stuff, avatar control, and skeleton modifications happen to be in my project!] 

A quick-and-dirty hack locally to comment out the 'slide', then rebuild, seems to confirm this indeed works better for all play area modes. Is that also your understanding of how Root should behave, and therefore the functionality I should tidy up and include in my new commit for review?

TL;DR - New Root orientation good. But origin should be copied over from the original Tracker itself?

dsnopek: Perhaps. It'd be great if @Malcolmnixon could give his opinion - it's his original code that based the root position on the hips, rather than using the root tracker.

plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -307,23 +307,27 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
...
-		Vector3 root_z = -hips.basis[Vector3::AXIS_X].cross(root_y);
-		Vector3 root_x = root_y.cross(root_z);
-		Vector3 root_o = hips.origin.slide(Vector3(0.0, 1.0, 0.0));
+		Vector3 hips_left = hips.basis.get_column(Vector3::AXIS_X);
+		Vector3 root_x = (hips_left.slide(Vector3(0.0, 1.0, 0.0))).normalized();
+		Vector3 root_z = root_x.cross(root_y);
+		Vector3 root_o = xr_body_tracker->get_joint_transform(XRBodyTracker::JOINT_ROOT).origin;
 		Transform3D root = Transform3D(root_x, root_y, root_z, root_o).orthonormalized();
 		xr_body_tracker->set_joint_transform(XRBodyTracker::JOINT_ROOT, root);
+		// Set tracker pose, velocities, confidence.
```
dsnopek, line 330: I don't think we really need this comment either - it's not really saying anything that the line of code itself says just as well.

dsnopek (commented):
Thanks!

The  code clean up looks mostly good to me! (There's still one new comment - called out below - that I don't think adds anything and could be removed)

But I'm not sure about the change in the root origin calculation. I think it'd be best to hear what @Malcolmnixon thinks about that

m4gr3d:
@Malcolmnixon Can you take a look at this PR when you get a chance.

Malcolmnixon (approved):
This looks fine to me. I'm primarily using rotation-only avatar driving - as my physical skeleton rarely matches any avatars, so I wasn't noticing these issues.

The-Cyber-Captain:
> This looks fine to me. I'm primarily using rotation-only avatar driving - as my physical skeleton rarely matches any avatars, so I wasn't noticing these issues.

Yep it's a sneaky one; noticeable on rotation-only too, actually, when you get up, move about a bit, and - most particularly - find yourself rotated on the (wrt tracked space) primary Y axis / XZ plane. I'm driving rotation-only too in practise. But, indeed, by no means obvious in low-impact movement sort of VRChat use or testing.

Anyway, thanks for getting this merged in there... and for 3.1.0 milestone-ing it. I see you @m4gr3d 😊
Should leave me one less thing to version-manage here when that rolls out, bumping up both my personal productivity and - hopefully - also usefulness back to the community. 🫡

PR #221, opened 2024-10-23
Fix `OpenXRFbPassthroughExtensionWrapper` from wiping out the next pointer chain for system properties

dsnopek:
This code in `OpenXRFbPassthroughExtensionWrapper::_set_system_properties_and_get_next_pointer()` will wipe out the next pointer chain for system properties:

```c++
uint64_t OpenXRFbPassthroughExtensionWrapper::_set_system_properties_and_get_next_pointer(void *p_next_pointer) {
	system_passthrough_properties.next = &system_passthrough_color_lut_properties;
	return reinterpret_cast<uint64_t>(&system_passthrough_properties);
}
```

Notice how it doesn't do anything with `p_next_pointer`. If there was already anything pointed to by `p_next_pointer`, it would be lost.

This PR makes sure that `p_next_pointer` is included, and it also makes sure the required extensions are enabled before adding each system properties struct.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
@@ -133,8 +133,15 @@ void OpenXRFbPassthroughExtensionWrapper::cleanup() {
 }
 
 uint64_t OpenXRFbPassthroughExtensionWrapper::_set_system_properties_and_get_next_pointer(void *p_next_pointer) {
-	system_passthrough_properties.next = &system_passthrough_color_lut_properties;
-	return reinterpret_cast<uint64_t>(&system_passthrough_properties);
+	if (fb_passthrough_ext) {
+		system_passthrough_properties.next = p_next_pointer;
+		p_next_pointer = &system_passthrough_properties;
+	}
+	if (meta_passthrough_color_lut_ext) {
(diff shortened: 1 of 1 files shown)


dsnopek:
This could be a reason to do a `3.0.2` release? I don't recall if we merged anything that was Godot 4.4 specific yet; if so, we may need to try out the branching scheme we discussed at one of the recent XR team meetings

devloglogan (approved):
LGTM!

m4gr3d (approved):

m4gr3d:
> This could be a reason to do a `3.0.2` release? I don't recall if we merged anything that was Godot 4.4 specific, or even just a new feature yet; if so, we may need to try out the branching scheme we discussed at one of the recent XR team meetings

We have a few features we can merge and do a `3.1.0` release. What's the urgency for pushing this fix to production?

dsnopek:
> What's the urgency for pushing this fix to production?

It breaks other extension wrappers that use system properties structs (or at least breaks the parts that use the system properties structs). So, for example, the eye gaze and hand tracking extensions that are in Godot use system properties structs, and this bug will prevent them from being populated. In those particular cases, this just means they could report that eye gaze and hand tracking aren't supported, when they actually are.

> We have a few features we can merge and do a `3.1.0` release.

But, yeah, if we're going to do a 3.1.0 release soon anyway, then there really isn't a need for a 3.0.2.

PR #222, opened 2024-10-25
Implement XR_FB_space_warp extension

devloglogan:
Adds an extension wrapper implementation of [XR_FB_space_warp](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_FB_space_warp).

### Setup

**UPDATED (2024-12-12)**: Both mobile and compatibility renderers are now supported.

In order to test AppSW out (compatibility renderer only at the moment):
- Build Godot with one of the following PRs (depending on which renderer you'd like to test):
  - https://github.com/godotengine/godot/pull/97151
  - https://github.com/godotengine/godot/pull/100283
- Switch the `thirdparty/godot-cpp` submodule of this repo to `master` branch

Then you should be able to build the plugin like normal. Currently in this PR, AppSW is enabled by default, so no need to enable it.

### Debug View

**UPDATED (2024-11-5)**: I've added a new Meta Space Warp sample project to this PR, feel free to use that for testing.

~~I've (minimally) added multiple scenarios of moving elements in the demo project (particles/multimesh/skeleton) for testing purposes.~~ If you'd like a visual of the underlying motion vectors that are generated, you can use the following `adb` command.

```
adb shell setprop debug.oculus.spaceWarpDebug 1 && adb shell setprop debug.oculus.MVOverlay 4 && adb shell setprop debug.oculus.MVOverlay.Alpha 0.8 && adb shell input keyevent "KEYCODE_POWER" && adb shell input keyevent "KEYCODE_POWER"
```

Sometimes I find the project crashes when running the above command, but the debug view is visible upon restarting the app. To exit this view mode you can run:

```
adb shell setprop debug.oculus.spaceWarpDebug 0 && adb shell input keyevent "KEYCODE_POWER" && adb shell input keyevent "KEYCODE_POWER"
```

### To Do
**UPDATED (2024-11-5)**: These have all been completed.
- [x] Account for locomotion by properly setting `appSpaceDeltaPose`.
- [x] Enable the use of the `XR_COMPOSITION_LAYER_SPACE_WARP_INFO_FRAME_SKIP_BIT_FB` flag to skip frames of spacewarp.
- [x] Have the extension not enabled by default, and determine the best way for users to enable/disable it.


plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,202 @@
...
+		return;
+	}
+
+	Ref<OpenXRInterface> openxr_interface = XRServer::get_singleton()->find_interface("OpenXR");
+	int view_count = openxr_interface->get_view_count();
+	int width = system_space_warp_properties.recommendedMotionVectorImageRectWidth;
+	int height = system_space_warp_properties.recommendedMotionVectorImageRectHeight;
+
+	motion_vector_swapchain_info = get_openxr_api()->openxr_swap_chain_info_create(0, XR_SWAPCHAIN_USAGE_SAMPLED_BIT | XR_SWAPCHAIN_USAGE_COLOR_ATTACHMENT_BIT, 0x881A, width, height, 1, view_count);
+	motion_vector_depth_swapchain_info = get_openxr_api()->openxr_swap_chain_info_create(0, XR_SWAPCHAIN_USAGE_SAMPLED_BIT | XR_SWAPCHAIN_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT, 0x88F0, width, height, 1, view_count);
```
dsnopek, line 109:
```suggestion
	motion_vector_swapchain_info = get_openxr_api()->openxr_swap_chain_info_create(0, XR_SWAPCHAIN_USAGE_SAMPLED_BIT | XR_SWAPCHAIN_USAGE_COLOR_ATTACHMENT_BIT, 0x881A, width, height, 1, view_count);
	motion_vector_depth_swapchain_info = get_openxr_api()->openxr_swap_chain_info_create(0, XR_SWAPCHAIN_USAGE_SAMPLED_BIT | XR_SWAPCHAIN_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT, 0x88F0, width, height, 1, view_count);
```

We should use `#define`s for the OpenGL constants here (ie `0x881A` and `0x88F0`).

Also, related to [this comment](https://github.com/godotengine/godot/pull/97151#pullrequestreview-2395820410) on the Godot PR, we should be checking both that (1) OpenXR is using OpenGL and (2) that these swapchain formats are supported by the runtime. But that'll need the new APIs discussed there first.

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,202 @@
...
+}
+
+bool OpenXRFbSpaceWarpExtensionWrapper::is_available() {
+	return available;
+}
+
+void OpenXRFbSpaceWarpExtensionWrapper::enable_space_warp(bool p_enable) {
+	if (enabled == p_enable) {
+		return;
+	}
```
dsnopek, line 195: Given that we don't have any logic in this function, this `if` isn't really helpful here

plugin/src/main/cpp/include/extensions/openxr_fb_space_warp_extension_wrapper.h
```diff
@@ -0,0 +1,91 @@
+/**************************************************************************/
+/*  openxr_fb_render_model_extension_wrapper.h                            */
```
dsnopek, line 2: Copy-paste-o!

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,202 @@
+/**************************************************************************/
+/*  openxr_fb_render_model_extension_wrapper.h                            */
```
dsnopek, line 2: Here too

plugin/src/main/cpp/include/extensions/openxr_fb_space_warp_extension_wrapper.h
```diff
@@ -0,0 +1,91 @@
...
+	void _on_session_created(uint64_t p_instance) override;
+	void _on_session_destroyed() override;
+	void _on_state_ready() override;
+	void _on_main_swapchains_created() override;
+	void _on_pre_render() override;
+	void _on_post_draw_viewport(const RID &p_render_target) override;
+
+	bool is_available();
+
+	void enable_space_warp(bool p_enable);
```
dsnopek, line 63: I wonder if `set_space_warp_enabled()` or `set_enable_space_warp()` would be more Godot-ish? We do have a handful of `enable_*()` functions, but I think the other two variations are more common.

dsnopek (commented):
Thanks, this is looking great so far!

Some additional notes about the demo

- I think this should get it's own sample rather than modifying the main "demo" project
- There is no plain `MeshInstance3D` moving, only the more advanced cases. We should have the simplest case shown too
- Having a button to enable/disable AppSW would be nice
- I think we should add a plane with a very busy texture (plaid?) behind all the moving objects to make the artifacts more obvious

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
devloglogan: Makes sense, I've converted these to defines and will work on the other functionality 👍

plugin/src/main/cpp/include/extensions/openxr_fb_space_warp_extension_wrapper.h
devloglogan: I agree, I've updated it to be `set_space_warp_enabled`!

devloglogan:
@dsnopek I'd done the previous XR sample projects as their own PRs following the merging of the corresponding extensions. I had planned to just keep the simple test scene in the demo project while this PR remained as a draft. I see nothing wrong with expanding on it and just adding an XR sample project to this PR as well though! I will do that.

devloglogan:
I've added a fair amount of updates:

* Moved the test scene into its own "Meta Space Warp" sample project, included in this PR.
  * Added a normal, moving `MeshInstance3D` to the scene
  * Space warp can now be enabled/disabled by pressing the A button.
  * Added movement/turning with the analog sticks.
  * Added a CC0 grass texture behind moving objects so users can see space warp artifacts more easily.
* Added a project setting for enabling space warp.
* Space warp is now automatically disabled if a non-opengl renderer is being used.
* Added a `skip_space_warp_frame()` function to skip space warp for the following frame.
* I believe I'm now accounting for locomotion correctly in `appSpaceDeltaPose`.

I will open this PR once https://github.com/godotengine/godot/pull/97151 is merged!

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+void OpenXRFbSpaceWarpExtensionWrapper::_on_main_swapchains_created() {
+	if (!fb_space_warp_ext) {
+		return;
+	}
+
+	Ref<OpenXRInterface> openxr_interface = XRServer::get_singleton()->find_interface("OpenXR");
+	int view_count = openxr_interface->get_view_count();
+
+	space_warp_info = memnew_arr(XrCompositionLayerSpaceWarpInfoFB, 2);
```
dsnopek, line 140: I don't see where this array is `memdelete()`d?

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+
+void OpenXRFbSpaceWarpExtensionWrapper::_on_session_created(uint64_t p_instance) {
+	if (!fb_space_warp_ext) {
+		return;
+	}
+
+	String graphics_api_name = get_openxr_api()->get_graphics_api_name();
+	if (graphics_api_name != "OpenGL") {
+		UtilityFunctions::print("Disabling XR_FB_space_warp extension; this extension is only implmented for OpenGL");
```
dsnopek, line 93: This should only print this if we were going to enable spacewarp in the first place, like if the project setting is enabled. I don't think that'll be the case, because I think `_on_session_created()` runs before `_on_session_ready()`, which is where the project setting is checked.

It should also perhaps be `print_verbose()` too?

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,252 @@
...
+void OpenXRFbSpaceWarpExtensionWrapper::set_space_warp_enabled(bool p_enable) {
+	enabled = p_enable;
+}
+
+void OpenXRFbSpaceWarpExtensionWrapper::skip_space_warp_frame() {
+	if (!fb_space_warp_ext || !enabled) {
+		return;
+	}
+
+	render_state.skip_space_warp_frame = true;
```
dsnopek, line 230: This can't just set this value, because this is probably not called on the render thread.

You'll need to have an internal function that is deferred to the render thread, like:

```c++
void OpenXRFbSpaceWarpExtensionWrapper::_skip_space_warp_frame() {
	render_state.skip_space_warp_frame = true;
}

void OpenXRFbSpaceWarpExtensionWrapper::skip_space_warp_frame() {
	if (!fb_space_warp_ext || !enabled) {
		return;
	}
	RenderingServer::get_singleton()->call_on_render_thread(callable_mp(this, &OpenXRFbSpaceWarpExtensionWrapper::_skip_space_warp_frame));
}
```

samples/meta-space-warp-sample/main.tscn
```diff
@@ -0,0 +1,184 @@
...
+[node name="BoneCube" parent="." instance=ExtResource("4_1bvp3")]
+transform = Transform3D(-4.37114e-08, 0, 1, 0, 1, 0, -1, 0, -4.37114e-08, -3, 1, -1)
+
+[node name="Label3D" type="Label3D" parent="BoneCube"]
+transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, -1.52612, 0, 0)
+text = "Skeleton"
+
+[node name="MovingCube" type="Node3D" parent="."]
+transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, -3, 2.399, -1)
+script = ExtResource("5_lquwl")
```
dsnopek, line 159: This is `moving_cube.gd`. For some reason, it's attached to both the parent `MovingCube` (which is a `Node3D`) and the child `MovingCube` (which is a `MeshInstance3D`)?

This causes the sample not to run for me. But if I detach the script from the parent here, then it runs fine.

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,252 @@
...
+	get_openxr_api()->set_velocity_depth_texture(motion_vector_depth_swapchain_image);
+
+	int target_width = system_space_warp_properties.recommendedMotionVectorImageRectWidth;
+	int target_height = system_space_warp_properties.recommendedMotionVectorImageRectHeight;
+	Size2i render_target_size = { target_width, target_height };
+	get_openxr_api()->set_velocity_target_size(render_target_size);
+
+	Transform3D world_transform = XRServer::get_singleton()->get_world_origin();
+	Transform3D delta_transform = render_state.previous_transform.affine_inverse() * world_transform;
+	Quaternion delta_quat = delta_transform.basis.get_quaternion();
```
dsnopek, line 192: I hate smooth rotation in VR, but we should probably have it in the sample, because otherwise we're not testing that `delta_quat` is accurate

samples/meta-space-warp-sample/main.gd
```diff
@@ -0,0 +1,70 @@
...
+	if not turn_timer.is_stopped():
+		return
+
+	if name == "primary":
+		if value.x > TURN_THRESHOLD:
+			rotate_player(TURN_ANGLE)
+			turn_timer.start()
+		if value.x < -TURN_THRESHOLD:
+			rotate_player(-TURN_ANGLE)
+			turn_timer.start()
```
dsnopek, line 51: We should be calling `skip_space_warp_frame()` when we do snap turns.

I'm not actually able to see any artifacts from not doing it, but theoretically, because we're including the rotation since the last frame in the `appSpaceDeltaPose`, we should see it generating an in-between "smear" frame (which is not what a developer would want on snap turn, and why we'd skip space warp).

Because I'm not seeing the expected artifact, I'm worried we might not be doing the rotation in `appSpaceDeltaPose` correctly, which is another reason why I think we should have smooth turn in the sample to confirm.

dsnopek (commented):
Thanks!

This is starting to look really good! I especially like all the changes to the sample project - it'll make it very easy for developers to see the sort of artifacts that AppSW can generate, and seeing them also helps us confirm that things are doing what we would expect :-)

samples/meta-space-warp-sample/main.gd
dsnopek: After discussing with Logan, and doing some experimentation, I've realized that I had incorrect ideas of what artifact we should (or should not) see here.

I think the current calculation of `appSpaceDeltaPose` is correct. However, I think it's still worth skipping AppSW on snap turns, and adding smooth turn to demonstrate the correctness of the calculation (which I was already seeing with snap turns not generating an artifact, but we won't any more if we skip AppSW on snap turns).

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
devloglogan: Whoops, added the deletion to `_on_session_destroyed()`!

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
devloglogan: Yea after further thought I don't see why the project setting check shouldn't be in `_on_session_created()`, so I moved it there to prevent this from happening. Changed it to verbose as well!

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
devloglogan: Ah my mistake, I've added this suggestion and it seems to work!

samples/meta-space-warp-sample/main.tscn
devloglogan: Huh I wonder why this causes it to not run for you, it runs just fine with both attached. But I've fixed this!

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
devloglogan: Smooth rotation mode added!

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+void OpenXRFbSpaceWarpExtensionWrapper::_on_main_swapchains_created() {
+	if (!fb_space_warp_ext) {
+		return;
+	}
+
+	Ref<OpenXRInterface> openxr_interface = XRServer::get_singleton()->find_interface("OpenXR");
+	int view_count = openxr_interface->get_view_count();
+
+	space_warp_info = memnew_arr(XrCompositionLayerSpaceWarpInfoFB, 2);
```
m4gr3d, line 151: `2` should probably be replaced by `view_count`.

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,272 @@
...
+	get_openxr_api()->openxr_swapchain_release(motion_vector_depth_swapchain_info);
+}
+
+bool OpenXRFbSpaceWarpExtensionWrapper::is_enabled() {
+	return fb_space_warp_ext && enabled;
+}
+
+void OpenXRFbSpaceWarpExtensionWrapper::set_space_warp_enabled(bool p_enable) {
+	enabled = p_enable;
+}
```
m4gr3d, line 234: Do we have a use-case for this setter, given space warp can also be enabled via the project settings?

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,289 @@
...
+	if (!fb_space_warp_ext || !enabled) {
+		return;
+	}
+
+	Ref<OpenXRInterface> openxr_interface = XRServer::get_singleton()->find_interface("OpenXR");
+	int view_count = openxr_interface->get_view_count();
+	int width = system_space_warp_properties.recommendedMotionVectorImageRectWidth;
+	int height = system_space_warp_properties.recommendedMotionVectorImageRectHeight;
+
+	String rendering_method = RenderingServer::get_singleton()->get_current_rendering_method();
```
dsnopek, line 139: Checking rendering method seems wrong here - what we care about is the graphics API used by OpenXR. The "mobile" renderer could be using Vulkan, DirectX or Metal, and in each of those cases, we'd need to use different ids for the texture formats.

Didn't we add a method to get the name of the graphics API used by OpenXR specifically for this purpose?

devloglogan: I've fixed this in my latest push by checking the render driver name here instead. I did end up removing the method returning the graphics API used by OpenXR since I deemed it not necessary, but I can always add it back in at some point if need be.

dsnopek:
@devloglogan Can you rebase this on `master`? You'll need to remove your changes to godot-cpp and `extension_api.json`

I think we should consider merging this, even though the changes to Godot's renderer haven't been merged yet. All the API changes were merged in PR https://github.com/godotengine/godot/pull/100282, so this should build and be compatible with Godot 4.4. And having AppSW in an official release of godot_openxr_vendors will make it easier to distribute Godot builds that have the renderer changes, because then we won't need an alternate build of godot_openxr_vendors too.

I'm going to discussion of this to the XR team meeting agenda

dsnopek:
@devloglogan The tests are failing because the images aren't matching what was captured in the `test1.vrs`. I'm guessing the sample project has changed since you created the `test1.vrs`? You'll probably need to recreate it, or, perhaps, just not have a VRS test file, since I don't know that the XR simulator can even do AppSW

devloglogan:
Some updates:
* I've added some more vertical movement to the sample project (was pretty much horizontal only until now), added a readme with a screenshot (which will link to the docs when they're written).
* Updated the VRS file, seems to be working now :)
* Git will now ignore `.import` files in the sample project screenshot directories. I also removed some unneeded git files that were recently merged in the hybrid app sample.

I'm going to open this PR since I also think it makes sense to consider merging it at this point!

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,289 @@
...
+		uint64_t value = reinterpret_cast<uint64_t>(ext.second);
+		result[key] = (godot::Variant)value;
+	}
+	return result;
+}
+
+uint64_t OpenXRFbSpaceWarpExtensionWrapper::_set_system_properties_and_get_next_pointer(void *p_next_pointer) {
+	system_space_warp_properties.next = p_next_pointer;
+	return reinterpret_cast<uint64_t>(&system_space_warp_properties);
+}
```
m4gr3d, line 80: We should check if space warp is supported and enabled before chaining.

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,289 @@
...
+		return;
+	} else {
+		enabled = true;
+	}
+
+	String rendering_method = (String)project_settings->get_setting_with_override("rendering/renderer/rendering_method");
+	if (rendering_method == "forward_plus") {
+		UtilityFunctions::print_verbose("Disabling XR_FB_space_warp extension; this extension is not supported in the forward plus renderer");
+		fb_space_warp_ext = false;
+		enabled = false;
```
m4gr3d, line 113: Can we just call `cleanup()`.

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,289 @@
...
+		swapchain_format = GL_RGBA16F;
+		depth_swapchain_format = GL_DEPTH24_STENCIL8;
+	} else if (rendering_driver_name == "vulkan") {
+		swapchain_format = VK_FORMAT_R16G16B16A16_SFLOAT;
+		depth_swapchain_format = VK_FORMAT_D24_UNORM_S8_UINT;
+	} else {
+		UtilityFunctions::print_verbose("Disabling XR_FB_space_warp extension; rendering driver is not supported: ", rendering_driver_name);
+		fb_space_warp_ext = false;
+		enabled = false;
+		get_openxr_api()->unregister_projection_views_extension(this);
```
m4gr3d, line 152: Can we just call `cleanup()`, and include `get_openxr_api()->unregister_projection_views_extension(this);` within the `cleanup` logic?

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,289 @@
...
+		space_warp_info[i].farZ = get_openxr_api()->get_render_state_z_near();
+		space_warp_info[i].nearZ = get_openxr_api()->get_render_state_z_far();
+	}
+
+	render_state.skip_space_warp_frame = false;
+	render_state.previous_transform = world_transform;
+}
+
+void OpenXRFbSpaceWarpExtensionWrapper::_on_post_draw_viewport(const RID &p_render_target) {
+	if (!fb_space_warp_ext || !enabled) {
```
m4gr3d, line 237: Here and elsewhere, it looks like we can replace this logic with `!is_enabled()`

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+void OpenXRFbSpaceWarpExtensionWrapper::add_project_setting() {
+	String p_name = "xr/openxr/extensions/application_space_warp";
+	if (!ProjectSettings::get_singleton()->has_setting(p_name)) {
+		ProjectSettings::get_singleton()->set_setting(p_name, false);
+	}
+
+	ProjectSettings::get_singleton()->set_initial_value(p_name, false);
+	ProjectSettings::get_singleton()->set_as_basic(p_name, true);
```
m4gr3d, line 283: Wondering if space warp should be considered an advanced setting?

m4gr3d (changes requested):

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
devloglogan: Hm... good question! The recommended layer resolution extension that was merged recently is also setup as a basic project setting. What should qualify something as being advanced?

devloglogan:
@m4gr3d thanks for the review, I've applied some of your suggestions. :)

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
m4gr3d: IMO, anything that should be enabled by default for new projects should be considered `basic`, otherwise it would fall under the advanced category. 
I don't expect space warp to be enabled by default for new projects hence why I think it should be advanced. 
Recommended layer resolution on the other hand, as its name indicates should be enabled by default.

m4gr3d (approved):

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
devloglogan: I think this makes sense, I've removed setting the project setting to basic. :)

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
devloglogan: I'm using the project setting to determine the initial "enabled" status and it looks like the timing of this function happens prior to `_on_session_created()`, when I read the project setting. Currently, if space warp is supported, this will always be set. Is this problematic?

dsnopek: I think the project setting stuff will be resolved by either https://github.com/GodotVR/godot_openxr_vendors/pull/272 or https://github.com/GodotVR/godot_openxr_vendors/pull/234.

Both aim to not enable OpenXR extensions at all if the project setting isn't set, which means you won't have to deal with the project setting in your code at all - just check if the OpenXR extension was enabled or not

docs/manual/meta/application_space_warp.rst
```diff
@@ -0,0 +1,20 @@
+Meta Application Space Warp
+===========================
+
+.. note::
+
+    Check out the `Meta Space Warp Sample Project <https://github.com/GodotVR/godot_openxr_vendors/tree/master/samples/meta-space-warp-sample>`_
+    for a working demo of Application Space Warp.
+
+Application Space Warp is an extension for Meta XR devices that allows applications to render at half FPS while the runtime
+generates the in-between frames, perserving a smooth experience for users while lowering the workload of the device's CPU/GPU.
```
dsnopek, line 10: Maybe extend this to say something about using the extra GPU power to generate better visuals?

doc_classes/OpenXRFbSpaceWarpExtensionWrapper.xml
```diff
@@ -0,0 +1,34 @@
...
+			<return type="bool" />
+			<description>
+				Checks if the extension is enabled or not.
+			</description>
+		</method>
+		<method name="set_space_warp_enabled">
+			<return type="void" />
+			<param index="0" name="enable" type="bool" />
+			<description>
+				Sets the extension as enabled or disabled.
```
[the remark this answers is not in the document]
devloglogan: Added!

doc_classes/OpenXRFbSpaceWarpExtensionWrapper.xml
```diff
@@ -0,0 +1,34 @@
...
+				Note: The Application Space Warp project setting must be enabled in order to enable Application Space Warp.
+			</description>
+		</method>
+		<method name="skip_space_warp_frame">
+			<return type="void" />
+			<description>
+				Skips space warp frame extrapolation for the next rendered frame.
```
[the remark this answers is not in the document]
devloglogan: Added!

dsnopek:
I just re-tested this with Meta Quest 3, and experienced a bug where disabling space warp in the demo (by pressing the A button on the right controller) wasn't actually disabling space warp for me.

Making these changes got it working in my testing, although, I didn't spend too much time on it, so this might not be the right way to go:

```
diff --git a/plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp b/plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
index d9af4eb..36655cc 100644
--- a/plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
@@ -84,7 +84,7 @@ uint64_t OpenXRFbSpaceWarpExtensionWrapper::_set_system_properties_and_get_next_
 }
 
 uint64_t OpenXRFbSpaceWarpExtensionWrapper::_set_projection_views_and_get_next_pointer(int p_view_index, void *p_next_pointer) {
-       if (fb_space_warp_ext) {
+       if (is_enabled()) {
                space_warp_info[p_view_index].next = p_next_pointer;
                return reinterpret_cast<uint64_t>(&space_warp_info[p_view_index]);
        } else {
@@ -200,6 +200,8 @@ void OpenXRFbSpaceWarpExtensionWrapper::_on_main_swapchains_created() {
 
 void OpenXRFbSpaceWarpExtensionWrapper::_on_pre_render() {
        if (!is_enabled()) {
+               get_openxr_api()->set_velocity_texture(RID());
+               get_openxr_api()->set_velocity_depth_texture(RID());
                return;
        }
```

docs/manual/meta/application_space_warp.rst
```diff
@@ -0,0 +1,20 @@
+Meta Application Space Warp
+===========================
+
+.. note::
+
+    Check out the `Meta Space Warp Sample Project <https://github.com/GodotVR/godot_openxr_vendors/tree/master/samples/meta-space-warp-sample>`_
+    for a working demo of Application Space Warp.
+
+Application Space Warp is an extension for Meta XR devices that allows applications to render at half FPS while the runtime
```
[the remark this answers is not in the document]
dsnopek: So, I just tested on Pico 4 Ultra, and Pico does support the `XR_FB_space_warp` extension, and I can artifacts that match expectations when I toggle it on and off in the sample.

However, it doesn't seem like it's working as well as on Meta. The artifacts are way worse and the motion isn't very smooth, like, it kind of feels like we got the artifacts, but not the actual frame synthesis, which is weird.

I'm also getting a mess of errors at startup:

Selection_297

So, I suspect Pico's implementation is doing something a little differently that we may need to account for. I don't have time to dig into it more now, but I'll try to take a look at it later.

devloglogan:
@dsnopek thanks for testing! The disabling/enabling must have broken in some of my changes a few weeks ago. I think your patch is all that is needed and it works well from my testing.

plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension_wrapper.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+void OpenXRFbSpaceWarpExtensionWrapper::_on_main_swapchains_created() {
+	if (!fb_space_warp_ext) {
+		return;
+	}
+
+	Ref<OpenXRInterface> openxr_interface = XRServer::get_singleton()->find_interface("OpenXR");
+	int view_count = openxr_interface->get_view_count();
+
+	space_warp_info = memnew_arr(XrCompositionLayerSpaceWarpInfoFB, view_count);
```
dsnopek: This works, but we've been moving towards using `LocalVector` recently, rather than manually managing the memory

samples/meta-space-warp-sample/project.godot
```diff
@@ -0,0 +1,40 @@
...
+[rendering]
+
+renderer/rendering_method="mobile"
+textures/vram_compression/import_etc2_astc=true
+
+[xr]
+
+openxr/enabled=true
+openxr/reference_space=2
+openxr/extensions/eye_gaze_interaction=true
```
dsnopek: We can probably disable `eye_gaze_interaction`

samples/meta-space-warp-sample/project.godot
```diff
@@ -0,0 +1,40 @@
...
+
+[debug]
+
+settings/stdout/verbose_stdout=true
+
+[layer_names]
+
+3d_physics/layer_1="Virtual Environment"
+3d_physics/layer_2="Scene Understanding"
+3d_physics/layer_3="Spatial Anchors"
```
dsnopek: Left over stuff from starting this project as a copy of another?

dsnopek (approved):
I just retested this with latest `master` and it seems to be working well!

I also skimmed through the code again, and only found a couple of little nit-picks

BastiaanOlij:
Thanks!!!

PR #226, opened 2024-11-18
Implement XR_META_recommended_layer_resolution extension

devloglogan:
Adds an extension wrapper implementation of [`XR_META_recommended_layer_resolution`](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_META_recommended_layer_resolution).

The extension can be enabled via project settings. When enabled, the extension will dynamically update the rendered image resolutions at a resolution of less or equal size to the initial swapchain size. I've primarily tested this just by spawning enough cube mushes in the demo project until FPS starts taking a hit, at which point the runtime will typically begin recommending lower resolutions as more cubes are added.

~~This PR relies on https://github.com/godotengine/godot/pull/99407, so I'll keep it as a draft until those changes are merged.~~

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_meta_recommended_layer_resolution_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_meta_recommended_layer_resolution_extension_wrapper.cpp
@@ -1,0 +1,147 @@
+/**************************************************************************/
+/*  openxr_meta_recommended_layer_resolution_extension_wrapper.cpp        */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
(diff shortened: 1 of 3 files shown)


dsnopek:
I did some testing with this on Quest 3, and saw really great performance gains!

In my rendering benchmarking project, it gathers some metrics at 3 different "markers" in the scene, one of which usually drops down to 55fps (from the target 72fps which all the other markers reach). However, when running it with this extension enabled, it keeps a steady 72fps (at the cost of visibly losing resolution, but it's not terrible). I added some print statements and I can see it scaling the resolution down at that marker, and then scaling it back up at the next.

This is really cool! :-)

The code in the PR looks good to me as well.

plugin/src/main/cpp/extensions/openxr_meta_recommended_layer_resolution_extension_wrapper.cpp
```diff
@@ -0,0 +1,147 @@
...
+		result[key] = (godot::Variant)value;
+	}
+	return result;
+}
+
+void OpenXRMetaRecommendedLayerResolutionExtensionWrapper::_on_instance_created(uint64_t p_instance) {
+	if (meta_recommended_layer_resolution_ext) {
+		bool result = initialize_meta_recommended_layer_resolution_extension((XrInstance)p_instance);
+		if (!result) {
+			UtilityFunctions::print("Failed to initialize meta_recommended_layer_resolution extension");
```
m4gr3d, line 74: Should be a `printerr`

devloglogan: Updated!

devloglogan:
Opening this since https://github.com/godotengine/godot/pull/99407 was merged :)

dsnopek:
@devloglogan Could you rebase on `master`? You'll need to get rid of your changes to `extension_api.json` to clear up the conflicts. And I suspect this will pass tests due to other changes on `master` too

devloglogan:
@dsnopek rebased :)

dsnopek (approved):
Thanks!

I just tested the latest version of this on Meta Quest 3 with my rendering benchmarking project, and it's working great with Godot 4.4-beta1! With recommended layer resolution turned on, I'm still seeing a steady 72 FPS (at the cost of visibly lower resolution) when going to the marker that usually has 55-60 FPS.

PR #227, opened 2024-11-21
Remove CMake from the build process

dsnopek:
Presently, we're using CMake to build the GDExtension (aka the C++ bits) on Android, but using scons to build for every other platform.

This causes a couple of problems:

- **We need to recreate what scons is already doing in CMake.** This led to some issues in the past, where I was expecting certain `#define`s to exist, because I know godot-cpp adds them, only to realize they weren't there when doing Android builds because they weren't in our custom CMake configuration here
- **To add a build option, we need to do it in both scons and CMake.** I started working on this PR because I wanted to remove Meta's headers and provide a build option to point to a local copy of the Meta headers (like we discussed at a recent XR Team meeting), only to realize there'd be added complexity for adding the option to both scons and CMake
- **CI can't cache the build artifacts from CMake leading to slower Android builds on CI.** Our current CI has caching for both scons and Gradle artifacts, but it can't cache the CMake artifacts. Perhaps there is a way to do it, but I couldn't figure it out initially, and if we switch to building with scons we'll get the already working scons caching

This PR switches to always using scons to build everything C++-related, and then Gradle just pulls those `.so`'s into the `.aar` files. This is similar to how Godot is built for Android.

This has some follow on effects:

- **We are no longer building the GDExtension multiple times for each vendor.** Previously, we rebuilt the GDExtension for each vendor with special defines, ie `META_VENDOR_ENABLED`, `PICO_VENDOR_ENABLED`, etc. We never ended up using those defines for anything, though. This will just save us build time :-)
- **We are no longer linking the `libopenxr_loader.so` to our `libgodotopenxrvendors.so`.** We never really needed to do this - we just need to make sure that `libopenxr_loader.so` gets loaded early enough. This PR switches to loading it explicitly in the Android plugin, which is working in my testing. (Note: the `libopenxr_loader.so` may be a different actual loader depending on the vendor, but it always has that same file name.)
- **The binary directories got shuffled around a bit.** If Gradle isn't building the native code itself via CMake, then `abiFilters` doesn't seem to work for excluding certain architectures in different flavors. So, I had shuffle around the directories for those binaries, so I could instead use `jniLibs.srcDirs` to point at the architectures we want. I also just removed the `armeabi-v7a` binaries for the Meta loader, because I think that's for like GearVR or Oculus Go or some other unsupported platform.

TODO:

- ~~**Add a non-functional CMake configuration that can be used to configure Android Studio for debugging.** We do this same thing for Godot.~~
- ~~**Test the plugin with other vendors and make sure it's still working!** I've tested builds from this PR on Meta Quest, and it works great. :-) I need to test on the HTC Vive XR Elite, and ensure it still works there.~~ But that's the only other Android headset I have - it'd be great to get some help with testing this from other folks with other headsets.

~~Marking as a DRAFT until those TODO's are completed.~~

**UPDATE:** I've tested with both Meta and HTC Vive XR Elite (which uses the Khronos loader), and it's now working on both of those. It'd be great if folks who have Pico, MagicLeap, Lynx, etc were able to test too, but I've done as much as I can personally.


dsnopek:
I've got this working with both Meta Quest and HTC Vive XR Elite now. I ended up having to include the Khronos loader .so files in the repo because the [one from gradle](https://central.sonatype.com/artifact/org.khronos.openxr/openxr_loader_for_android/1.0.34) is setup to be a "prefab" to use when building with CMake, and I couldn't get gradle to just pull in the .so files. Personally, I think this is fine - we're including the .so's for all the other loaders already, so this is no different.

Unfortunately, that's all the headsets I have to test! It'd be great if other folks could test the builds on other headsets too, but there isn't any more that I can do personally.

I've also added a non-functional `CMakeLists.txt` which should allow debugging via Android Studio, but I'm not super experienced with Android Studio, so I'd love feedback on if I got that right (@m4gr3d could you give it a try?)

Anyway, I'm taking this out of draft now :-)

**PS:** Look at how fast the fully cached builds are! The Android build finishes in ~2 minutes, compared to ~20 minutes before this PR.

.github/workflows/build-addon-on-push.yml
```diff
@@ -86,12 +86,21 @@ jobs:
...
           cd ..
+      - name: Create Android extension library
+        run: |
+          cd aar
+          scons platform=${{ matrix.platform }} target=template_debug arch=arm64 custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
+          scons platform=${{ matrix.platform }} target=template_release arch=arm64 custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
+          scons platform=${{ matrix.platform }} target=template_debug arch=x86_64 custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
+          scons platform=${{ matrix.platform }} target=template_release arch=x86_64 custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
+          cd ..
+        if: matrix.platform == 'android'
```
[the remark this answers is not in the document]
dsnopek: Since we have some AARs that contain _both_ arm64 and x86_64 (ex Khronos and MagicLeap), I think keeping them together, rather than making more items on the matrix makes the most sense.

Otherwise, we either end up with 3 new items on the matrix (like "android-arm64", "android-x86_64" and "android-arm64-plus-x86_64") or we have to make the workflow more complicated by having an earlier job that builds arm64 and x86_64 in parallel and then combines the artifacts in a subsequent job for Gradle to build the AARs.

The current way is simpler than either option.

SConstruct
```diff
@@ -47,6 +48,31 @@ if env["platform"] == "macos":
...
+    android_target = "release" if env["target"] == "template_release" else "debug"
+    android_arch = ""
+    if env["arch"] == "arm32":
+        android_arch = "armeabi-v7a"
+    elif env["arch"] == "arm64":
+        android_arch = "arm64-v8a"
+    elif env["arch"] == "x86_32":
+        android_arch = "x86"
+    elif env["arch"] == "x86_64":
+        android_arch = "x86_64"
```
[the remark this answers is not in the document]
dsnopek: Ok! Dropped the 32-bit ones in my latest push

rpavlik:
Regarding the Khronos loader...

> I ended up having to include the Khronos loader .so files in the repo because the [package from gradle](https://central.sonatype.com/artifact/org.khronos.openxr/openxr_loader_for_android/1.0.34) is setup to be a "prefab" to use when building with CMake, and I couldn't get gradle to just pull in the .so files.

I am confused how Gradle could pull them if they are in `jniLibs` but not if they are in e.g. `prefab/modules/openxr_loader/libs/android.arm64-v8a/` . Is there some Gradle thing using jniLibs that I wasn't aware of?

I'm happy to put them also in e.g. `jniLibs/arm64-v8a` if that helps. (it's a zip file, making a second copy of something in it doesn't take up more space.) Does this tree look OK to you?

```
.
├── jniLibs
│   ├── arm64-v8a
│   │   └── libopenxr_loader.so
│   ├── armeabi-v7a
│   │   └── libopenxr_loader.so
│   ├── x86
│   │   └── libopenxr_loader.so
│   └── x86_64
│       └── libopenxr_loader.so
└── openxr
    ├── AndroidManifest.xml
    ├── META-INF
    │   └── LICENSE
    └── prefab
        ├── modules
        │   ├── headers
        │   │   └── include
        │   │       └── openxr
        │   │           ├── openxr.h
        │   │           ├── openxr_loader_negotiation.h
        │   │           ├── openxr_platform_defines.h
        │   │           ├── openxr_platform.h
        │   │           ├── openxr_reflection.h
        │   │           ├── openxr_reflection_parent_structs.h
        │   │           └── openxr_reflection_structs.h
        │   └── openxr_loader
        │       ├── libs
        │       │   ├── android.arm64-v8a
        │       │   │   ├── abi.json
        │       │   │   └── libopenxr_loader.so
        │       │   ├── android.armeabi-v7a
        │       │   │   ├── abi.json
        │       │   │   └── libopenxr_loader.so
        │       │   ├── android.x86
        │       │   │   ├── abi.json
        │       │   │   └── libopenxr_loader.so
        │       │   └── android.x86_64
        │       │       ├── abi.json
        │       │       └── libopenxr_loader.so
        │       └── module.json
        └── prefab.json

19 directories, 23 files

```

.github/workflows/build-addon-on-push.yml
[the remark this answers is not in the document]
rpavlik: MagicLeap should be able to use the standard loader since "forever", fwiw. Didn't even know they had a separate one.

dsnopek:
> I am confused how Gradle could pull them if they are in `jniLibs` but not if they are in e.g. `prefab/modules/openxr_loader/libs/android.arm64-v8a/` . Is there some Gradle thing using jniLibs that I wasn't aware of?

If the files were in jniLibs, then you should be able to do this to pull in the .so files:

```
dependencies {
    khronosApi "org.khronos.openxr:openxr_loader_for_android:$versions.openxrVersion"
}
```

> (The only way I've seen publicly to use AARs with jniLibs in them is this: https://github.com/googlevr/cardboard/blob/master/hellocardboard-android/build.gradle#L55-L67 )

I suppose we could do something like that here too, and manually unzip the .aar file to get the files...

> I'm happy to put them also in e.g. `jniLibs/arm64-v8a` if that helps. (it's a zip file, making a second copy of something in it doesn't take up more space.) Does this tree look OK to you?

I think within the .aar file, they should actually be under `jni/arm64-v8a` (rather than `jniLibs` which is the variable used within Gradle). I'm not a Gradle expert, though :-)

.github/workflows/build-addon-on-push.yml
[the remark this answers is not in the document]
dsnopek: We have it using the Khronos loader too. I'm not sure why we build a separate AAR file for it actually? Might be historical at this point. We can discuss on the XR team to see if we want to drop it, and just direct folks to use the Khronos AAR

rpavlik:
OK, guess that's my bad... it does appear that `jni/` is the magic dir name in an aar. LMK if this one works for you, it appears to automatically trigger copying the native lib without even touching cmake or prefab. 
[openxr_loader_for_android-1.1.43.zip](https://github.com/user-attachments/files/18189063/openxr_loader_for_android-1.1.43.zip)
(rename to aar, use like `implementation files("${project.rootDir}/libs/openxr_loader_for_android-1.1.43.aar")`)

dsnopek:
@rpavlik Thanks! Unfortunately, I wasn't able to make it work. When using a local AAR file, I got errors like:

```
Direct local .aar file dependencies are not supported when building an AAR. The resulting AAR would be broken because the classes and Android resources from any local .aar file dependencies would not be packaged in the resulting AAR. Previous versions of the Android Gradle Plugin produce broken AARs in this case too (despite not throwing this error). The following direct local .aar file dependencies of the :plugin project caused this error: /home/dsnopek/Sync/Projects/default/godot_openxr_vendors/plugin/aars/openxr_loader_for_android-1.1.43.aar
```

It's possible that it could work with a non-local AAR from Maven, or maybe configuring it in a different way (I also tried using a flat file repository, which also didn't work, but maybe I'm missing some magic incantation).

Anyway, I ended up changing the PR to manually extract the AAR and copy the .so files, rather than depending on Gradle's built-in dependency handling (similar to the `hellocardboard-android` sample you linked earlier). It's not pretty, but it works, and means we don't have to keep the .so files in the repo. :-) So, I think this is probably good enough.

.github/workflows/build-addon-on-push.yml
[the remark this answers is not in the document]
m4gr3d: > by having an earlier job that builds arm64 and x86_64 in parallel

Looking at the workflow, the gradle build step does not make use of the matrix, so it can be moved outside of it similar to the asset generation step.

This would allow us to combine the following:

```
- name: Create Desktop extension library
        run: |
          cd aar
          scons platform=${{ matrix.platform }} target=template_debug ${{ matrix.flags }} custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
          scons platform=${{ matrix.platform }} target=template_release ${{ matrix.flags }} custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
          cd ..
      - name: Create Android extension library
        run: |
          cd aar
          scons platform=${{ matrix.platform }} target=template_debug arch=arm64 custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
          scons platform=${{ matrix.platform }} target=template_release arch=arm64 custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
          scons platform=${{ matrix.platform }} target=template_debug arch=x86_64 custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
          scons platform=${{ matrix.platform }} target=template_release arch=x86_64 custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
          cd ..
        if: matrix.platform == 'android'
```

into the following with the addition of an `Android X86_64` matrix:

```
- name: Create extension library
        run: |
          cd aar
          scons platform=${{ matrix.platform }} target=template_debug ${{ matrix.flags }} custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
          scons platform=${{ matrix.platform }} target=template_release ${{ matrix.flags }} custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json build_profile=thirdparty/godot_cpp_build_profile/build_profile.json
          cd ..
```

The scons steps will run ahead of time going through all the matrix permutations, then when that's completed the gradle build step will run using the result of the previous Android runs.

The current logic is kind of doing that but repeating the `Android arm64` run twice:
- The first time as part of the `Create Desktop extension library` step as it goes through the `Android arm64` matrix 
- The second time when it's run explicitly as part of the `Create Android extension library` step

plugin/build.gradle
```diff
@@ -9,6 +9,44 @@ ext {
...
+        copy {
+            from zipTree(khronosOpenxrLoader.singleFile)
+            into khronosOpenxrTempDir
+        }
+    }
+}
+
+task copyKhronosOpenxrLoader(dependsOn: extractKhronosOpenxrLoader) {
+    def khronosOpenxrTargetDirArm64 = file("../thirdparty/khronos_openxr_sdk/arm64-v8a/arm64-v8a")
+    def khronosOpenxrTargetDirX86_64 = file("../thirdparty/khronos_openxr_sdk/x86_64/x86_64")
```
m4gr3d, line 27: Should those two be temp dirs as well since we're deleting them on cleanup?

plugin/build.gradle
```diff
@@ -25,12 +63,6 @@ android {
             abiFilters "arm64-v8a"
         }
     }
-    externalNativeBuild {
-        cmake {
-            path file('CMakeLists.txt')
```
m4gr3d, line 30: We still need this so Android Studio knows which CMakeLists.txt file to use as reference.
You want to disable the build tasks tied to it though; for reference see https://github.com/godotengine/godot/blob/master/platform/android/java/nativeSrcsConfigs/build.gradle#L45-L56

plugin/build.gradle
```diff
@@ -42,47 +74,93 @@ android {
...
+            jniLibs.srcDirs = [
+                'src/main/libs/debug/arm64-v8a',
+                '../thirdparty/khronos_openxr_sdk/arm64-v8a',
+            ]
+        }
+        picoRelease {
+            jniLibs.srcDirs = [
+                'src/main/libs/release/arm64-v8a',
+                '../thirdparty/khronos_openxr_sdk/arm64-v8a',
+            ]
```
m4gr3d, line 163: Why are we not using the default `jniLibs` directories instead?
Do all the vendors have jni dependencies in the same `src/main/libs` sub-folders?

plugin/build.gradle
```diff
@@ -128,38 +206,18 @@ android {
 
     buildFeatures {
         prefab true
```
m4gr3d, line 208: We can remove `prefab` as well since we no longer make use of it.

plugin/build.gradle
```diff
@@ -128,38 +206,18 @@ android {
...
 
 task cleanAssets(type: Delete) {
     // Delete the 'addons' directory in the 'assets' folder
     delete("src/main/assets/addons")
 }
 
-task cleanCxx(type: Delete) {
-    delete(".cxx")
-}
-
```
m4gr3d, line 162: We'll still need this task.

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXR.kt
```diff
@@ -43,6 +43,13 @@ abstract class GodotOpenXR(godot: Godot?) : GodotPlugin(godot) {
 		protected val TAG = GodotOpenXR::class.java.simpleName
 
 		init {
+			try {
+				Log.v(TAG, "Loading openxr_loader library")
+				System.loadLibrary("openxr_loader")
```
m4gr3d, line 48: Let's keep them in the same `try..catch` block. With the current logic, the first `loadLibrary` could fail, and we would still attempt the second `loadLibrary`. By keeping them together, if one fails they both fail.

plugin/build.gradle
dsnopek: > Why are we not using the default `jniLibs` directories instead?

I don't think we can, because there's no libraries that are shared by every AAR. They do all include arm64 libraries, but since we have to make the distinction between debug and release, they are all different.

> Do all the vendors have jni dependencies in the same `src/main/libs` sub-folders?

They all have some dependencies from there, just not all the same ones.

plugin/build.gradle
dsnopek: That makes sense! Changed in my latest push

.github/workflows/build-addon-on-push.yml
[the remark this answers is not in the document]
dsnopek: Ah, I get what you're saying now! I'll get that into my next push (may take a couple tries because CI :-))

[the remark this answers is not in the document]
dsnopek: I have the C++ builds all split up on the matrix now, and the Gradle part is done as part of assembling the asset. So, I'm going to resolve this thread.

However, this isn't quite ready yet, because there's something else weird happening on CI that I'll describe in another comment.

dsnopek:
I think have all review addressed now, and CI doesn't fail, however, the artifacts don't look quite right and I need to figure that before this is ready.

For some reason, in the artifact built on CI, all the AARs have the Khronos loader. Locally, this doesn't happen, I can unzip the AARs and see the correct loader in there. So, I need to figure out what's going on with that.

Also, the binaries are way smaller, like 10% of their previous size. For example, the Khronos AAR goes from 28mb to 2.2mb.  The prefab stuff isn't there any more, but the biggest difference is just from the `libgodotopenxrvendors.so` getting much smaller. It's possible that the `scons` configuration is just more optimized for size than the cmake one, and everything is fine. But I need to investigate and make sure that's true.

dsnopek:
Alright, I think this is ready now!

Removing the Gradle caching from CI fixed the problem with all the AARs including the Khronos loader. I unzipped and inspected them all, and they all have the correct loader in the artifact from CI. I don't think the Gradle caching really did much for us anyway. With the caching disabled, the job that runs gradle completed in ~1 minute, so that seems fine.

Regarding the size: I think the cmake builds were so big because they couldn't use the `build_profile` feature in the scons configuration. In fact, I'm not sure why we thought it was reasonable for the relatively small amount of C++ code in this extension to result in a ~20mb binary! With the amount of code here, ~2mb sound about right.

I also tested the binaries in the artifact from CI with:

- Meta Quest 3
- Pico 4 Ultra
- HTC Vive XR Elite
- Linux (for doing the export from the Godot editor)

They all worked fine!

Unfortunately, that's only testing two loaders: the Meta one and the Khronos one. But I don't have any hardware that uses any of the other loaders.

dsnopek:
I've just rebased this on latest `master`, however, I've also updated godot-cpp in order to pull in PR https://github.com/godotengine/godot-cpp/pull/1628 which was just merged today.

That PR fixes a compile time regression in godot-cpp, and since one of the major advantages of this PR is to improve build times, I think it makes sense to pull it in here as well.

plugin/build.gradle
```diff
@@ -42,47 +81,93 @@ android {
...
-                    arguments "-DFLAVOR=pico"
-                }
-            }
+        }
+    }
+
+    sourceSets {
+        khronosDebug {
+            jniLibs.srcDirs = [
+                'src/main/libs/debug/arm64-v8a',
```
m4gr3d, line 106: This and its release version are the same for all flavors, so we can define them once in the `sourceSets` scope using:
```
debug.jniLibs.srcDirs += ['src/main/libs/debug/arm64-v8a']
release.jniLibs.srcDirs += ['src/main/libs/release/arm64-v8a']
```

The other definitions should be updated to use `+=` instead of `=`, this way you're adding to the existing list of source dirs instead of overwriting it.

m4gr3d (changes requested):

plugin/build.gradle
dsnopek: For some reason, this just doesn't work for me. When I change it to this:

```
        debug.jniLibs.srcDirs += ['src/main/libs/debug/arm64-v8a']
        release.jniLibs.srcDirs += ['src/main/libs/release/arm64-v8a']

        // ... snip ...

        metaDebug {
            jniLibs.srcDirs += [
                '../thirdparty/ovr_openxr_mobile_sdk/OpenXR/Libs/Android/Debug',
            ]
        }
        metaRelease {
            jniLibs.srcDirs += [
                '../thirdparty/ovr_openxr_mobile_sdk/OpenXR/Libs/Android/Release',
            ]
        }
```

Then the Meta .aar file _only_ includes the loader, and doesn't include the GDExtension .so file:

```
Archive:  ./plugin/build/outputs/aar/godotopenxr-meta-debug.aar
  Length      Date    Time    Name
---------  ---------- -----   ----
       76  1980-02-01 00:00   R.txt
     1152  1980-02-01 00:00   AndroidManifest.xml
    12859  1980-02-01 00:00   classes.jar
        0  1980-02-01 00:00   res/
        0  1980-02-01 00:00   res/values/
      455  1980-02-01 00:00   res/values/values.xml
        0  1980-02-01 00:00   assets/
        0  1980-02-01 00:00   assets/addons/
        0  1980-02-01 00:00   assets/addons/godotopenxrvendors/
     1294  1980-02-01 00:00   assets/addons/godotopenxrvendors/plugin.gdextension
      156  1980-02-01 00:00   META-INF/com/android/build/gradle/aar-metadata.properties
        0  1980-02-01 00:00   jni/
        0  1980-02-01 00:00   jni/arm64-v8a/
  9825808  1980-02-01 00:00   jni/arm64-v8a/libopenxr_loader.so
---------                     -------
  9841800                     14 files
```

It's like `metaDebug.jniLibs` overrides `debug.jniLibs`.

What am I doing wrong?

plugin/CMakeLists.txt
```diff
@@ -1,37 +1,32 @@
...
 
-## Common dependencies
-include(${PROJECT_SOURCE_DIR}/src/main/common.cmake)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+set(GODOT_CPP_DIR "${CMAKE_CURRENT_LIST_DIR}/../thirdparty/godot-cpp")
+set(OPENXR_HEADERS_DIR "${CMAKE_CURRENT_LIST_DIR}/../thirdparty/openxr/include")
+set(COMMON_LIB_HEADERS_DIR "${CMAKE_CURRENT_LIST_DIR}/cpp/include")
```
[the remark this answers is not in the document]
dsnopek: Updated in my latest push (along with the other requested `CMakeLists.txt` changes)

plugin/build.gradle
dsnopek: Alright, this is in my latest push now!

plugin/build.gradle
```diff
@@ -42,47 +81,86 @@ android {
...
+        picoDebug {
+            jniLibs.srcDirs += [
+                "${khronosOpenxrLoaderTargetDir}/arm64-v8a",
+            ]
+        }
+        picoRelease {
+            jniLibs.srcDirs += [
+                "${khronosOpenxrLoaderTargetDir}/arm64-v8a",
+            ]
         }
```
m4gr3d, line 164: We can consolidate the ones that repeat further

```suggestion
        debug.jniLibs.srcDirs += ['src/main/libs/debug/arm64-v8a']
        release.jniLibs.srcDirs += ['src/main/libs/release/arm64-v8a']

        khronos.jniLibs.srcDirs += [
                "${khronosOpenxrLoaderTargetDir}/arm64-v8a",
                "${khronosOpenxrLoaderTargetDir}/x86_64",
        ]
        khronosDebug.jniLibs.srcDirs += ['src/main/libs/debug/x86_64']
        khronosRelease.jniLibs.srcDirs += ['src/main/libs/release/x86_64']

        lynx.jniLibs.srcDirs += ['../thirdparty/lynx_openxr_sdk']

        magicleap.jniLibs.srcDirs += [
                "${khronosOpenxrLoaderTargetDir}/arm64-v8a",
                "${khronosOpenxrLoaderTargetDir}/x86_64",
        ]
        magicleapDebug.jniLibs.srcDirs += ['src/main/libs/debug/x86_64']
        magicleapRelease.jniLibs.srcDirs += ['src/main/libs/release/x86_64']

        metaDebug.jniLibs.srcDirs += ['../thirdparty/ovr_openxr_mobile_sdk/OpenXR/Libs/Android/Debug']
        metaRelease.jniLibs.srcDirs += ['../thirdparty/ovr_openxr_mobile_sdk/OpenXR/Libs/Android/Release']

        pico.jniLibs.srcDirs += ["${khronosOpenxrLoaderTargetDir}/arm64-v8a"]
```

SConstruct
```diff
@@ -47,6 +48,27 @@ if env["platform"] == "macos":
...
+        "{}/main/libs/{}/{}/{}/lib{}{}".format(
+            android_src_path,
+            android_target,
+            android_arch,
+            android_arch,
+            project_name,
+            env["SHLIBSUFFIX"],
+        ),
+        source=sources,
+    )
```
m4gr3d, line 71: This case breaks the gdextension library configuration for the Android editor in https://github.com/GodotVR/godot_openxr_vendors/blob/master/demo/addons/godotopenxrvendors/plugin.gdextension#L9-L12.

Without this case, builds for the Android platform fall under the following case which create the `so` file under the `binary_path` folder matching the gdextension configuration; e.g: `demo/addons/godotopenxrvendors/.bin/android/template_debug/arm64/libgodotopenxrvendors.so` for the `template_debug`, `arm64` Android build.

So we can either remove this case, and copy the `so` binaries generated by the following case to the location defined in the `build.gradle` config, or do the reverse, keep the case, and copy the `so` binaries generated under `src/main/libs` to the `demo` directory so they match the gdextension configuration.

m4gr3d (changes requested):

plugin/build.gradle
dsnopek: Ok, thanks! I've switched to your suggestion in my latest push. Unzipping the .aar files still looks good in my testing.

SConstruct
dsnopek: Ah, I didn't notice that!

In the latest push, I've updated the `SConstruct` to build the .so into the original directory, and then copy it to the place where Gradle needs it to be.

dsnopek:
@m4gr3d Thanks for the review! I've addressed both points in my latest push

PR #229, opened 2024-11-28
feat: add permission request toggles for Meta XR features

Kimau:
Add project settings to control permission requests for Meta XR features:
- Body tracking
- Eye tracking
- Face tracking
- Scene API

Each permission can now be toggled via project settings while maintaining backwards compatibility with default true values.


Kimau:
Ignore completely wrong, going to rewrite

Issue #224, opened 2024-11-02
Extension needs to execute code between xrBeginFrame and xrEndFrame


erik-kallen:
Hi!

I'm trying (like others, it seems) to implement the XR_META_environment_depth extension. I'm making decent progress, but I appear to have hit a wall:

There is a method (xrAcquireEnvironmentDepthImageMETA) that must be called between xrBeginFrame and xrEndFrame. I can't find any way to do this. _on_pre_render() is invoked before xrBeginFrame.

Is there anything I'm missing? Or is it possible to do this in some way I'm not thinking of?

dsnopek:
I think that `OpenXRExtensionWrapper::on_pre_draw_viewport()` runs in-between `xrBeginFrame()` and `xrEndFrame()`, but it looks like we aren't exposing that to `OpenXRExtensionWrapperExtension` - I'm not sure why. It should be relatively easy to expose that! I'll make a PR and see if anyone on the XR team has any objections.

dsnopek:
I just posted PR https://github.com/godotengine/godot/pull/98831 for Godot which exposes both of those virtual methods to GDExtension. Please let me know if they end up working for your use-case or not!

erik-kallen:
Thank you, though I probably won't have the time to check it out this week.

Issue #133, opened 2024-04-29
Depth API


SixArne:
Hi,

Are there any plans on implementing a depth API for meta devices?

I would love to contribute and help but I have never worked on open source before. 
(I do have a very solid knowledge of C++)

I expected to find the Meta Native SDK somewhere in the repository but I couldn't find it.

Regards,

dsnopek:
We really want to implement the Depth API, and will get around to it eventually, but no one really started to look into this yet. If you'd like to give it a try, that'd be great!

There's a possibility that this will require making some changes to Godot itself, but there's no way to tell without digging into it.

> I expected to find the Meta Native SDK somewhere in the repository but I couldn't find it.

We do include the Meta headers - you can find the header for the necessary OpenXR extension here:

https://github.com/GodotVR/godot_openxr_vendors/blob/master/thirdparty/ovr_openxr_mobile_sdk/OpenXR/Include/openxr/meta_environment_depth.h

dsnopek:
I've been researching this a tiny bit, and here's some links to resources:

- Meta's docs: https://developers.meta.com/horizon/documentation/native/android/mobile-depth
- Sample using it: https://github.com/meta-quest/Meta-OpenXR-SDK/tree/main/Samples/XrSamples/XrPassthroughOcclusion

Looking through the sample, it looks like we're expected to change our fragment shader to consult the depth texture that we get from the OpenXR extension. Here's the relevant bit: https://github.com/meta-quest/Meta-OpenXR-SDK/blob/main/Samples/XrSamples/XrPassthroughOcclusion/Src/XrPassthroughOcclusionGl.cpp#L632

In order to make this work, we'll either need to:

1. Modify Godot so that the builtin shader templates have the ability to do this when given a depth texture, and then allow GDExtension to provide that texture to Godot, or
2. Require that users modify all of their shaders to do this, and then godot_openxr_vendors just makes this depth texture available for users to do that

Nr 2 could potentially be made easier if Godot allowed users to override Godot's shader templates - see PR https://github.com/godotengine/godot/pull/94427

In any case, this won't be the easiest OpenXR extension to integrate with Godot

maunvz:
[...]

PR #230, opened 2024-12-01
[WIP] Integrate XR_META_environment_depth

maunvz:
Starting a draft for integrating XR_META_environment_depth, based on the documentation / samples:

https://developers.meta.com/horizon/documentation/native/android/mobile-depth
https://github.com/meta-quest/Meta-OpenXR-SDK/tree/main/Samples/XrSamples/XrPassthroughOcclusion

As written, the plugin can initialize the depth extension, allocate the swapchain, and receive depth buffer updates every frame. Testing on a Quest 3, debug logs showed valid OpenGL texIds, resolution as 512x512, and the swapchain index updates every few frames.

Next step is to integrate the depth buffer into the rendering pipeline so it can occlude (hole punch?) the godot scene based on environment depth info. I haven't quite figure out where / how to plug that in, or if it's even possible from an extension instead of the main engine source

### DIFF
--- a/demo/main.gd
+++ b/demo/main.gd
@@ -10,6 +10,8 @@ var hand_tracking_source: Array[OpenXRInterface.HandTrackedSource]
 @onready var left_controller_model: OpenXRFbRenderModel = $XROrigin3D/LeftHand/LeftControllerFbRenderModel
 @onready var right_controller_model: OpenXRFbRenderModel = $XROrigin3D/RightHand/RightControllerFbRenderModel
 
+var first_frame = false
+
 # Called when the node enters the scene tree for the first time.
 func _ready():
 	xr_interface = XRServer.find_interface("OpenXR")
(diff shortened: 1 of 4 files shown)


maunvz:
Some more notes: The swapchain's format / intended usage isn't exactly clear from the documentation, which just says
"The format and convention is the same as for regular OpenGL depth textures."

The fragment shader in the sample is a bit more helpful:
https://github.com/meta-quest/Meta-OpenXR-SDK/blob/main/Samples/XrSamples/XrPassthroughOcclusion/Src/XrPassthroughOcclusionGl.cpp#L641

dsnopek:
Just a couple links to some of the earlier discussion about how to integrate this extension:

- https://github.com/GodotVR/godot_openxr_vendors/issues/224
- https://github.com/GodotVR/godot_openxr_vendors/issues/133

erik-kallen:
@maunvz I also made a draft PR (#232) for this. I don't think we should compete, but I started work before this PR was up

maunvz:
@erik-kallen oh nice, looks like yours is farther along as well. I'll leave some notes/comments on there if you plan to keep working on this, I probably won't have time to make more progress on mine in the near future

erik-kallen:
I think they are in similar states, since neither of them actually works. I won't have time to work on it for the rest of the year, either.

PR #232, opened 2024-12-06
Another WIP for the FB_environment_depth extension

erik-kallen:
Sorry for opening another PR for this, but I had already started working when the other appeared, and I believe it adds some value to have it.

This is my (unfortunately not working) attempt to integrate the FB_environment_depth extension. My idea is that the extension will populate global shader parameters, which can then be used in custom shaders. Perhaps it could be useful to somehow use this to prepopulate the depth buffer (at least as an option), but that would be far above what I am capable of doing.

Requires Godot 4.4.

More information in inline comments.

### DIFF
--- a/demo/depth.gdshader
+++ b/demo/depth.gdshader
@@ -1,0 +1,23 @@
+shader_type spatial;
+
+global uniform sampler2DArray environment_depth_texture;
+global uniform mat4 environment_depth_view_projection_left;
+global uniform mat4 environment_depth_view_projection_right;
+global uniform mat4 environment_depth_projection_projection_left;
+global uniform mat4 environment_depth_projection_projection_right;
+
+void vertex() {
+	// Called for every vertex the material is visible on.
+}
+
+void fragment() {
+	// Called for every pixel the material is visible on.
+	float depth = texture(environment_depth_texture, vec3(UV, float(VIEW_INDEX))).r;
+	ALBEDO = vec3(depth);
+	//ALBEDO = vec3(1, 1, 1);
+}
+
+//void light() {
+	// Called for every pixel for every light affecting the material.
+	// Uncomment to replace the default light processing function with this one.
+//}
--- a/demo/main.gd
+++ b/demo/main.gd
@@ -10,6 +10,8 @@ var hand_tracking_source: Array[OpenXRInterface.HandTrackedSource]
 @onready var left_controller_model: OpenXRFbRenderModel = $XROrigin3D/LeftHand/LeftControllerFbRenderModel
 @onready var right_controller_model: OpenXRFbRenderModel = $XROrigin3D/RightHand/RightControllerFbRenderModel
 
+var depth_extension_wrapper
+
 # Called when the node enters the scene tree for the first time.
 func _ready():
 	xr_interface = XRServer.find_interface("OpenXR")
@@ -22,6 +24,20 @@ func _ready():
 	for hand in OpenXRInterface.HAND_MAX:
 		hand_tracking_source[hand] = xr_interface.get_hand_tracking_source(hand)
 
+	if Engine.has_singleton("OpenXRFbEnvironmentDepthExtensionWrapper"):
+		await await get_tree().create_timer(1.0).timeout
+		
+		depth_extension_wrapper = Engine.get_singleton("OpenXRFbEnvironmentDepthExtensionWrapper")
+		print(depth_extension_wrapper.prepare());
+		print(depth_extension_wrapper.supports_environment_depth)
+		print(depth_extension_wrapper.supports_hand_removal)
+		depth_extension_wrapper.enable_hand_removal = true
+		depth_extension_wrapper.started = true
+
+		#await await get_tree().create_timer(5.0).timeout
+		#depth_extension_wrapper.started = false
+	else:
+		print("No wrapper :(")
 
 func _on_session_stopping() -> void:
 	if "--xrsim-automated-tests" in OS.get_cmdline_user_args():
@@ -55,7 +71,6 @@ func _physics_process(_delta: float) -> void:
 					right_hand.pose = "grip"
 
 		hand_tracking_source[hand] = source
-
 
 func _on_left_controller_fb_render_model_render_model_loaded() -> void:
 	left_hand_mesh.hide()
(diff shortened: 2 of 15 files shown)


demo/depth.gdshader
```diff
@@ -0,0 +1,23 @@
+shader_type spatial;
+
+global uniform sampler2DArray environment_depth_texture;
```
erik-kallen, line 3: The PR does not work at all, and I suspect this to be the reason. The texture has a DEPTH_COMPONENT16 format, so I guess it should be used with a shadow sampler. But I can't figure out if and how Godot supports those.

demo/main.gd
```diff
@@ -22,6 +24,20 @@ func _ready():
 	for hand in OpenXRInterface.HAND_MAX:
 		hand_tracking_source[hand] = xr_interface.get_hand_tracking_source(hand)
 
+	if Engine.has_singleton("OpenXRFbEnvironmentDepthExtensionWrapper"):
+		await await get_tree().create_timer(1.0).timeout
```
erik-kallen, line 28: For some reason, the game crashes when deployed to my Quest 3 if the extension is started too early. This will obviously need to be handled better in the future.

plugin/src/main/cpp/extensions/openxr_fb_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,625 @@
...
+#include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/classes/rendering_server.hpp>
+#include <godot_cpp/templates/local_vector.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+#include <openxr/openxr_platform.h>
+
+#include "extensions/openxr_fb_environment_depth_extension_wrapper.h"
+
+/////////////// xr_linear.h
```
erik-kallen, line 40: I took some math code from an official sample, and that code requires this OpenXR header. I don't know if it isn't included in the project for a reason, but the math can probably be done with Godot's math classes instead, but that will be a later problem.

plugin/src/main/cpp/extensions/openxr_fb_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,625 @@
...
+	GDEXTENSION_INIT_XR_FUNC_V(xrGetEnvironmentDepthSwapchainStateMETA);
+	GDEXTENSION_INIT_XR_FUNC_V(xrDestroyEnvironmentDepthSwapchainMETA);
+	GDEXTENSION_INIT_XR_FUNC_V(xrEnumerateEnvironmentDepthSwapchainImagesMETA);
+	GDEXTENSION_INIT_XR_FUNC_V(xrAcquireEnvironmentDepthImageMETA);
+
+	return true;
+}
+
+bool OpenXRFbEnvironmentDepthExtensionWrapper::prepare() {
+	// TODO: Verify that the shader parameters exist, if possible
```
erik-kallen, line 156: It would possibly be nice if the extension could create the shader parameters, but they need to exist when editing the shader. Would at least be nice to check if they exist, but the relevant methods seem to only be intended from use in the editor.

plugin/src/main/cpp/extensions/openxr_fb_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,625 @@
...
+		UtilityFunctions::print("xrEnumerateEnvironmentDepthSwapchainImagesMETA failed to get count!");
+		UtilityFunctions::print(get_openxr_api()->get_error_string(result));
+		xrDestroyEnvironmentDepthSwapchainMETA(swapchain);
+		swapchain = nullptr;
+		xrDestroyEnvironmentDepthProviderMETA(provider);
+		provider = nullptr;
+		return false;
+	}
+
+	struct XrSwapchainImageOpenGLKHR {
```
erik-kallen, line 223: This is fishy. How do I know which graphics API to use? With trial and error I have determined that it is OpenGL (not ES, as I thought), at least when running it through Quest Link.

Should there be some #ifdefs here? Or can it be checked at runtime? Though if it is Vulkan (or Direct3D, but I assume that is not supported), the struct layout is different. But I assume they can be converted to the same RIDs.

demo/depth.gdshader
```diff
@@ -0,0 +1,23 @@
...
+global uniform mat4 environment_depth_view_projection_left;
```
erik-kallen, line 4: Preferrably these would instead be array of matrices. But that does not seem to be possible, at least not in any way I have managed to find.

plugin/src/main/cpp/extensions/openxr_fb_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,625 @@
...
+		provider = nullptr;
+		return false;
+	}
+
+	swapchain_image_rids.clear();
+	for (int i = 0; i < image_count; i++) {
+		swapchain_image_rids.push_back(
+			RenderingServer::get_singleton()->texture_create_from_native_handle(
+				RenderingServer::TextureType::TEXTURE_TYPE_LAYERED,
+				Image::Format::FORMAT_RH, // GL_DEPTH_COMPONENT16 according to the docs
```
erik-kallen, line 254: Is this correct? The image is GL_DEPTH_COMPONENT16 (OpenGL) or VK_FORMAT_D16_UNORM (for Vulkan)

plugin/src/main/cpp/extensions/openxr_fb_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,625 @@
...
+		return;
+	}
+
+	double near_z = depth_image.nearZ;
+	double far_z = depth_image.farZ;
+
+	Projection view_projections[2];
+	Projection projection_projections[2];
+
+	for (int i = 0; i < 2; ++i) {
```
erik-kallen, line 375: This is the code I needed xr_linear.h for. Very much not finished.

erik-kallen:
I won't be doing anything more with this, so closing

dsnopek:
Superseded by https://github.com/GodotVR/godot_openxr_vendors/pull/273

PR #234, opened 2024-12-16
Make Settings for the individual extensions

Kimau:
You often don't want to use all the extensions, so this adds project settings to control which extensions are enabled for the addon.

I've found it useful for my current game.

### DIFF
--- a/plugin/src/main/cpp/register_types.cpp
+++ b/plugin/src/main/cpp/register_types.cpp
@@ -84,9 +84,101 @@
 
 using namespace godot;
 
+static void define_global_bool(ProjectSettings *project_settings, const String &p_name, bool p_default_value) {
+	if (!project_settings->has_setting(p_name)) {
+		project_settings->set_setting(p_name, p_default_value);
+	}
+
+	project_settings->set_initial_value(p_name, true);
+	project_settings->set_as_basic(p_name, false);
+	Dictionary property_info;
+	property_info["name"] = p_name;
+	property_info["type"] = Variant::Type::BOOL;
+	property_info["hint"] = PROPERTY_HINT_NONE;
+	project_settings->add_property_info(property_info);
+}
+
+static bool global_bool_get(const String &p_name) {
+	ProjectSettings *project_settings = ProjectSettings::get_singleton();
+	if (project_settings == nullptr) {
+		return false;
+	}
+
+	return project_settings->get_setting_with_override(p_name);
+}
+
+static void add_plugin_core_settings() {
+	ProjectSettings *project_settings = ProjectSettings::get_singleton();
+	if (project_settings == nullptr) {
+		return;
+	}
+
+	define_global_bool(project_settings, "xr/openxr/openxr/hybrid_app", false);
+
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_body_tracking", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_composition_layer_alpha_blend", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_composition_layer_secure_content", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_composition_layer_settings", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_face_tracking", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_hand_tracking_aim", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_hand_tracking_capsules", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_hand_tracking_mesh", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_passthrough", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_render_model", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_scene_capture", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_scene", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_spatial", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/htc_facial_tracking", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/htc_passthrough", true);
+}
+
+void add_plugin_project_settings() {
+	ProjectSettings *project_settings = ProjectSettings::get_singleton();
+	if (project_settings == nullptr) {
+		return;
+	}
+
+	{
+		// Add the 'automatically_request_runtime_permissions' project setting
+		String request_permissions_setting = "xr/openxr/extensions/automatically_request_runtime_permissions";
+		if (!project_settings->has_setting(request_permissions_setting)) {
+			// Default value is `true` to match prior plugin behavior
+			project_settings->set_setting(request_permissions_setting, true);
+		}
+
+		project_settings->set_initial_value(request_permissions_setting, true);
+		project_settings->set_as_basic(request_permissions_setting, false);
+		Dictionary property_info;
+		property_info["name"] = request_permissions_setting;
+		property_info["type"] = Variant::Type::BOOL;
+		property_info["hint"] = PROPERTY_HINT_NONE;
+		project_settings->add_property_info(property_info);
+	}
+
+	{
+		String collision_shape_2d_thickness = "xr/openxr/extensions/meta_scene_api/collision_shape_2d_thickness";
+		if (!project_settings->has_setting(collision_shape_2d_thickness)) {
+			project_settings->set_setting(collision_shape_2d_thickness, 0.1);
+		}
+
+		project_settings->set_initial_value(collision_shape_2d_thickness, 0.1);
+		project_settings->set_as_basic(collision_shape_2d_thickness, false);
+		Dictionary property_info;
+		property_info["name"] = collision_shape_2d_thickness;
+		property_info["type"] = Variant::Type::FLOAT;
+		property_info["hint"] = PROPERTY_HINT_NONE;
+		project_settings->add_property_info(property_info);
+	}
+
+	OpenXRFbHandTrackingAimExtensionWrapper::add_project_setting(project_settings);
+	OpenXRMetaRecommendedLayerResolutionExtensionWrapper::add_project_setting(project_settings);
+}
+
 void initialize_plugin_module(ModuleInitializationLevel p_level) {
 	switch (p_level) {
 		case MODULE_INITIALIZATION_LEVEL_CORE: {
+			add_plugin_core_settings();
+
 			ClassDB::register_class<OpenXRFbPassthroughExtensionWrapper>();
 			OpenXRFbPassthroughExtensionWrapper::get_singleton()->register_extension_wrapper();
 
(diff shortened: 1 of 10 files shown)


plugin/src/main/cpp/register_types.cpp
```diff
@@ -81,74 +81,180 @@
...
-			OpenXRFbBodyTrackingExtensionWrapper::get_singleton()->register_extension_wrapper();
+	return project_settings->get_setting_with_override(p_name);
+}
 
-			ClassDB::register_class<OpenXRFbHandTrackingMeshExtensionWrapper>();
-			OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->register_extension_wrapper();
 
-			ClassDB::register_class<OpenXRFbHandTrackingAimExtensionWrapper>();
-			OpenXRFbHandTrackingAimExtensionWrapper::get_singleton()->register_extension_wrapper();
+void add_plugin_core_settings() {
```
dsnopek, line 109: I think we probably want all these internal functions to be `static`?

```suggestion
void static add_plugin_core_settings() {
```

plugin/src/main/cpp/register_types.cpp
```diff
@@ -81,74 +81,180 @@
...
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_scene_capture", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_scene", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/fb_spatial", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/htc_facial_tracking", true);
+	define_global_bool(project_settings, "xr/openxr/extensions/htc_passthrough", true);
+}
 
-			ClassDB::register_class<OpenXRFbCompositionLayerSecureContentExtensionWrapper>();
-			OpenXRFbCompositionLayerSecureContentExtensionWrapper::get_singleton()->register_extension_wrapper();
+void add_plugin_project_settings() {
```
dsnopek, line 133:
```suggestion
static void add_plugin_project_settings() {
```

dsnopek (commented):
Thanks!

At a high-level, this is a really great idea! I'll try to test this when I have a moment.

Which Godot versions have you tried it with? I suspect this won't work with Godot 4.3 and earlier, because the `ProjectSettings` singleton isn't available early enough. PR https://github.com/godotengine/godot/pull/98862 was recently merged to Godot `master` (and I personally tested it with a use case like your PR :-)), so I think this may need to be Godot 4.4+

plugin/src/main/cpp/register_types.cpp
```diff
@@ -158,32 +264,66 @@ void initialize_plugin_module(ModuleInitializationLevel p_level) {
...
-			ClassDB::register_class<OpenXRFbSpatialEntityUser>();
-			ClassDB::register_class<OpenXRFbPassthroughGeometry>();
-			ClassDB::register_class<OpenXRMetaPassthroughColorLut>();
-
-			OpenXRFbHandTrackingAimExtensionWrapper::get_singleton()->add_project_setting();
+			if (global_bool_get("xr/openxr/extensions/fb_passthrough")) {
+				Engine::get_singleton()->register_singleton("OpenXRFbPassthroughExtensionWrapper", OpenXRFbPassthroughExtensionWrapper::get_singleton());
+				ClassDB::register_class<OpenXRFbPassthroughGeometry>();
+				ClassDB::register_class<OpenXRMetaPassthroughColorLut>();
+			}
```
dsnopek, line 271: For the node types, I don't think we should make registering them conditional on the project setting, otherwise you have to restart the Godot editor for the node types to become available to use. So, I think this one, and all the following shouldn't be checking project settings

There may be an argument for always registering the classes for the extension wrappers also just not calling `register_extension_wrapper()`, so that the editor is aware that the classes exist, and what their methods are, because a couple of those singletons have public methods.

Kimau: yeah I need to dig into this, check all the safety checks are there

Kimau:
So because this requires 4.4 should we branch this or hold it off?

It's fine on my local branch but not sure if we want to restrict this to 4.4? Is there a version define we can do to make this back compat?

dsnopek:
> So because this requires 4.4 should we branch this or hold it off?
> 
> It's fine on my local branch but not sure if we want to restrict this to 4.4? Is there a version define we can do to make this back compat?

It's just something to keep in mind when we're thinking about the timing of merging this. Like, just to consciously decided when `master` is no longer Godot 4.3-compatible, and not accidentally merge something that requires 4.4 before we mean to. However, I guess our CI seems to have caught it in this case anyway :-)

m4gr3d:
I'm concerned about this approach as if not done right, this would be another mechanism for users to shoot themselves in the foot..
We have extensions that depend on each other (e.g: `fb_scene_extension` & `fb_spatial_entity_extension`), but the proposed mechanism does not take this into account, which means that as a user, I could enable `fb_scene_extension` and disable `fb_spatial_entity_extension` and be confused about why the former is not working.

The proper mechanism IMO would be to group related extensions under a common setting, so you enable / disable all related settings together. However this would increase the complexity and maintenance burden of this feature..

Kimau:
It would be a nice improvement with error messages or similar. Not too bothered as I've catered to my needs lovely but I without the ability to turn off stuff you don't need honestly this addon becomes bloat that is unusable in serious projects.

The question is if we need to make all those improvements in this PR or in a later one. I think the maintenance burden is required if you want to keep it friendly. 

Maybe in each extension wrapper we add a dependency macro? To enforce enable and order of init

dsnopek:
Personally, I think implementing this is worth the maintenance burden.

Right now, we enable all extensions, but don't necessarily use them. Imagine we implemented all the vendor extensions supported by Meta, for example: that's like ~50 extensions. On the Godot side, this means registering ~50 extension wrappers in Godot, which all get their "hooks" invoked (ie `_on_process()`, `_on_pre_render()`, etc). And also internally within the OpenXR runtime, it's going to initialize everything for all ~50 extensions to be used.

If a given app is only going to use, for example, 2 of those extensions, it's much cleaner to just not enable the other ~48.

The biggest downside to this in my mind is that, up until now, we've been primarily using export settings for this, and now we end up project settings too. We'll need to figure out how we want to split these things up between the two places.

Kimau:
Just registering you want an extension enabled to the runtime on boot costs memory and perf.
The issue on using export settings is it leads to a false testing experience for non exporting builds which increase the "release only" style bug

BastiaanOlij:
Looks good, few things that were discussed during the XR meeting:
- classes should always be registered, even when something is not enabled, because Godot doesn't like it when classes disappear but you may still have code referencing it. Registering the classes in ClassDB does not result in them actually being used.
- need to look at whether we're too granular or if we should group some of the settings. A lot of our implementation requires extensions to be enabled together due to dependencies.
- need to see if we can remove some of the export settings and have them just react to the new project settings. Here we should also introduce feature tags for each export platform so you can override an project setting for a specific platform (possibly in a separate PR).

But overall, this is great!

Kimau:
Sorry I've been missing the meetings. Life is hectic and I've been head down and they are late for me.

Agree on the classes thing.
Overall I think the granularity is required to be taken seriously. The OpenXR spec already bundles stuff pretty decently most of the time. 
Though we could add more editor level logic with tooltips and automatic ticking unticking of deps. 

Will try make time for this but unlikely for a couple of weeks

Kimau:
Cleaned up so all the classes are registered always

plugin/src/main/cpp/register_types.cpp.orig
```diff
@@ -0,0 +1,379 @@
+/**************************************************************************/
```
dsnopek, line 1: Looks like this `.orig` file accidentally got commited

plugin/src/main/cpp/register_types.cpp
```diff
@@ -157,33 +245,67 @@ void initialize_plugin_module(ModuleInitializationLevel p_level) {
...
+
+			if (global_bool_get("xr/openxr/extensions/fb_scene")) {
+				Engine::get_singleton()->register_singleton("OpenXRFbSceneExtensionWrapper", OpenXRFbSceneExtensionWrapper::get_singleton());
+			}
+
+			if (global_bool_get("xr/openxr/extensions/fb_hand_tracking_aim")) {
+				Engine::get_singleton()->register_singleton("OpenXRFbHandTrackingAimExtensionWrapper", OpenXRFbHandTrackingAimExtensionWrapper::get_singleton());
+
+				OpenXRFbHandTrackingAimExtensionWrapper::get_singleton()->add_project_setting();
```
dsnopek, line 276: We shouldn't conditionally add project settings based on other project settings, that won't really work. This method needs to be run always

plugin/src/main/cpp/register_types.cpp
```diff
@@ -157,33 +245,67 @@ void initialize_plugin_module(ModuleInitializationLevel p_level) {
...
+			}
 
-			ClassDB::register_class<OpenXRFbRenderModel>();
+			if (global_bool_get("xr/openxr/extensions/fb_render_model")) {
+				Engine::get_singleton()->register_singleton("OpenXRFbRenderModelExtensionWrapper", OpenXRFbRenderModelExtensionWrapper::get_singleton());
+			}
+
+			if (global_bool_get("xr/openxr/extensions/fb_scene_capture")) {
+				Engine::get_singleton()->register_singleton("OpenXRFbSceneCaptureExtensionWrapper", OpenXRFbSceneCaptureExtensionWrapper::get_singleton());
+			}
```
dsnopek, line 260: I don't know about separating `fb_scene` and `fb_scene_capture`. That seems a little too granular?

And for that matter, you can't really do anything with `fb_scene` or `fb_scene_capture` without `fb_spatial` - at best you can have the user do a scene capture, but you can't access the data. Should these all be grouped together under `fb_spatial`?

plugin/src/main/cpp/register_types.cpp
dsnopek: This one still needs to be addressed

BastiaanOlij:
Discussing this further in the XR meeting, we feel that grouping things more probably doesn't help much in this case, there are some core extension that this does apply to. We're already grouping the spatial extension for instance here, the rest probably can remain gradular.

@m4gr3d and @dsnopek , if you can give this another look to see if we're happy we can move forward

@Kimau if you can remove the duplicate extension and rebase, we can probably move forward with this.

plugin/src/main/cpp/register_types.cpp.orig
Kimau: removed that

plugin/src/main/cpp/register_types.cpp
Kimau: Good shout

plugin/src/main/cpp/register_types.cpp
Kimau: agreed reworked things to be static so it doesn't require instantiation.

Kimau:
Reworked some bits with registering extensions took a bit longer than expected to resolve some merge problems.
Bit of merge nightmare to track through. Also I made the project settings stuff static to avoid instantiations.

plugin/src/main/cpp/register_types.cpp
```diff
@@ -167,36 +257,66 @@ void initialize_plugin_module(ModuleInitializationLevel p_level) {
...
-			ClassDB::register_class<OpenXRFbPassthroughGeometry>();
 			ClassDB::register_class<OpenXRMetaPassthroughColorLut>();
-
 			ClassDB::register_class<OpenXRHybridApp>();
-			Engine::get_singleton()->register_singleton("OpenXRHybridApp", OpenXRHybridApp::get_singleton());
 
-			OpenXRFbHandTrackingAimExtensionWrapper::get_singleton()->add_project_setting();
-			OpenXRMetaRecommendedLayerResolutionExtensionWrapper::get_singleton()->add_project_setting();
+			if (global_bool_get("xr/openxr/extensions/fb_passthrough")) {
+				Engine::get_singleton()->register_singleton("OpenXRFbPassthroughExtensionWrapper", OpenXRFbPassthroughExtensionWrapper::get_singleton());
```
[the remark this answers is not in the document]
Kimau: By registering the singleton you are forcing a bunch of memory alloc though. 
Surely you want the gdscript to break in this case?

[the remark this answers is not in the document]
dsnopek: > By registering the singleton you are forcing a bunch of memory alloc though.

The extension wrappers themselves don't take much memory, and they won't actually do anything if you don't call `register_extension_wrapper()`.

> Surely you want the gdscript to break in this case?

No, I don't think we ever want GDScript to break in this way.

This means the existence of these singletons in the editor will depend on the project setting AND the registration only changes at editor startup time. So, if you had the setting disabled when you started the editor, and then decided you wanted to use one of these singletons, so you enable the setting - if you typed the name of the singleton in the script editor, it would be a error, because the singleton doesn't exist. You would need to restart the editor for the singleton to be valid.

Unless we're going to force an editor restart any time these settings are changed? There is a mechanism for that, but I'd prefer to avoid using it.

Kimau:
Closing up as it sounds like @dsnopek version is more aligned.

PR #237, opened 2024-12-22
Add option to enable permissions necessary for sharing of Meta's spatial anchors

dsnopek:
It turns out that the `com.oculus.permission.IMPORT_EXPORT_IOT_MAP_DATA` permission is necessary for users of Meta's spatial anchors to share anchors with other users (for making co-located multiplayer experiences).

See https://developers.meta.com/horizon/documentation/native/android/openxr-spatial-anchors-developer-setup/

This PR adds a new export option, which, if enabled, will add this permission.

### DIFF
--- a/plugin/src/main/cpp/export/meta_export_plugin.cpp
+++ b/plugin/src/main/cpp/export/meta_export_plugin.cpp
@@ -124,6 +124,15 @@ MetaEditorExportPlugin::MetaEditorExportPlugin() {
 			PROPERTY_USAGE_DEFAULT,
 			false,
 			false);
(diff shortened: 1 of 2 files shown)


m4gr3d (approved):
Do you know if that's a runtime permission we need to prompt the user for?

dsnopek:
> Do you know if that's a runtime permission we need to prompt the user for?

Just adding this to the `AndroidManifest.xml` was enough to get it working in my testing.

(I've been slowly working on a co-located multiplayer experience as a side project for a few months :-))

PR #241, opened 2025-01-06
Fix wall, floor and ceiling collision shapes with Jolt physics

dsnopek:
When running the Meta Scene Sample with Jolt physics, it gets errors like this:

```
01-06 11:49:51.772 10894 11064 E godot   : ERROR: Failed to build Jolt Physics box shape with {half_extents=(2.301814, 0.000000, 1.808793) margin=0.040000}. It returned the following error: 'Invalid convex radius'. This shape belongs to 'StaticBody3D:<StaticBody3D#39040583057>' and 0 other object(s).
01-06 11:49:51.772 10894 11064 E godot   :    at: _build (modules/jolt_physics/shapes/jolt_box_shape_3d.cpp:45)
```

This is because we're using `BoxShape3D`s for the 2D spatial entities (like walls, ceiling and floor) but using a `0` for the Y size.

~~This PR fixes the issue by using `0.001` instead.~~

**UPDATE:** In order to prevent tunneling, I've decided to make the `BoxShape3D` be `0.1` thick instead, and moved the shape back, so that one side is flush with where the plane should be. Most walls are thicker than 10cm, but we can't really be sure what things will get a 2D shape, so this feels like a safe-ish default to me. I suppose we could make it configurable? Or, developers can grab the data and generate their own shapes as well, so perhaps that's not necessary

### DIFF
--- a/plugin/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
+++ b/plugin/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
@@ -36,6 +36,7 @@
 #include <godot_cpp/classes/concave_polygon_shape3d.hpp>
 #include <godot_cpp/classes/mesh_instance3d.hpp>
 #include <godot_cpp/classes/plane_mesh.hpp>
+#include <godot_cpp/classes/project_settings.hpp>
 #include <godot_cpp/classes/surface_tool.hpp>
 #include <godot_cpp/templates/local_vector.hpp>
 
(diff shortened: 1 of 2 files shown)


plugin/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
```diff
@@ -353,18 +353,20 @@ Node3D *OpenXRFbSpatialEntity::create_collision_shape() const {
 
 		return collision_shape;
 	} else if (is_component_enabled(COMPONENT_TYPE_BOUNDED_2D)) {
+		const float COLLISION_SHAPE_2D_THICKNESS = 0.1;
```
BastiaanOlij, line 356: Would be good to take the precaution and have this be a project settings.

BastiaanOlij (approved):
Cool! LGTM

plugin/src/main/cpp/classes/openxr_fb_spatial_entity.cpp
dsnopek: I've added a project setting for this, which defaults to `0.1`

PR #243, opened 2025-01-11
Emit correct signal from `OpenXRFbSpatialEntity.erase_from_storage()`

dsnopek:
Calling `OpenXRFbSpatialEntity.erase_from_storage()` leads to the saved signal emitting, rather than the erased signal.

Looks like a copy-paste error, using the `_on_save_to_storage()` callback rather than the correct `_on_erase_from_storage()` callback.


m4gr3d (approved):

m4gr3d:
@dsnopek Since we haven't create the `3.x` branch, or merged any of the `4.x` PRs, this should go into a 3.1.1 milestone.

We should just generate a new 3.1.1 release as soon as we merge this PR. In the process, we should fix the issue with the `MavenCentral Publish` workflow.

dsnopek:
I'm going to go ahead and merge this just so it doesn't get forgotten. We have quite the collection of PRs that are Godot 4.4 only that are almost ready, and I don't want one of those to accidentally get merged before this one.

I've got an idea for addressing the "MavenCentral Publish" workflow, which I'll put in another PR.

PR #244, opened 2025-01-13
Add performance settings extension

Kimau:


devloglogan:
I've opened the upstream PR exposing the same functionality, as well as signals exposing the performance level notifications: https://github.com/godotengine/godot/pull/101597

I'll go ahead and close this PR since I believe (per here and RC) we are all in agreement that this will live in upstream. But thank you again for taking the time to make this PR!

PR #245, opened 2025-01-15
Support making hybrid apps for Meta headsets

dsnopek:
This is moved over from https://github.com/godot-sdk-integrations/godot-meta-toolkit/pull/1, and reworked in such a way that it should be easy to support other headset vendors (but this PR only has Meta support). In fact, since https://github.com/godotengine/godot/pull/101387 works, I think we should be able to make this work on Pico as well, ~~but I haven't tried it yet.~~

**UPDATE:** I made PR https://github.com/GodotVR/godot_openxr_vendors/pull/246 that builds on this one to add support for Pico headsets.

**NOTE:** This PR only works with Godot v4.4-beta1 and later (and so I've put this on the `4.x` milestone).

You can use this export setting to make a hybrid app:

Selection_261

You can choose **Start As Immersive** or **Start As Panel**.

And it provides this API for developers (no longer built in GDScript - it's now a singleton implemented in C++):

```
# Returns true or false for if this is a hybrid app
OpenXRHybridApp.is_hybrid_app()

# Returns an enum with values: HYBRID_MODE_NONE, HYBRID_MODE_IMMERSIVE or HYBRID_MODE_PANEL
OpenXRHybridApp.get_mode()

# Allows switching modes and passing a string as "launch data"
OpenXRHybridApp.switch_mode(OpenXRHybridApp.HYBRID_MODE_PANEL, '{"info": "the info!"}')

# And this function lets you get the "launch data"
OpenXRHybridApp.get_launch_data()
```

I've included a sample app, with the "famous" rotating turkey. The turkey's current rotation is passed as "launch data" when switching modes, so it'll start from the same rotation that it left off on.


plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRHybridAppInternal.kt
```diff
@@ -0,0 +1,105 @@
...
+
+		return null
+	}
+
+	@UsedByGodot
+	fun hybridAppSwitchTo(mode: Int, data: String = ""): Boolean {
+		if (hybridMode == mode) return false
+
+		val context = getActivity()
+		if (context == null) return false
```
m4gr3d, line 79: nit: In Kotlin, you can use the following instead:

```
val context = getActivity() ?: return false
```

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRHybridAppInternal.kt
```diff
@@ -0,0 +1,105 @@
...
+
+		val activityName = if (mode == 0) IMMERSIVE_ACTIVITY else PANEL_ACTIVITY
```
m4gr3d, line 81: nit: should be `if (mode == IMMERSIVE_MODE)`

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRHybridAppInternal.kt
```diff
@@ -0,0 +1,105 @@
...
+		private const val PANEL_ACTIVITY = "org.godotengine.openxr.vendors.GodotPanelApp"
+
+		private const val IMMERSIVE_MODE = 0
+		private const val PANEL_MODE = 1
+
+		private const val INTENT_EXTRA_DATA = "godot_openxr_vendors_data"
+	}
+
+	private var hybridMode: Int = IMMERSIVE_MODE
+	private var hybridLaunchData: String? = null
```
m4gr3d, line 58: Let's make this charitable non-null and have the default value be `""`

plugin/src/main/cpp/export/export_plugin.cpp
```diff
@@ -90,6 +90,22 @@ Dictionary OpenXREditorExportPlugin::_get_vendor_toggle_option(const String &ven
 			false);
 }
 
+String OpenXREditorExportPlugin::_get_hybrid_app_option_name(const String &p_prefix) const {
+	return p_prefix + String("/hybrid_app");
```
m4gr3d, line 94: This may be a sign we should be using a project setting instead of an export option.. 
With the latter, we need a different export option for each censor and if we add more hybrid options, these options will also repeat for each censor requiring users to check them all for each vendor they support.
With the former, it's the opposite.. one set of project settings for all supported vendors with the flexibility to make vendor specific changes via feature flags.

plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -237,6 +238,22 @@ TypedArray<Dictionary> MetaEditorExportPlugin::_get_export_options(const Ref<Edi
...
+	}
+
+	HybridType hybrid_type = _get_hybrid_app_option_value();
+	if (hybrid_type != HYBRID_TYPE_DISABLED) {
+		// Unless this is a hybrid app that launches as a panel, then we want to force this option on.
+		// Otherwise, it needs to be off, so that the panel activity can be the one that's launched by default.
+		overrides["package/show_in_app_library"] = (hybrid_type != HYBRID_TYPE_START_AS_PANEL);
+	}
+
+	return overrides;
```
m4gr3d, line 254: Seems this logic should be in the base class since I would expect it to be the same for all supported vendors.

plugin/src/main/cpp/classes/openxr_hybrid_app.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+String OpenXRHybridApp::get_launch_data() const {
+	if (!is_hybrid_app()) {
+		return "";
+	}
+
+	Object *godot_openxr_singleton = Engine::get_singleton()->get_singleton("GodotOpenXRHybridAppInternal");
+	if (!godot_openxr_singleton) {
+		return String();
```
m4gr3d, line 93: nit: for consistency, we can also return `""` here

plugin/src/main/cpp/export/meta_export_plugin.cpp
dsnopek: So, I tried this, but it's a little awkward when the child class also wants to implement `_get_export_options_overrides()`, because both need to do the same `if (!_supports_platform(platform) || !_is_vendor_plugin_enabled())` logic.

I suppose that's not really a big deal, since none of this code runs very often (only in the export dialog or during an actual export). I can switch back to that.

plugin/src/main/cpp/export/export_plugin.cpp
m4gr3d: Note that even if we use a project setting, we should keep the detection logic the same, i.e: keep relying on feature tags, as this would help cover certain use-cases.

For example, consider a developer with an existing flat game that wants to turn his game into a hybrid app when running on XR devices; here's the flow I see the developer following:
- Turn on the OpenXR project setting and configure as needed
- Turn on the hybrid app project setting and configure as needed
- ... development ...
- Create an export preset for non-XR devices; that preset will set `xr_features/xr_mode` to `regular` which will disable all the vendors plugin export logic. This will ensure that on non-XR devices, only the panel version of the project is run with no access to hybrid capabilities
- Create an export preset for XR devices; that preset will set `xr_features/xr_mode` to `openxr` which will enable all the vendors plugin export logic, including the addition of the hybrid feature tag

plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -475,6 +497,34 @@ String MetaEditorExportPlugin::_get_android_manifest_application_element_content
 		contents += "        <meta-data android:name=\"com.oculus.ossplash.background\" android:value=\"passthrough-contextual\" />\n";
 	}
 
+	HybridType hybrid_type = _get_hybrid_app_option_value();
+	if (hybrid_type != HYBRID_TYPE_DISABLED) {
+		contents +=
+				"        <activity android:name=\"org.godotengine.openxr.vendors.GodotPanelApp\" "
+				"android:process=\":GodotPanelApp\" "
+				"android:label=\"@string/godot_project_name_string\" "
+				"android:theme=\"@style/GodotAppSplashTheme\" "
```
m4gr3d, line 506: This presumes that this plugin will run with the Godot Android template which is not guaranteed (e.g: developers can integrate this plugin's aar binary into their own project with a different structure), in which case the `GodotAppSlashTheme` will be missing.

To counter that, we should replicate the themes in https://github.com/godotengine/godot/blob/master/platform/android/java/app/res/values/themes.xml in this project (with different names to avoid collision in the common case), and refer to the new themes instead.
Note that we don't need to duplicate the icons specified in the themes as those are provided by the Godot Android library which is a required dependency for this plugin.

plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -475,6 +497,34 @@ String MetaEditorExportPlugin::_get_android_manifest_application_element_content
 		contents += "        <meta-data android:name=\"com.oculus.ossplash.background\" android:value=\"passthrough-contextual\" />\n";
 	}
 
+	HybridType hybrid_type = _get_hybrid_app_option_value();
+	if (hybrid_type != HYBRID_TYPE_DISABLED) {
+		contents +=
+				"        <activity android:name=\"org.godotengine.openxr.vendors.GodotPanelApp\" "
+				"android:process=\":GodotPanelApp\" "
+				"android:label=\"@string/godot_project_name_string\" "
```
m4gr3d, line 505: Let's omit the `android:label` attribute; this will make this activity default to the one defined by the app.

plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -475,6 +497,34 @@ String MetaEditorExportPlugin::_get_android_manifest_application_element_content
...
+				"android:theme=\"@style/GodotAppSplashTheme\" "
+				"android:launchMode=\"singleInstancePerTask\" "
+				"android:excludeFromRecents=\"false\" "
+				"android:exported=\"true\" "
+				"android:screenOrientation=\"landscape\" "
+				"android:configChanges=\"orientation|keyboardHidden|screenSize|smallestScreenSize|density|keyboard|navigation|screenLayout|uiMode\" "
+				"android:resizeableActivity=\"true\" "
```
m4gr3d, line 512: We'll need to duplicate the logic in https://github.com/godotengine/godot/blob/master/platform/android/export/gradle_export_util.cpp#L266-L271 for the `excludeFromRecents`, `screenOrientation` and `resizeableActivity` attributes.

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRHybridAppInternal.kt
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		hybridLaunchData = activity.intent.getStringExtra(INTENT_EXTRA_DATA)
+
+		return null
+	}
+
+	@UsedByGodot
+	fun hybridAppSwitchTo(mode: Int, data: String = ""): Boolean {
+		if (hybridMode == mode) return false
```
m4gr3d, line 76: We should also check if this is an XR device before attempting the operation using https://github.com/godotengine/godot/blob/master/platform/android/java/lib/src/org/godotengine/godot/utils/DeviceUtils.kt#L57.

plugin/src/main/cpp/classes/openxr_hybrid_app.cpp
dsnopek: I went the other way, and made the `""` into `String()` so they are consistently `String()`

plugin/src/main/cpp/export/export_plugin.cpp
m4gr3d: > Yeah, that makes sense. But do we want to do that here? Or, save it for the big "move stuff to project settings" change as we discussed earlier?

Let's do that in this PR, especially since the other settings we'll be migrating will be under a different scope (e.g: `xr/openxr/hybrid_app` vs `xr/openxr/features`).

dsnopek:
Thanks! All review should be addressed in my latest push.

However, moving from an export option to a project setting didn't work when using the "Remote Debug" button (it wouldn't recognize changes to the setting). I just posted PR https://github.com/godotengine/godot/pull/101658 which will fix this.

(FYI, it worked as an export setting, because changing an export setting causes the export options overrides to be updated. However, updating a project setting doesn't. But Godot is updating the export options overrides before attempting an export in all other situations aside from "Remote Debug", so my PR changes that, rather than directly responding to project settings changes.)

m4gr3d (commented):
It's looking great! Just a few feedback to address and it should be good to go!

plugin/src/main/cpp/export/khronos_export_plugin.cpp
```diff
@@ -134,6 +134,8 @@ Dictionary KhronosEditorExportPlugin::_get_export_options_overrides(const Ref<go
 		overrides["khronos_xr_features/htc/lip_expression"] = MANIFEST_FALSE_VALUE;
 	}
 
+	overrides.merge(OpenXREditorExportPlugin::_get_export_options_overrides(p_platform));
```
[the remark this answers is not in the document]
dsnopek: > Is the intent to avoid making multiple calls to _supports_platform and _is_vendor_plugin_enabled?

Yeah, that was the intent.

If we put this up at the beginning, the parent method could bail based on the `_supports_platform()` and `_is_vendor_plugin_enabled()`, but it won't fully bail, because then the child method will do the same thing again.

I could re-arrange this so that it's like:

```
	if (!_supports_platform(p_platform)) {
		return Dictionary();
	}
	Dictionary overrides = OpenXREditorExportPlugin::_get_export_options_overrides(p_platform);
```

I don't know if that's really better, though?

plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -280,6 +280,11 @@ PackedStringArray MetaEditorExportPlugin::_get_export_features(const Ref<EditorE
 		features.append(EYE_GAZE_INTERACTION_FEATURE);
 	}
 
+	// Add a feature to indicate that this is a hybrid app.
+	if (_get_hybrid_app_setting_value() != HYBRID_TYPE_DISABLED) {
```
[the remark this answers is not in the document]
dsnopek: Wouldn't the `_is_vendor_plugin_enabled()` check also prevent this from being used on non-XR devices? A non-XR device wouldn't enable the Meta plugin.

samples/hybrid-app-sample/panel_switcher.gd
```diff
@@ -0,0 +1,28 @@
...
+
+var hybrid_app_mode := OpenXRHybridApp.HYBRID_MODE_NONE
+var data := {}
+
+func _ready() -> void:
+	hybrid_app_mode = OpenXRHybridApp.get_mode()
+	if hybrid_app_mode == OpenXRHybridApp.HYBRID_MODE_IMMERSIVE:
+		switch_button.text = "Switch To Panel"
+	elif hybrid_app_mode == OpenXRHybridApp.HYBRID_MODE_PANEL:
+		switch_button.text = "Switch To Immersive"
```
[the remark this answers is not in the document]
dsnopek: Good idea! Changed in my latest push

plugin/src/main/cpp/export/khronos_export_plugin.cpp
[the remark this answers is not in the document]
dsnopek: I ended up just swapping over the the alternative above in my latest push :-)

plugin/src/main/cpp/export/meta_export_plugin.cpp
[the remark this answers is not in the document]
m4gr3d: The vendor plugin can still be enabled while xr mode is set to `Regular`. I just tested this and while it does show a warning in the export preset, it doesn't prevent export and as such the non-XR build I made ended up with XR specific manifest additions.

So the check here would be an additional safeguard in case a developer inadvertently enables a XR plugin in a non-XR export preset.

samples/hybrid-app-sample/export_presets.cfg
```diff
@@ -0,0 +1,248 @@
...
+meta_xr_features/use_overlay_keyboard=false
+meta_xr_features/use_experimental_features=false
+meta_xr_features/boundary_mode=0
+khronos_xr_features/htc/hand_tracking=0
+khronos_xr_features/htc/tracker=0
+khronos_xr_features/htc/eye_tracking=0
+khronos_xr_features/htc/lip_expression=0
+khronos_xr_features/vendors=0
+meta_xr_features/use_anchor_sharing=false
+meta_xr_features/hybrid_app=1
```
m4gr3d, line 243: Leftover from the previous version.

plugin/src/main/cpp/export/meta_export_plugin.cpp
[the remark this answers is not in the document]
dsnopek: Alright! I added the check in my latest push

m4gr3d (approved):
Looks great!

PR #246, opened 2025-01-15
[DRAFT] Support making hybrid apps for Pico headsets

dsnopek:
This builds on PR https://github.com/GodotVR/godot_openxr_vendors/pull/245 to attempt to add hybrid app support for Pico headsets as well.

It _almost_ works!

Switching from immersive to panel works fine. Switching from immersive to panel works-ish, but the panel is minimized, and you have to click it on the toolbar again for it to appear, but it will appear in a panel.

Marking as DRAFT until we can get it working a little better.


samples/hybrid-app-sample/main.gd
```diff
@@ -59,9 +59,10 @@ func _process(_delta: float) -> void:
 		panel_switcher_layer.update_pointer(t.origin, -t.basis.z, pointer_pressed)
 
 func _on_right_controller_pointer_button_pressed(p_name: String) -> void:
-	if p_name == 'trigger_click':
+	print("button: ", p_name)
+	if p_name == 'trigger_click' or p_name == 'select_button':
```
dsnopek, line 63: I'm not sure what's going on with this. I took a quick peek at the action map, and it seemed like I should be getting "trigger_click" for Pico 4 controllers, but instead I'm getting "select_button". Needs to be investigated more.

plugin/src/main/cpp/export/pico_export_plugin.cpp
```diff
@@ -200,5 +222,34 @@ String PicoEditorExportPlugin::_get_android_manifest_application_element_content
...
+				"android:configChanges=\"orientation|keyboardHidden|screenSize|smallestScreenSize|density|keyboard|navigation|screenLayout|uiMode\" "
+				"android:resizeableActivity=\"true\" "
+				"tools:ignore=\"UnusedAttribute\" >\n"
+				"          <intent-filter>\n"
+				"            <action android:name=\"android.intent.action.MAIN\" />\n"
+				"            <category android:name=\"android.intent.category.DEFAULT\" />\n";
+
+		if (hybrid_type == HYBRID_TYPE_START_AS_PANEL) {
+			contents += "            <category android:name=\"android.intent.category.LAUNCHER\" />\n";
+		}
```
m4gr3d, line 247: We should try to move this common logic into the base class, and provide overrides to add the platform specific bits.

dsnopek: This one is a bit tricky to centralize, because different platform could have special things that need to put anywhere in this XML.

Between Meta and Pico, though, we could probably merge them fairly easily - the only difference is that Meta needs an additional `<category>`, and Pico needs an additional `<meta-data>` (which doesn't need to be included with this block, it can be put before or after so that's fine). So, if we had a function that could generate this with optional stuff to put in `<intent-filter>` then it could be shared in this case.

m4gr3d: I'd expect the `<activity android:name="org.godotengine.openxr.vendors.GodotPanelApp" ...>` tag to remain the same between the different platforms, so we could just add a helper function to generate that tag and return it for insertion here.
Since we need to add logic to customize the activity attributes based on export options, we can consolidate that logic in that helper method.

dsnopek: Ok! I've done that in my latest push

dsnopek:
Rebased now that PR https://github.com/GodotVR/godot_openxr_vendors/pull/245 has been merged

The diff is really quite small now!

plugin/src/main/cpp/export/pico_export_plugin.cpp
```diff
@@ -200,5 +205,25 @@ String PicoEditorExportPlugin::_get_android_manifest_application_element_content
 		contents += "        <meta-data tools:node=\"replace\" android:name=\"Hand_Tracking_HighFrequency\" android:value=\"1\" />\n";
 	}
 
+	HybridType hybrid_type = _get_hybrid_app_setting_value();
+	if (hybrid_type != HYBRID_TYPE_DISABLED) {
+		contents += "        <meta-data tools:node=\"replace\" android:name=\"pvr.app.type\" android:value=\"2d\" />\n";
```
m4gr3d, line 210: I'm curious about this meta-data.. should its value be `vr` when we set the immersive activity as the default launcher?

samples/hybrid-app-sample/export_presets.cfg
```diff
@@ -245,3 +245,252 @@ pico_xr_features/face_tracking=0
 pico_xr_features/hand_tracking=0
 xr_features/enable_magicleap_plugin=false
 magicleap_xr_features/hand_tracking=0
+pico_xr_features/hybrid_app=0
```
m4gr3d, line 248: Leftover from previous logic

samples/hybrid-app-sample/main.gd
m4gr3d: Could this be related to the issue mentioned in RC, where Pico defaults to a compat mode?

m4gr3d (approved):
Looks good!

plugin/src/main/cpp/export/pico_export_plugin.cpp
dsnopek: I think it might just control where the app is sorted in the library in headset. I need to experiment with that a bit.

But if that is correct, then, yeah, perhaps we should only set it when we start as panel, with the idea that the app is primarily a panel app.

samples/hybrid-app-sample/main.gd
dsnopek: Yep, I think it's the same thing, but I haven't had a chance to investigate further yet.

But we'll definitely want to have the action map working correctly before taking this out of draft.

dsnopek: It looks like there is a new controller interaction profile for Pico 4 Ultra (`/interaction_profiles/bytedance/pico4s_controller` - note the `s` in `pico4s`) and that is somehow related to it falling back on "Simple controller", because if I patch that into Godot, I'm able to get controller input normally. I'll make a Godot PR later, but I haven't yet managed to track down documentation on that interaction profile, so it'll have to be a draft until we can confirm all the paths

dsnopek: See https://github.com/godotengine/godot/pull/103234

PR #252, opened 2025-01-23
Fix passthrough sample color map display bug

devloglogan:
In the passthrough sample, when looking at the color map passthrough mode, the used color gradient is supposed to be displayed on a mesh in front of the user. This wasn't working properly because `albedo_texture` requires a `GradientTexture2D`, not just a `Gradient` resource. This PR fixes that!


m4gr3d (approved):

dsnopek (approved):

dsnopek:
Cherry-picked for 3.x in PR https://github.com/GodotVR/godot_openxr_vendors/pull/253

PR #265, opened 2025-02-07
Clean scons when doing a Gradle clean

dsnopek:
This makes `./gradlew clean` also do `scons -c` for all our platforms and architectures

There's probably some clever way to avoid duplicating so much between `buildSconsArtifacts` and `cleanScons`, but I went for the quick and easy approach. I suppose I could make a function that can create all the tasks with an argument to switch between build or clean?

Marking as DRAFT so I can mess around with that later

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ task clean(type: Delete) {
     dependsOn 'cleanAddon'
     dependsOn 'cleanBuildFromSamples'
     dependsOn 'cleanZippedSamples'
+    dependsOn 'cleanScons'
 
     dependsOn ':plugin:clean'
 }
@@ -53,11 +54,10 @@ task buildPlugin {
     finalizedBy "copyBuildToSamples"
 }
 
-/**
- * Build the scons artifacts for the project
+/*
+ * Find scons executable path
  */
-task buildSconsArtifacts {
-    // Find scons' executable path
+def getSconsExecutableFile() {
     File sconsExecutableFile = null
     def sconsName = "scons"
     def sconsExts = (org.gradle.internal.os.OperatingSystem.current().isWindows()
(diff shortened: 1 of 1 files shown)


m4gr3d:
> There's probably some clever way to avoid duplicating so much between `buildSconsArtifacts` and `cleanScons`, but I went for the quick and easy approach. I suppose I could make a function that can create all the tasks with an argument to switch between build or clean?

Yes that sounds like a good idea; add / remove the `-c` and update the naming from `build` / `clean` based on the flag

dsnopek:
Updated in my latest push to eliminate the code duplication as discussed above

build.gradle
```diff
@@ -81,54 +81,112 @@ task buildSconsArtifacts {
         }
     }
 
+    return sconsExecutableFile
+}
+
+def createSconsTasks(File sconsExecutableFile, boolean clean) {
+    def defaultArgs = [
+        "--directory=.",
+        "platform=android",
```
m4gr3d, line 90: Based on the logic below, seems like `platform=android` should be removed from the `defaultArgs`.

build.gradle
```diff
@@ -81,54 +81,112 @@ task buildSconsArtifacts {
...
+        // Android.
+        dependsOn 'cleanGodotOpenXRVendorsAndroidArm64Debug'
+        dependsOn 'cleanGodotOpenXRVendorsAndroidArm64Release'
+        dependsOn 'cleanGodotOpenXRVendorsAndroidX86_64Debug'
+        dependsOn 'cleanGodotOpenXRVendorsAndroidX86_64Release'
+
+        // Desktop.
+        dependsOn 'cleanGodotOpenXRVendorsDesktopDebug'
+        dependsOn 'cleanGodotOpenXRVendorsDesktopRelease'
+    }
```
m4gr3d, line 187: Can the rest of the logic be moved within `createSconsTasks` as well? Seems like everything left out is identical in logic as well when taking into account the `build` / `clean` prefix switch.

dsnopek: Not with my level of Gradle knowledge :-)

We need to have the exception stuff in a `doFirst { ... }` so it only happens when the task actually runs, and you can't put a `doFirst` in a function, it's just a task thing. Similar for the `dependsOn` - I don't know how to do that in a function rather than a task.

build.gradle
dsnopek: Ah, thanks! Fixed in my latest pushed

m4gr3d (approved):

PR #264, opened 2025-02-07
Fix hybrid app support for recent Godot Android library changes

dsnopek:
We're currently [failing CI due](https://github.com/GodotVR/godot_openxr_vendors/pull/259#issuecomment-2644064620) to recent changes to Godot Android library, that affect our hybrid app support.

This PR is attempting to fix that!

This now compiles, however, I still have to test it, so keeping as a DRAFT for now

PR #266, opened 2025-02-11
Update for Android library changes in Godot 4.4-beta4

dsnopek:
These are the changes necessary to make hybrid apps compile and work with the (not-yet-committed) Godot PR https://github.com/godotengine/godot/pull/102669.

It's built on top of https://github.com/GodotVR/godot_openxr_vendors/pull/264, with the assumption that we'll merge that soon so our CI will work and we can resume other stuff, and that this PR will be merged after 4.4-beta4 releases (with the changes from https://github.com/godotengine/godot/pull/102669)

### DIFF
--- a/plugin/src/main/java/org/godotengine/openxr/vendors/GodotPanelApp.kt
+++ b/plugin/src/main/java/org/godotengine/openxr/vendors/GodotPanelApp.kt
@@ -31,7 +31,6 @@ package org.godotengine.openxr.vendors
 
 import android.os.Bundle
 import android.util.Log
-import androidx.annotation.CallSuper
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import org.godotengine.godot.GodotActivity
 import org.godotengine.godot.BuildConfig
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/org/godotengine/openxr/vendors/GodotPanelApp.kt
```diff
@@ -61,10 +60,9 @@ class GodotPanelApp : GodotActivity() {
 		super.onCreate(savedInstanceState)
 	}
 
-	@CallSuper
-	protected override fun updateCommandLineParams(args: Array<String>) {
-		// Force XR to be turned off and discard other params.
-		super.updateCommandLineParams(arrayOf("--xr_mode_regular", "--xr-mode", "off"))
+	override fun getCommandLine(): MutableList<String> {
+		// Force XR to be turned off.
+		return mutableListOf("--xr_mode_regular", "--xr-mode", "off")
```
m4gr3d, line 65: We'd want to append to the parent command line arguments to ensure we don't inadvertently break some features. If there is conflicts with the arguments we're adding, we should remove the conflicting arguments and keep the rest.

dsnopek: Updated in my latest push

dsnopek:
This is working for me in the hybrid app sample with the (now released) Godot 4.4-beta4! Taking out of DRAFT status

dsnopek:
Looks like we're going to be failing CI until this PR is merged - see this recent test build: https://github.com/GodotVR/godot_openxr_vendors/actions/runs/13393965119/job/37408499157

m4gr3d (approved):

PR #267, opened 2025-02-11
Add support for `XR_FB_composition_layer_image_layout`

dsnopek:
Selection_268

I'm not happy about exposing this extension to developers (it's something they shouldn't have to worry about), but I think it's the best way at the moment to allow developers to work around issue https://github.com/godotengine/godot/issues/102639

Ultimately, I think the above issue is the result of `XR_KHR_android_surface_swapchain` being under-specified and different vendors all doing different things in their implementations. Hopefully, we can get that fixed in the OpenXR spec and eventually not need this anymore.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
@@ -1,0 +1,126 @@
+/**************************************************************************/
+/*  openxr_fb_composition_layer_image_layout_extension_wrapper.cpp        */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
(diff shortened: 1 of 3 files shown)


plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
```diff
@@ -0,0 +1,127 @@
...
+		return reinterpret_cast<uint64_t>(p_next_pointer);
+	}
+
+	const XrCompositionLayerBaseHeader *layer = reinterpret_cast<const XrCompositionLayerBaseHeader *>(p_layer);
+
+	if (!layer_structs.has(layer)) {
+		layer_structs[layer] = {
+			XR_TYPE_COMPOSITION_LAYER_IMAGE_LAYOUT_FB, // type
+			nullptr, // next
+		};
```
devloglogan, line 87: I know this is being set at the bottom of this function, but I'd had it in my head for some reason that when initializing like this, we should still explicitly write out all fields when possible.

```suggestion
		layer_structs[layer] = {
			XR_TYPE_COMPOSITION_LAYER_IMAGE_LAYOUT_FB, // type
			nullptr, // next
			0 // flags
		};
```

plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
```diff
@@ -0,0 +1,127 @@
...
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.h"
+
+#include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
```
devloglogan, line 33: Doesn't look like this is actually used so might as well remove it.

devloglogan (approved):
Code LGTM, just some little nits.

plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
dsnopek: Removed!

plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
devloglogan: Fair enough!

BastiaanOlij (approved):
LGTM

PR #272, opened 2025-02-20
Use project settings avoid enabling unneeded OpenXR extensions

dsnopek:
This is an alternative to PR https://github.com/GodotVR/godot_openxr_vendors/pull/234, but with basically the exact same goal: "try not to enable OpenXR extensions that are unneeded"

The main difference in approach between this PR and the other, is that this one always registers the singletons with `Engine::get_singleton()->register_singleton()` (which exposes them to GDScript), but only does `OpenXRExtensionWrapperExtension::register_extension_wrapper()` if their project setting is enabled (which registers them with `OpenXRAPI` and will lead to their OpenXR extensions being requested).

It's working in my testing! If you enable "Verbose stdout", then Godot will output the OpenXR extensions that have been enabled, and I can see that they are only enabled if the project setting is set. I also tested both the main demo, and the Meta scene sample, and was able to confirm that they were missing functionality when the project settings were disabled, but worked perfect when enabled.

(Also, notably, they didn't crash but still sort of worked when the project settings were disabled, which I thought was kind of cool!)

But I'm marking this as a DRAFT for now, because there are a ton of questions to discuss (which we can talk about at the XR team meeting tonight), like:

- Should the settings be named with the vendor in the name?
- Or, should the ones with the same functionality but multiple vendor extensions use the same name? (like, we have "fb_passthrough" and "htc_passthrough" - should it just be "passthrough" that enables both?)
- What naming scheme or hierarchy should we use?
- Are there some that shouldn't have project settings? (for example, I personally think the composition layer ones shouldn't have project settings - it would create too much confusion, IMO)
- Which export settings do we eliminate in favor of project settings? Currently, I have the scene/spatial anchor settings completely replacing the equivalent export settings. Do we do this for all the ones that pair with extensions? Or, only those that are basically Android only? Or all settings, even ones that aren't related to OpenXR extensions?
  - Note: Some of the permissions have "optional" vs "required" - but that concept only exists in the `AndroidManifest.xml`, not the OpenXR extension. So, for extensions that work on both Android and PCVR, the "optional" vs "required" part is meaningless on PCVR

Lots to hash out :-)


dsnopek:
I've updated this PR for _most_ of the things we discussed at our last meeting about it, with the following exceptions due to technical issues:

- **Composition layers showing warnings when the extensions aren't enabled.** The `OpenXRCompositionLayer` node is from Godot itself, and it would need to implement `get_configuration_warnings()` - we can't do it from this GDExtension, without adding a new "hook" to do it, and since we are targeting Godot 4.4, that's not something we can do right now. So, in the meantime, the composition layer extensions are enabled by default to prevent confusion. We can add a hook for ourselves in Godot 4.5 and then make this nicer in the future
- **For None/Optional/Required settings, showing only "None" or hiding the option, when the extension isn't enabled.** I tried to get both ideas working, but unfortunately, changing project settings doesn't trigger export options to get rebuilt, and so they'll show the old options until you switch presets or change an export option that affects visibility. This would create confusion. Again, this is something we can fix in Godot 4.5 (or maybe even 4.4.2?), but can't do right now

m4gr3d:
> * **For None/Optional/Required settings, showing only "None" or hiding the option, when the extension isn't enabled.** I tried to get both ideas working, but unfortunately, changing project settings doesn't trigger export options to get rebuilt, and so they'll show the old options until you switch presets or change an export option that affects visibility. This would create confusion. Again, this is something we can fix in Godot 4.5 (or maybe even 4.4.2?), but can't do right now

@dsnopek Could [`EditorExportPlugin#_should_update_export_options(...)`](https://docs.godotengine.org/en/stable/classes/class_editorexportplugin.html#class-editorexportplugin-private-method-should-update-export-options) be used to address this?

We should be able to store the initial value of the project settings when the export preset is first open, then compare against the current value within that method.

dsnopek:
> @dsnopek Could [`EditorExportPlugin#_should_update_export_options(...)`](https://docs.godotengine.org/en/stable/classes/class_editorexportplugin.html#class-editorexportplugin-private-method-should-update-export-options) be used to address this?

Thanks for the suggestion! I totally forgot that we had that on `EditorExportPlugin`, and I just tried it, and it works. :-) In my next update, I'll switch to using that

dsnopek:
This PR is updated to:

- Hide export options if the corresponding project setting for the required extension isn't enabled
- Remove the "None" from the None/Optional/Required options (defaulting to "Optional") if it depends on an extension being enabled in project settings (there is a Pico export option where I kept "None" because it wasn't associated with an extension)
- Remove the export option entirely if it was a Yes/No and we can tell if it's enabled or not entirely from the project setting
- Properly clean up extensions that were registered as singletons, but not registered with `OpenXRAPI` - this was the source of the CI failures when running with XrSim on Windows

This should be ready for review again!

BastiaanOlij:
XR meeting: now that the changes have been done as discussed, this should be good to go. Needs a final review but we should move forward with this as soon as possible.

plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
```diff
@@ -108,13 +110,31 @@ float OpenXRFbHandTrackingMesh::get_scale_override() const {
 	return OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->get_scale_override(hand);
 }
 
+PackedStringArray OpenXRFbHandTrackingMesh::_get_configuration_warnings() const {
+	PackedStringArray warnings;
+	if (!ProjectSettings::get_singleton()->get_setting_with_override("xr/openxr/extensions/meta/hand_tracking_mesh")) {
+		warnings.push_back("The render model extension isn't enabled in project settings. Please enable `xr/openxr/extensions/meta/hand_tracking_mesh` to use this node.");
```
m4gr3d, line 116:
```suggestion
		warnings.push_back("The hand tracking mesh extension isn't enabled in project settings. Please enable `xr/openxr/extensions/meta/hand_tracking_mesh` to use this node.");
```

plugin/src/main/cpp/include/export/meta_export_plugin.h
```diff
@@ -109,6 +100,12 @@ class MetaEditorExportPlugin : public OpenXREditorExportPlugin {
 	Dictionary _support_quest_2_option;
 	Dictionary _support_quest_3_option;
 	Dictionary _support_quest_pro_option;
+
+	mutable bool should_update_options = false;
```
m4gr3d, line 104: Should be `_should_update_options` for consistency.

plugin/src/main/cpp/include/export/pico_export_plugin.h
```diff
@@ -58,19 +57,26 @@ class PicoEditorExportPlugin : public OpenXREditorExportPlugin {
...
 
-	Dictionary _eye_tracking_option;
+private:
 	Dictionary _face_tracking_option;
 	Dictionary _hand_tracking_option;
 
-private:
 	bool _is_eye_tracking_enabled() const;
+
+	mutable bool should_update_options = false;
```
m4gr3d, line 75: Should be `_should_update_options` for consistency.

plugin/src/main/cpp/register_types.cpp
```diff
@@ -86,109 +86,165 @@
...
+	return (bool)ProjectSettings::get_singleton()->get_setting_with_override(p_name);
+}
+
+static void _register_extension_with_openxr(OpenXRExtensionWrapperExtension *p_extension) {
+	p_extension->register_extension_wrapper();
+	extensions_registered.push_back(p_extension);
+}
+
+static void _register_extension_as_singleton(const StringName &p_name, OpenXRExtensionWrapperExtension *p_extension) {
+	Engine::get_singleton()->register_singleton(p_name, p_extension);
```
m4gr3d, line 116: Could you use `get_class_static()` (or related `GDCLASS` functions) to get the class name of the extension?

plugin/src/main/cpp/register_types.cpp
```diff
@@ -245,6 +298,31 @@ void terminate_plugin_module(ModuleInitializationLevel p_level) {
...
+			Engine::get_singleton()->unregister_singleton("OpenXRFbSpatialEntityStorageExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRFbSpatialEntityQueryExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRFbSpatialEntityContainerExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRFbSceneExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRFbHandTrackingAimExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRFbHandTrackingCapsulesExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRFbCompositionLayerDepthTestExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRFbCompositionLayerSettingsExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRHtcFacialTrackingExtensionWrapper");
+			Engine::get_singleton()->unregister_singleton("OpenXRHtcPassthroughExtensionWrapper");
```
m4gr3d, line 314: Couldn't this be done in the `for` loop as well?

dsnopek:
@m4gr3d Thanks for the review! I've included all your suggestions in my latest push :-)

m4gr3d (approved):
Looks great!

plugin/src/main/cpp/register_types.cpp
```diff
@@ -86,109 +86,169 @@
...
-			Engine::get_singleton()->register_singleton("OpenXRFbSpatialEntityQueryExtensionWrapper", OpenXRFbSpatialEntityQueryExtensionWrapper::get_singleton());
-			Engine::get_singleton()->register_singleton("OpenXRFbSpatialEntityContainerExtensionWrapper", OpenXRFbSpatialEntityContainerExtensionWrapper::get_singleton());
-			Engine::get_singleton()->register_singleton("OpenXRFbSceneExtensionWrapper", OpenXRFbSceneExtensionWrapper::get_singleton());
-			Engine::get_singleton()->register_singleton("OpenXRFbHandTrackingAimExtensionWrapper", OpenXRFbHandTrackingAimExtensionWrapper::get_singleton());
-			Engine::get_singleton()->register_singleton("OpenXRFbHandTrackingCapsulesExtensionWrapper", OpenXRFbHandTrackingCapsulesExtensionWrapper::get_singleton());
-			Engine::get_singleton()->register_singleton("OpenXRFbCompositionLayerDepthTestExtensionWrapper", OpenXRFbCompositionLayerSettingsExtensionWrapper::get_singleton());
-			Engine::get_singleton()->register_singleton("OpenXRFbCompositionLayerSettingsExtensionWrapper", OpenXRFbCompositionLayerSettingsExtensionWrapper::get_singleton());
```
m4gr3d, line 189: @dsnopek I just noticed these two were not carried over below; was the omission intentional?

m4gr3d: Thanks for the clarification! I'll close the other PR.

PR #273, opened 2025-02-20
Yet another implementation of XR_META_environment_depth (using global shader uniforms)

dsnopek:
Fixes https://github.com/GodotVR/godot_openxr_vendors/issues/133

This is an alternative to the existing PRs to implement XR_META_environment_depth (ex https://github.com/GodotVR/godot_openxr_vendors/pull/230 and  https://github.com/GodotVR/godot_openxr_vendors/pull/232), with the notable advantage that I've managed to get a test project successfully doing occlusion (the others don't quite make it all the way there):

https://github.com/user-attachments/assets/cac6cd09-efbd-4b3b-b312-da7fc8c0601c

(Please ignore my very messy desk :-))

The implementation works by taking the data from Meta's depth API, then stashing it in global shader uniforms.

This allows developers to:

1. Use the data in their own shaders, to do occlusion with the environment, or any other purpose they can think of. This PR includes an example of such a shader (which is being used for the cube in the video).
2. Or, use the `OpenXRMetaEnvironmentDepth` node which will fill the depth buffer using a fullscreen triangle, and then all rendering can be occluded by the environment

This PR needs Godot PR https://github.com/godotengine/godot/pull/106880 to work, which was merged for Godot 4.5. However, it doesn't depend on any GDExtension changes from Godot 4.5, only an internal re-ordering of some stuff, so this _can_ be compiled into a build that will work with Godot 4.4.

I've added runtime checks for the current Godot version, in order to only actually enable this feature with Godot 4.5 or later, so we don't have to hold off merging it until later.

**UPDATE (2025-06-03):** I've re-worked this PR from PoC to an actually shippable form!


dsnopek:
The latest version of this is working with Vulkan now too!

dsnopek:
This in now in a shippable form (rather than a PoC) and ready for review!

It builds on top of https://github.com/godotengine/godot/pull/106880 which was just merged for Godot 4.5. However, I don't think we need to wait until Godot 4.5 to merged this: it doesn't depend on any GDExtension changes from Godot 4.5, only an internal re-ordering of some stuff, so this can be compiled into a build that will work with Godot 4.4. I've simply added a runtime check, that will disable this feature (including its settings in the editor) when not using Godot 4.5 or later.

The latest version also adds a `OpenXRMetaEnvironmentDepth` node that will populate Godot depth buffer with reprojected data from `XR_META_environment_depth`, which will (by default) allow everything rendered by Godot to be occluded by real world objects.

In the Meta scene sample, it will enable this extension, which will use the node by default. However, by pressing the B button, you can toggle between the node, and a material that's applied only to the test cube, in order to demonstrate both methods of using this data.

Please let me know what you think!

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,746 @@
...
+static const char *META_ENVIRONMENT_DEPTH_INV_PROJECTION_VIEW_LEFT_NAME = "META_ENVIRONMENT_DEPTH_INV_PROJECTION_VIEW_LEFT";
+static const char *META_ENVIRONMENT_DEPTH_INV_PROJECTION_VIEW_RIGHT_NAME = "META_ENVIRONMENT_DEPTH_INV_PROJECTION_VIEW_RIGHT";
+static const char *META_ENVIRONMENT_DEPTH_FROM_CAMERA_PROJECTION_LEFT_NAME = "META_ENVIRONMENT_DEPTH_FROM_CAMERA_PROJECTION_LEFT";
+static const char *META_ENVIRONMENT_DEPTH_FROM_CAMERA_PROJECTION_RIGHT_NAME = "META_ENVIRONMENT_DEPTH_FROM_CAMERA_PROJECTION_RIGHT";
+static const char *META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_LEFT_NAME = "META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_LEFT";
+static const char *META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_RIGHT_NAME = "META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_RIGHT";
+
+static const char *META_ENVIRONMENT_DEPTH_REPROJECTION_SHADER_CODE =
+		"shader_type spatial;\n"
+		//"render_mode unshaded, shadows_disabled, cull_disabled, depth_draw_always;\n"
```
devloglogan: Is there a point in keeping this commented out line?

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,746 @@
...
+		"\tclip_back.z += PROJECTION_MATRIX[2][2] * z_adjustment;\n"
+		"#endif // USE_DEPTH_OFFSET_SCALE\n"
+		"\tclip_back = depth_to_camera_proj * clip_back;\n"
+		"\thighp float camera_ndc_z = clip_back.z / clip_back.w;\n"
+		"#if CURRENT_RENDERER == RENDERER_COMPATIBILITY\n"
+		"\thighp float camera_depth = camera_ndc_z * 0.5 + 0.5;\n"
+		"#else\n"
+		"\thighp float camera_depth = camera_ndc_z;\n"
+		"#endif\n"
+		//"\tALBEDO = vec3(1.0 - camera_depth);\n"
```
devloglogan: Another commented line to possibly remove.

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,746 @@
...
+#define XR_USE_GRAPHICS_API_OPENGL_ES
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+#include <jni.h>
+
+#define XR_USE_GRAPHICS_API_VULKAN
+#include <vulkan/vulkan.h>
+
+#include <openxr/openxr_platform.h>
+#endif
```
devloglogan:
```suggestion
#endif // ANDROID_ENABLED
```
Doesn't hurt to add 🙂

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,746 @@
...
+		}
+	}
+}
+
+void OpenXRMetaEnvironmentDepthExtensionWrapper::_on_instance_destroyed() {
+	cleanup();
+}
+
+void OpenXRMetaEnvironmentDepthExtensionWrapper::_on_session_created(uint64_t p_session) {
+}
```
devloglogan: Is there a point in keeping this if it's empty?

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,746 @@
...
+	if (depth_provider == XR_NULL_HANDLE) {
+		if (!create_depth_provider()) {
+			return;
+		}
+	}
+
+	XrEnvironmentDepthHandRemovalSetInfoMETA info = {
+		XR_TYPE_ENVIRONMENT_DEPTH_HAND_REMOVAL_SET_INFO_META, // type
+		nullptr, // next
+		p_enable,
```
devloglogan:
```suggestion
		p_enable, // enabled
```

doc_classes/OpenXRMetaEnvironmentDepth.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRMetaEnvironmentDepth" inherits="VisualInstance3D" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Populates Godot's depth buffer with reprojected data from the [code]XR_META_environment_depth[/code] extension.
+	</brief_description>
+	<description>
+		Populates Godot's depth buffer with reprojected data from the [code]XR_META_environment_depth[/code] extension, which will (by default) allow everything rendered by Godot to be occluded by real world objects.
+		If you want certain objects not be occluded, then set the [code]render_priority[/code] on their materials to a value lower than [member render_priority], so it's rendered before this node fills the depth buffer. But if this isn't flexible enough for your needs, you can write your own shaders that directly use the environment depth data.
```
devloglogan:
```suggestion
		If you want certain objects to not be occluded, set the [code]render_priority[/code] on their materials to a value lower than [member render_priority] so they're rendered before this node fills the depth buffer. If that isn't flexible enough for your needs, you can write your own shaders that directly use the environment depth data.
```
Besides the missing word, this is just wording I like better. 🙂

doc_classes/OpenXRMetaEnvironmentDepthExtensionWrapper.xml
```diff
@@ -0,0 +1,69 @@
...
+		<method name="is_environment_depth_supported">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if environment depth is supported; otherwise, [code]false[/code].
+			</description>
+		</method>
+		<method name="is_hand_removal_supported">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if hand removal is enabled; otherwise, [code]false[/code].
```
devloglogan:
```suggestion
				Returns [code]true[/code] if hand removal is supported; otherwise, [code]false[/code].
```

devloglogan (commented):
Still need to test but did a look through the PR. It's looking really good! Only found some very minor things.

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
dsnopek: I've been using this (and the other commented out line) a ton in debugging issues in the reprojection. It allows you to see the reprojected data in color, rather than just filling the depth values.

But, yeah, we probably don't need to have this in the merged PR! I'll remove it

dsnopek: I removed that comment, and added another one referring to https://github.com/godotengine/godot/pull/107238

Once that one gets merged, I'll update the shader per the notes, and pull that comment out too

samples/meta-scene-sample/main.tscn
```diff
@@ -59,12 +66,19 @@ pose = &"grip"
...
+unique_name_in_owner = true
 
 [node name="Label3D" type="Label3D" parent="XROrigin3D/RightHand"]
 transform = Transform3D(1, 0, 0, 0, -4.37114e-08, 1, 0, -1, -4.37114e-08, 0, -0.124379, -0.185838)
 pixel_size = 0.001
-text = "[A] Place Floating Anchor
-[Trigger] Place/Remove Anchors"
+text = "[Trigger] Place/Remove Anchors
+[A] Place Floating Anchor
+[B] Toggle Global Depth"
```
devloglogan: I'd like these swapped because the B button is above the A button on the controller. It's written that way on the left controller label.

devloglogan (approved):
Seems to work great from my testing! Enabling/disabling/reprojection offset/render priority worked as I expected.

samples/meta-scene-sample/main.tscn
dsnopek: Done!

plugin/src/main/cpp/classes/openxr_meta_environment_depth.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+	OpenXRMetaEnvironmentDepthExtensionWrapper *env_depth_ext = OpenXRMetaEnvironmentDepthExtensionWrapper::get_singleton();
+	if (env_depth_ext) {
+		env_depth_ext->connect("openxr_meta_environment_depth_started", callable_mp(this, &OpenXRMetaEnvironmentDepth::_on_environment_depth_started));
+		env_depth_ext->connect("openxr_meta_environment_depth_stopped", callable_mp(this, &OpenXRMetaEnvironmentDepth::_on_environment_depth_stopped));
+	}
+
+	RenderingServer *rs = RenderingServer::get_singleton();
+	if (rs) {
+		rs->instance_geometry_set_cast_shadows_setting(get_instance(), RenderingServer::SHADOW_CASTING_SETTING_OFF);
```
m4gr3d, line 155: What does this do?

plugin/src/main/cpp/register_types.cpp
```diff
@@ -394,6 +415,7 @@ void add_plugin_project_settings() {
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/anchor_api", false);
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/anchor_sharing", false);
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/scene_api", false);
+	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/environment_depth", false);
```
m4gr3d: Should the setting also be gated by Godot 4.5?

plugin/src/main/cpp/classes/openxr_meta_environment_depth.cpp
dsnopek: So, we're using a fullscreen triangle to render the reprojected data into Godot's depth buffer. This makes it so that triangle won't cast shadows

plugin/src/main/cpp/register_types.cpp
dsnopek: Yes, it should! Thanks for catching that :-) Updated in my latest push

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,741 @@
...
+static const char *META_ENVIRONMENT_DEPTH_PROJECTION_VIEW_LEFT_NAME = "META_ENVIRONMENT_DEPTH_PROJECTION_VIEW_LEFT";
+static const char *META_ENVIRONMENT_DEPTH_PROJECTION_VIEW_RIGHT_NAME = "META_ENVIRONMENT_DEPTH_PROJECTION_VIEW_RIGHT";
+static const char *META_ENVIRONMENT_DEPTH_INV_PROJECTION_VIEW_LEFT_NAME = "META_ENVIRONMENT_DEPTH_INV_PROJECTION_VIEW_LEFT";
+static const char *META_ENVIRONMENT_DEPTH_INV_PROJECTION_VIEW_RIGHT_NAME = "META_ENVIRONMENT_DEPTH_INV_PROJECTION_VIEW_RIGHT";
+static const char *META_ENVIRONMENT_DEPTH_FROM_CAMERA_PROJECTION_LEFT_NAME = "META_ENVIRONMENT_DEPTH_FROM_CAMERA_PROJECTION_LEFT";
+static const char *META_ENVIRONMENT_DEPTH_FROM_CAMERA_PROJECTION_RIGHT_NAME = "META_ENVIRONMENT_DEPTH_FROM_CAMERA_PROJECTION_RIGHT";
+static const char *META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_LEFT_NAME = "META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_LEFT";
+static const char *META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_RIGHT_NAME = "META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_RIGHT";
+
+static const char *META_ENVIRONMENT_DEPTH_REPROJECTION_SHADER_CODE =
```
m4gr3d: Can we use a raw string literal so we can drop the newlines and tab characters.

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,741 @@
...
+
+void OpenXRMetaEnvironmentDepthExtensionWrapper::_on_instance_destroyed() {
+	cleanup();
+}
+
+void OpenXRMetaEnvironmentDepthExtensionWrapper::_on_session_destroyed() {
+	destroy_depth_provider();
+}
+
+static void xrMatrix4x4f_to_godot_projection(XrMatrix4x4f *m, Projection &p) {
```
m4gr3d: Can you move this to the `util` file.

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,741 @@
...
+				{
+						XR_TYPE_ENVIRONMENT_DEPTH_IMAGE_VIEW_META, // type
+						nullptr, // next
+				},
+		}
+	};
+
+	XrResult result = xrAcquireEnvironmentDepthImageMETA(depth_provider, &acquire_info, &depth_image);
+	if (XR_FAILED(result)) {
+		UtilityFunctions::printerr("Failed to acquire environment depth image: ", get_openxr_api()->get_error_string(result));
```
m4gr3d:
```suggestion
		UtilityFunctions::printerr("Failed to acquire environment depth image: ", openxr_api->get_error_string(result));
```

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,741 @@
...
+}
+
+RID OpenXRMetaEnvironmentDepthExtensionWrapper::get_reprojection_mesh() {
+	if (reprojection_mesh.is_null()) {
+		reprojection_shader.instantiate();
+		reprojection_material.instantiate();
+
+		update_reprojection_material(true);
+
+		reprojection_material->set_render_priority(-50);
```
m4gr3d: Shouldn't this refer to `reprojection_render_priority`?

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,741 @@
...
+		vertices[2] = Vector3(-1.0f, 3.0f, 1.0f);
+
+		Array arr;
+		arr.resize(RenderingServer::ARRAY_MAX);
+		arr[RenderingServer::ARRAY_VERTEX] = vertices;
+
+		reprojection_mesh.instantiate();
+		reprojection_mesh->add_surface_from_arrays(Mesh::PRIMITIVE_TRIANGLES, arr);
+		reprojection_mesh->surface_set_material(0, reprojection_material);
+		reprojection_mesh->set_custom_aabb(AABB(Vector3(-1000, -1000, -1000), Vector3(2000, 2000, 2000)));
```
m4gr3d, line 402: Is the custom aabb set to a bogus value?

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
dsnopek: Ah, yes, it should!

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
dsnopek: It's set to a large enough value that it can't ever be frustum culled :-)

I got that value in particular from [here](https://docs.godotengine.org/en/latest/classes/class_geometryinstance3d.html#class-geometryinstance3d-property-custom-aabb) - it says:

> To avoid frustum culling, set [custom_aabb](https://docs.godotengine.org/en/latest/classes/class_geometryinstance3d.html#class-geometryinstance3d-property-custom-aabb) to a very large AABB that covers your entire game world such as AABB(-10000, -10000, -10000, 20000, 20000, 20000).

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
dsnopek: I wrote it the way we generally do those in Godot itself, but, yes, it would be much nicer to work with as a raw literal string. Switched in my latest push!

plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp
dsnopek: Done!

BastiaanOlij:
Decision in XR meeting: Once rebased, we can merge this.

PR #275, opened 2025-02-27
Passthrough extensions should override real alpha blend mode, if enabled

dsnopek:
~~This (partially) depends on Godot PR https://github.com/godotengine/godot/pull/103338 and also shouldn't be merged until we merge one of the PRs that allow enabling/disabling OpenXR extension via project settings (like https://github.com/GodotVR/godot_openxr_vendors/pull/272 or https://github.com/GodotVR/godot_openxr_vendors/pull/234), which is why it's marked as a DRAFT.~~

This PR is to fix issues with headsets that support both a real alpha blend mode, and one of the passthrough extensions that can emulate it.

Because the passthrough extensions can provide additional functionality beyond just turning passthrough mode on and off (which is all the alpha blend mode can do), the idea is that if the developer has enabled a passthrough extension, it should be used rather than the real alpha blend mode.

That's what this PR does! Rather than the FB and HTC extensions only doing the emulation if there is no support for alpha blend mode (either from emulation or the real thing), it will do the emulation even if there is a real alpha blend mode (but not if some other extension has already decided to do emulation).

This works in my testing on Pico 4 Ultra

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
@@ -182,8 +182,10 @@ void OpenXRFbPassthroughExtensionWrapper::_on_instance_created(uint64_t p_instan
 void OpenXRFbPassthroughExtensionWrapper::_on_session_created(uint64_t p_session) {
 	XrSession session = (XrSession)p_session;
 	if (fb_passthrough_ext) {
-		// If it's already supported, then we don't want to emulate it.
-		if (get_openxr_api()->is_environment_blend_mode_alpha_supported() != OpenXRAPIExtension::OPENXR_ALPHA_BLEND_MODE_SUPPORT_NONE) {
+		// If another extension is already emulating alpha blend mode, then we don't attempt to.
+		if (get_openxr_api()->is_environment_blend_mode_alpha_supported() == OpenXRAPIExtension::OPENXR_ALPHA_BLEND_MODE_SUPPORT_EMULATING) {
+			// Act as if the extension is not enabled.
+			fb_passthrough_ext = false;
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
Not sure if this is correct David, if the extension is enabled, we trigger that we are emulating alpha blend mode. But now you're turning the extension off in this scenario.

This change should be that the extension is governed by a project setting.
If the user enables the FB passthrough extension, and it is supported, then we will be emulating alpha blend mode (which with the fix in Godot core should now work properly). 
If the user disables the FB passthrough extension, then we won't emulate the alpha blend mode, and thus can only do passthrough if the real alpha blend mode is supported.

So the only thing we should be adding here, and checking here, is a project settings for FB passthrough and one for HTC passthrough, and not register these extensions when not enabled.

dsnopek:
> Not sure if this is correct David, if the extension is enabled, we trigger that we are emulating alpha blend mode. But now you're turning the extension off in this scenario.

No, this specific code is about turning the extension off.

The old version would turn the extension off if there was real alpha blend available, which we don't want to do because we want to override that. The new version in this PR is only turning it off if another extension is doing emulation.

> So the only thing we should be adding here, and checking here, is a project settings for FB passthrough and one for HTC passthrough, and not register these extensions when not enabled.

I agree that we should add a project setting (but not in this PR), however, it shouldn't be checked in this spot. The project setting should control whether or not `register_extension_wrapper()` is called on the extension, and if that isn't called, we'll never even get to the code in this PR.

Anyway, this PR isn't addressing the project setting - the project setting should be done first in one of the other PRs.

This PR is just changing the incorrect check, where the passthrough extensions would disable themselves if there was a real alpha blend mode.

BastiaanOlij:
Gotcha! Makes sense!

dsnopek:
Now that PR https://github.com/GodotVR/godot_openxr_vendors/pull/272 is merged, I'm taking this one out of DRAFT

devloglogan (approved):
LGTM, tested on Pico 4 Ultra and, when enabled in project settings, it's now doing passthrough via `XR_FB_passthrough`. 😄

PR #278, opened 2025-03-03
Update `CHANGES.md` for 4.0.0-beta1

dsnopek:
We don't appear to list pre-releases in `CHANGES.md`, so this adds the changes under `4.0.0`. I figure we can just keep adding stuff under that heading until we do the stable release of `4.0.0`

This also adds the `3.1.2` changelog entries, which were never in `master`, because we did that on `3.x` after that one branched off of `master`

The entries are based on `git log --oneline 3.1.2-stable..`


m4gr3d (approved):

CHANGES.md
```diff
@@ -1,5 +1,21 @@
 # Change history for the Godot OpenXR loaders asset
 
+## 4.0.0
+
+- Support making hybrid apps for Meta headsets
+- Add support for `XR_FB_android_surface_swapchain_create`
+- Implement `XR_META_recommended_layer_resolution`
+- Remove CMake from the build process
+- Implement instance splash screen for Meta headsets
```
devloglogan, line 9:
```suggestion
- Implement instant splash screen for Meta headsets
```

dsnopek: Thanks!

PR #284, opened 2025-03-21
Fix standard manifest entries

rpavlik:
I have not tested this, but this corrects the manifest entries for the "Khronos" variant to match the entries added by the manifest in the loader AAR. This should increase compatibility of this build variant (as long as the activity is getting the right intent filter, which sadly can't be done this easily via merging manifest files)

### DIFF
--- a/plugin/src/khronos/AndroidManifest.xml
+++ b/plugin/src/khronos/AndroidManifest.xml
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <!-- Tell the system this app works in either 3dof or 6dof mode -->
+    <!-- Tell storefronts this app works in either 3dof or 6dof mode,
+         and do not filter out devices that did not ship with VR support. -->
     <uses-feature
         android:name="android.hardware.vr.headtracking"
-        android:required="true"
+        android:required="false"
         android:version="1"/>
 
     <!-- Permissions needed by OpenXR -->
--- a/plugin/src/khronos/AndroidManifest.xml
+++ b/plugin/src/khronos/AndroidManifest.xml
@@ -13,7 +13,10 @@
 
     <queries>
         <intent>
-            <action android:name="org.khronos.openxr.OpenXRRuntimeService"/>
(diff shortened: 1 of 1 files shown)


rpavlik (commented):
(alternately you could just use the dependency the normal way, since as I mentioned, the manifest to merge is in the aar)

plugin/src/khronos/AndroidManifest.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <!-- Tell the system this app works in either 3dof or 6dof mode -->
+    <!-- Tell storefronts this app works in either 3dof or 6dof mode,
+         and do not filter out devices that did not ship with VR support. -->
     <uses-feature
         android:name="android.hardware.vr.headtracking"
-        android:required="true"
+        android:required="false"
```
dsnopek, line 8: Should we make this configurable? Because we don't know if the developer intends for the store to filter out devices without VR support

rpavlik: So I am working on a writeup about this here, but basically, because as far as I know, `<use-feature>` only looks at the static configuration of the device, putting "true" here, in the best case scenario, limits things to only all-in-one devices and excludes phones using a plug-in viewer.  (Worst case, this might actually filter out everything except the Daydream compatible devices still around out there.)  I would rather leave this entry out entirely, since it appears to be ignored most of the time, than to include it as "required=true"

dsnopek: Ok, thanks!

Here's some Daydream-era docs about this tag:

https://developers.google.com/vr/develop/android/3dof-to-6dof

I don't know that folks really make 3DoF apps with Godot anymore, but I agree that not getting filtered out at all is probably the preferable outcome, since I could imagine a modern Godot app within a single APK that could work either flat (non-XR) or 6DoF depending on what the hardware is capable of

dsnopek (approved):
Thanks! Aligning with the `AndroidManifest.xml` in the official loader AAR makes sense

m4gr3d (approved):

PR #290, opened 2025-04-09
Add individual project settings for each composition layer extension

dsnopek:
In PR https://github.com/GodotVR/godot_openxr_vendors/pull/272, I put all the Meta composition layer extensions under a single project setting.

My thinking at the time was that they are so small (they just add optional settings to `OpenXRCompositionLayer` nodes) and there's a whole bunch of them, such that it would be too much for most users.

However, I'm starting to re-think that! If any of these extensions are deprecated in favor of a new extension, or have any other unexpected interactions with other extensions, it would be nice to be able to disable them individually.

This PR adds individual project settings for each extension.

What do you folks think?

### DIFF
--- a/plugin/src/main/cpp/register_types.cpp
+++ b/plugin/src/main/cpp/register_types.cpp
@@ -214,12 +214,27 @@ void initialize_plugin_module(ModuleInitializationLevel p_level) {
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


dsnopek:
Per last week's XR team meeting, we're going to stick with the single permission for now, to keep the number of settings smaller and more managable. We can split them out later if needed.

PR #291, opened 2025-04-11
Improve hand tracking related code in demo project

devloglogan:
I use the demo project for testing stuff pretty often and I've wanted to update the code that hides/displays controller and hand models based on hand tracking for a while now, since going off of hand tracking source for everything seems incorrect. Now this should be the case:

- If there is hand tracking data:
  - Hand meshes will display
  - Fb render model will display if loaded and the hand tracking data is inferred from controller data, otherwise controller models are hidden
- If there is no hand tracking data:
  - Hand meshes are hidden
  - Controller models are shown

### DIFF
--- a/demo/hand_controller.gd
+++ b/demo/hand_controller.gd
@@ -1,0 +1,36 @@
+extends XRController3D
+
+@onready var hand_cube: MeshInstance3D = $HandCube
+@onready var render_model: OpenXRFbRenderModel = $ControllerFbRenderModel
+
+func _process(_delta: float) -> void:
+	var hand_tracker_name = "/user/hand_tracker/left" if tracker == "left_hand" \
+			else "/user/hand_tracker/right"
+	var hand_tracker_hand: OpenXRInterface.Hand = OpenXRInterface.HAND_LEFT if tracker == "left_hand" \
+		else OpenXRInterface.HAND_RIGHT
+
+	var interface: OpenXRInterface = XRServer.find_interface("OpenXR")
+
+	var hand_tracker: XRHandTracker = XRServer.get_tracker(hand_tracker_name)
+	if hand_tracker and hand_tracker.has_tracking_data:
+		# If hand tracking data is provided by controller,
+		# and controller model is available, show it with hand mesh.
+		if hand_tracker.hand_tracking_source == XRHandTracker.HAND_TRACKING_SOURCE_CONTROLLER \
+				and render_model.has_render_model_node():
+			render_model.show()
+			interface.set_motion_range(hand_tracker_hand, OpenXRInterface.HAND_MOTION_RANGE_CONFORM_TO_CONTROLLER)
+		# If hand tracking data is not provided by controller,
+		# and controller model is available, hide it.
+		elif render_model.has_render_model_node():
+			render_model.hide()
+			interface.set_motion_range(hand_tracker_hand, OpenXRInterface.HAND_MOTION_RANGE_UNOBSTRUCTED)
+
+		# Always hide placeholder hand cubes if we have hand tracking data.
+		hand_cube.hide()
+	else:
+		# Show render model if availabe, otherwise show hand cubes.
+		if render_model.has_render_model_node():
+			render_model.show()
+			hand_cube.hide()
+		else:
+			hand_cube.show()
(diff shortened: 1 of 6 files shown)


demo/hand_controller.gd
```diff
@@ -0,0 +1,23 @@
...
+		if hand_tracker.hand_tracking_source == XRHandTracker.HAND_TRACKING_SOURCE_CONTROLLER \
+				and render_model.has_render_model_node():
+			render_model.show()
+		elif render_model.has_render_model_node():
+			render_model.hide()
+		else:
+			hand_cube.hide()
+	else:
+		if render_model.has_render_model_node():
+			render_model.show()
```
m4gr3d, line 21: Do we need to explicitly hide the hand cube here?

demo/hand_controller.gd
```diff
@@ -0,0 +1,23 @@
...
+@onready var hand_cube: MeshInstance3D = $HandCube
+@onready var render_model: OpenXRFbRenderModel = $ControllerFbRenderModel
+
+func _process(_delta: float) -> void:
+	var hand_tracker_name = "/user/hand_tracker/left" if tracker == "left_hand" \
+			else "/user/hand_tracker/right"
+
+	var hand_tracker: XRHandTracker = XRServer.get_tracker(hand_tracker_name)
+	if hand_tracker and hand_tracker.has_tracking_data:
+		if hand_tracker.hand_tracking_source == XRHandTracker.HAND_TRACKING_SOURCE_CONTROLLER \
```
m4gr3d, line 12: Can you add comments describing the intended flow.

m4gr3d (changes requested):

demo/hand_controller.gd
devloglogan: Hm in my testing this wasn't a problem but I think you're right in that it should be here, so I've added it.

demo/hand_controller.gd
devloglogan: Comments added.

dsnopek:
Rebasing on `master` should fix CI

dsnopek (commented):
Thanks! This looks great - the logic for hiding/showing the controller and hand meshes makes sense to me.

However, I have one suggestion: Use `XRInterface.set_motion_range()` to make the hand tracking data conform to the controller when showing both the controller and hand meshes.

I tested this out with this patch and it worked nicely:

```patch
diff --git a/demo/hand_controller.gd b/demo/hand_controller.gd
index 3d4b9c7..7ef993d 100644
--- a/demo/hand_controller.gd
+++ b/demo/hand_controller.gd
@@ -6,6 +6,10 @@ extends XRController3D
 func _process(_delta: float) -> void:
        var hand_tracker_name = "/user/hand_tracker/left" if tracker == "left_hand" \
                        else "/user/hand_tracker/right"
+       var hand_tracker_hand: OpenXRInterface.Hand = OpenXRInterface.HAND_LEFT if tracker == "left_hand" \
+                       else OpenXRInterface.HAND_RIGHT
+
+       var interface: OpenXRInterface = XRServer.find_interface("OpenXR")
 
        var hand_tracker: XRHandTracker = XRServer.get_tracker(hand_tracker_name)
        if hand_tracker and hand_tracker.has_tracking_data:
@@ -14,10 +18,13 @@ func _process(_delta: float) -> void:
                if hand_tracker.hand_tracking_source == XRHandTracker.HAND_TRACKING_SOURCE_CONTROLLER \
                                and render_model.has_render_model_node():
                        render_model.show()
+                       interface.set_motion_range(hand_tracker_hand, OpenXRInterface.HAND_MOTION_RANGE_CONFORM_TO_CONTROLLER)
+
                # If hand tracking data is not provided by controller,
                # and controller model is available, hide it.
                elif render_model.has_render_model_node():
                        render_model.hide()
+                       interface.set_motion_range(hand_tracker_hand, OpenXRInterface.HAND_MOTION_RANGE_UNOBSTRUCTED)
 
                # Always hide placeholder hand cubes if we have hand tracking data.
                hand_cube.hide()
```

devloglogan:
@dsnopek ah I'd wondered how this was supposed to be accomplished for a while, glad I know now. Works great. I've applied your patch, thank you!!

dsnopek (approved):

PR #297, opened 2025-04-21
Update docs for meta dynamic resolution

devloglogan:
We renamed the project setting, but never updated the docs! Also the setting is enabled by default and the previous doc didn't say that. This PR fixes that.

I still refer to it as recommended layer resolution once to indicate which OpenXR extension it is, since I figure that makes sense?


dsnopek (approved):
Thanks! Looks good to me

> I still refer to it as recommended layer resolution once to indicate which OpenXR extension it is, since I figure that makes sense?

Yeah, that makes sense. I probably would have used its actual low-level extension name (ie `XR_META_recommended_layer_resolution`) in this case, but I think what you have is fine

m4gr3d (approved):

PR #298, opened 2025-04-29
Fix composition layer next chains

devloglogan:
Some composition layer structs are not always setting next pointers properly, this fixes that.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
@@ -82,7 +82,7 @@ uint64_t OpenXRFbCompositionLayerImageLayoutExtensionWrapper::_set_viewport_comp
 	if (!layer_structs.has(layer)) {
 		layer_structs[layer] = {
 			XR_TYPE_COMPOSITION_LAYER_IMAGE_LAYOUT_FB, // type
-			nullptr, // next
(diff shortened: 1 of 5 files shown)


plugin/src/main/cpp/extensions/openxr_fb_composition_layer_image_layout_extension_wrapper.cpp
```diff
@@ -82,7 +82,7 @@ uint64_t OpenXRFbCompositionLayerImageLayoutExtensionWrapper::_set_viewport_comp
 	if (!layer_structs.has(layer)) {
 		layer_structs[layer] = {
 			XR_TYPE_COMPOSITION_LAYER_IMAGE_LAYOUT_FB, // type
-			nullptr, // next
+			p_next_pointer, // next
```
devloglogan, line 85: In this case the next pointer is *always* set immediately after this, so setting it here feels unneeded. But it also feels strange removing the `next` parameter out of the initialization haha. 🙂 I'm just going to leave it but if anyone else thinks it should be removed, let me know.

dsnopek: I think this is fine. We have to set something here, and we'll only end up setting this twice on the first frame that uses this extension - it'll only be set once on all subsequent frames

dsnopek (approved):
Thanks, this looks good to me!

m4gr3d (approved):

PR #302, opened 2025-05-14
Implement XR_META_boundary_visibility extension

devloglogan:
This PR adds support for the XR_META_boundary_visibility extension. This extension isn't in the OpenXR spec yet, so I've added the ability to provide the path for local Meta OpenXR preview headers to scons/gradle. When provided this will include the headers in the compilation and add a `META_HEADERS_ENABLED` define.

So, if you want to build the plugin with meta headers, run a build like this:
```
./gradlew buildPlugin -Pmeta_headers=absolute/path/to/meta_openxr_preview
```

I've also added `compile_commands.json` to `.gitignore`

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -3,8 +3,8 @@ from glob import glob
 from pathlib import Path
 
 env = SConscript("thirdparty/godot-cpp/SConstruct")
-
-opts = Variables('custom.py')
+opts = Variables('custom.py', ARGUMENTS)
+opts.Add(PathVariable("meta_headers", "Path to the directory containing Meta OpenXR preview headers", None))
 opts.Update(env)
 
 # Add common includes
@@ -12,6 +12,12 @@ env.Append(CPPPATH=[
     "#plugin/src/main/cpp/include/",
     "#thirdparty/openxr/include/",
     ])
+
+# Include Meta OpenXR preview headers if provided
+meta_headers = env.get("meta_headers")
+if meta_headers:
+    env.Append(CPPDEFINES=["META_HEADERS_ENABLED"])
+    env.Append(CPPPATH=[meta_headers])
 
 sources = []
 sources += Glob("#plugin/src/main/cpp/*.cpp")
--- a/build.gradle
+++ b/build.gradle
@@ -91,6 +91,11 @@ def createSconsTasks(File sconsExecutableFile, boolean clean) {
         "build_profile=thirdparty/godot_cpp_build_profile/build_profile.json",
     ]
 
+    if (project.hasProperty("meta_headers")) {
+        def meta_headers_path = project.property("meta_headers")
+        defaultArgs << "meta_headers=${meta_headers_path}"
+    }
+
     if (clean) {
         defaultArgs << "-c"
     }
--- a/plugin/src/main/cpp/export/meta_export_plugin.cpp
+++ b/plugin/src/main/cpp/export/meta_export_plugin.cpp
@@ -433,6 +433,13 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
 		contents += "    <uses-permission android:name=\"com.oculus.permission.USE_SCENE\" />\n";
 	}
 
+#ifdef META_HEADERS_ENABLED
+	// Check for boundary visibility
+	if ((bool)project_settings->get_setting_with_override("xr/openxr/extensions/meta/boundary_visibility")) {
+		contents += "    <uses-permission android:name=\"com.oculus.permission.BOUNDARY_VISIBILITY\" />\n";
+	}
+#endif // META_HEADERS_ENABLED
+
 	// Check for overlay keyboard
 	bool use_overlay_keyboard_option = _get_bool_option("meta_xr_features/use_overlay_keyboard");
 	if (use_overlay_keyboard_option) {
--- a/plugin/src/main/cpp/extensions/openxr_meta_boundary_visibility_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_meta_boundary_visibility_extension_wrapper.cpp
@@ -1,0 +1,157 @@
+/**************************************************************************/
+/*  openxr_meta_boundary_visibility_extension_wrapper.cpp                 */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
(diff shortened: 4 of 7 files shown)


plugin/src/main/cpp/include/extensions/openxr_meta_boundary_visibility_extension_wrapper.h
```diff
@@ -0,0 +1,91 @@
...
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifdef META_HEADERS_ENABLED
+#ifndef OPENXR_META_BOUNDARY_VISIBILITY_EXTENSION_WRAPPER_H
+#define OPENXR_META_BOUNDARY_VISIBILITY_EXTENSION_WRAPPER_H
+
+#include <meta_boundary_visibility.h>
```
dsnopek, line 34: Hm, we probably don't want to setup the include path such that we're doing a "top-level" include like this.

Could we make it so that the include path is one level up, so like:

```
#include "meta_openxr_preview/meta_boundary_visibility.h"
```

This should be pretty easy to do in the `SConstruct`

SConstruct
```diff
@@ -13,6 +13,12 @@ env.Append(CPPPATH=[
     "#thirdparty/openxr/include/",
     ])
 
+# Include Meta OpenXR preview headers if provided
+meta_headers = env.get("meta_headers")
+if meta_headers:
+    env.Append(CPPDEFINES=["META_HEADERS_ENABLED"])
+    env.Append(CPPPATH=[meta_headers])
```
dsnopek, line 20: This is untested, but we could do something like:

```python
meta_headers_parts = os.path.split(meta_headers)
if meta_headers_parts[-1] == 'meta_openxr_preview':
    meta_headers = os.path.join(meta_headers[:-1])
```

Alternatively, we could require that the path end in "meta_openxr_preview" and then always remove it?

Or, we could require the path to the top-level "ovr_openxr_mobile_sdk_XX" directory, and then add the "/OpenXR" do that?

plugin/src/main/cpp/extensions/openxr_meta_boundary_visibility_extension_wrapper.cpp
```diff
@@ -0,0 +1,157 @@
...
+}
+
+bool OpenXRMetaBoundaryVisibilityExtensionWrapper::_on_event_polled(const void *p_event) {
+	if (!meta_boundary_visibility_ext) {
+		return false;
+	}
+
+	if (static_cast<const XrEventDataBuffer *>(p_event)->type == XR_TYPE_EVENT_DATA_BOUNDARY_VISIBILITY_CHANGED_META) {
+		XrEventDataBoundaryVisibilityChangedMETA *boundary_visibility_event = (XrEventDataBoundaryVisibilityChangedMETA *)p_event;
+		current_boundary_visibility = boundary_visibility_event->boundaryVisibility;
```
dsnopek, line 103: Should we emit a signal for this?

dsnopek (commented):
Thanks!

Overall, this looks good to me, and opens a path for supporting some of the vendor extensions that haven't been added to the spec yet.

We'll definitely still need to discuss that with the rest of the team, though - it's a good topic for this week's XR team meeting :-)

plugin/src/main/cpp/include/extensions/openxr_meta_boundary_visibility_extension_wrapper.h
devloglogan: I've made it so this is the case now!

plugin/src/main/cpp/extensions/openxr_meta_boundary_visibility_extension_wrapper.cpp
devloglogan: I've added a signal 🙂

devloglogan:
Addressed David's review and added some XML documentation for the extension wrapper in the latest push.

SConstruct
```diff
@@ -13,6 +14,14 @@ env.Append(CPPPATH=[
     "#thirdparty/openxr/include/",
     ])
 
+# Include Meta OpenXR preview headers if provided
+meta_headers = env.get("meta_headers")
+if meta_headers:
+    if os.path.basename(meta_headers) == "meta_openxr_preview":
```
dsnopek, line 20: I'm not sure using only `os.path.basename()` is sufficient. If the user passes in `absolute/path/to/meta_openxr_preview/` (with a trailing slash), you're going to get `""` as the result.

Maybe:

```suggestion
    meta_headers = meta_headers.rstrip("/\\")
    if os.path.basename(meta_headers) == "meta_openxr_preview":
```

This will also help the `os.path.dirname()` that comes later give the right result

devloglogan: This should be fixed now! I opted for `os.path.normpath()` instead.

dsnopek (approved):
Thanks! This is looking good to me now :-)

build.gradle
```diff
@@ -91,6 +91,11 @@ def createSconsTasks(File sconsExecutableFile, boolean clean) {
         "build_profile=thirdparty/godot_cpp_build_profile/build_profile.json",
     ]
 
+    if (project.hasProperty("meta_headers")) {
+        def meta_headers_path = project.property("meta_headers")
```
m4gr3d: Can you check that `meta_headers_path` is not `null` or empty; you can see an example in https://github.com/godotengine/godot/blob/64b09905c7b2877f8aef99d8b63e73e5d31acfb9/platform/android/java/app/config.gradle#L330

plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -433,6 +433,13 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
 		contents += "    <uses-permission android:name=\"com.oculus.permission.USE_SCENE\" />\n";
 	}
 
+#ifdef META_HEADERS_ENABLED
```
m4gr3d, line 437: We should add a task to track removing those guards when the feature becomes part of the OpenXR spec, and add a TODO pointing to that task.

build.gradle
devloglogan: I've added this check!

plugin/src/main/cpp/export/meta_export_plugin.cpp
devloglogan: I've created issue #304 to track these, and added todo comments to all of the `ifdef`s pointing to it.

PR #305, opened 2025-06-04
Implement XR_FB_color_space extension

devloglogan:
This PR adds support for the XR_FB_color_space extension.

Color space can be set in project settings or via the singleton at runtime, by default I have the color space set to Rec. 709. The reasoning (from my limited understanding of color space) is because Godot works in a sRGB / Rec. 709 color space.

~~I was torn on if this extension should be enabled by default. Setting the color space to Rec. 709 does slightly alter the color output, so those with existing projects may not want it enabled by default? I have it disabled for now, but could easily be swayed if others want it enabled.~~
2025-06-11 Edit: I've made it so that color space is enabled by default. ~~By default it looks like Quest 3 actually uses the Rec. 709 color space, so I don't think this will be problematic at all.~~
2025-06-12 Edit: Just kidding, it definitely uses a different color space than Rec. 709 enabled by default, so I've set the default color space to `Runtime Default`. This way nothing should change for any existing projects.

Also fixed a typo in `openxr_fb_render_model_extension_wrapper.cpp`.

### DIFF
--- a/doc_classes/OpenXRFbColorSpaceExtensionWrapper.xml
+++ b/doc_classes/OpenXRFbColorSpaceExtensionWrapper.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRFbColorSpaceExtensionWrapper" inherits="OpenXRExtensionWrapperExtension" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Wraps the [code]XR_FB_color_space[/code] extension.
+	</brief_description>
+	<description>
+		Wraps the [code]XR_FB_color_space[/code] extension.
+		A color space defines how digital color values are translated into actual visible colors. Specific displays have a native color space that they are capable of producing.
+		This extension allows applications to tell the OpenXR runtime what color space we are rendering in, so that it can potentially translate from that into the native color space of the headset's display.
+		For more information on color management, see: [url=https://developers.meta.com/horizon/resources/color-management-guide/]Color Management in Meta Quest Headsets[/url]
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="get_color_space">
+			<return type="int" enum="OpenXRFbColorSpaceExtensionWrapper.ColorSpace" />
+			<description>
+				Returns the currently active color space.
+			</description>
+		</method>
+		<method name="get_native_color_space">
+			<return type="int" enum="OpenXRFbColorSpaceExtensionWrapper.ColorSpace" />
+			<description>
+				Returns the native color space of the XR device.
+			</description>
+		</method>
+		<method name="get_supported_color_spaces">
+			<return type="Array" />
+			<description>
+				Returns an array of supported color spaces that we can pass to [method set_color_space].
+			</description>
+		</method>
+		<method name="is_enabled">
+			<return type="bool" />
+			<description>
+				Checks if the extension is enabled or not.
+			</description>
+		</method>
+		<method name="set_color_space">
+			<return type="void" />
+			<param index="0" name="color_space" type="int" enum="OpenXRFbColorSpaceExtensionWrapper.ColorSpace" />
+			<description>
+				Sets the color space that the app is rendering in.
+			</description>
+		</method>
+	</methods>
+	<constants>
+		<constant name="COLOR_SPACE_RUNTIME_DEFAULT" value="0" enum="ColorSpace">
+			The color space the OpenXR runtime selects by default, this may vary depending on the XR device.
+			This is not a valid option for [method set_color_space].
+		</constant>
+		<constant name="COLOR_SPACE_UNMANAGED" value="1" enum="ColorSpace">
+			Unmanaged color space. Using this will disable color correction.
+		</constant>
+		<constant name="COLOR_SPACE_REC2020" value="2" enum="ColorSpace">
+			Rec. 2020 color space.
+		</constant>
+		<constant name="COLOR_SPACE_REC709" value="3" enum="ColorSpace">
+			Rec. 709 color space.
+		</constant>
+		<constant name="COLOR_SPACE_RIFT_CV1" value="4" enum="ColorSpace">
+			Rift CV1 color space.
+		</constant>
+		<constant name="COLOR_SPACE_RIFT_S" value="5" enum="ColorSpace">
+			Rift S color space.
+		</constant>
+		<constant name="COLOR_SPACE_QUEST" value="6" enum="ColorSpace">
+			Quest color space.
+		</constant>
+		<constant name="COLOR_SPACE_P3" value="7" enum="ColorSpace">
+			P3 color space.
+		</constant>
+		<constant name="COLOR_SPACE_ADOBE_RGB" value="8" enum="ColorSpace">
+			Adobe RGB color space.
+		</constant>
+	</constants>
+</class>
--- a/plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
@@ -1,0 +1,275 @@
+/**************************************************************************/
+/*  openxr_fb_color_space_extension_wrapper.cpp                           */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "extensions/openxr_fb_color_space_extension_wrapper.h"
+
+#include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/classes/project_settings.hpp>
+
+using namespace godot;
+
+OpenXRFbColorSpaceExtensionWrapper *OpenXRFbColorSpaceExtensionWrapper::singleton = nullptr;
+
+OpenXRFbColorSpaceExtensionWrapper *OpenXRFbColorSpaceExtensionWrapper::get_singleton() {
+	if (singleton == nullptr) {
+		singleton = memnew(OpenXRFbColorSpaceExtensionWrapper());
+	}
+	return singleton;
+}
+
+OpenXRFbColorSpaceExtensionWrapper::OpenXRFbColorSpaceExtensionWrapper() :
+		OpenXRExtensionWrapperExtension() {
+	ERR_FAIL_COND_MSG(singleton != nullptr, "An OpenXRFbColorSpaceExtensionWrapper singleton already exists");
+
+	request_extensions[XR_FB_COLOR_SPACE_EXTENSION_NAME] = &fb_color_space_ext;
+	singleton = this;
+}
+
+OpenXRFbColorSpaceExtensionWrapper::~OpenXRFbColorSpaceExtensionWrapper() {
+	cleanup();
+}
+
+Array OpenXRFbColorSpaceExtensionWrapper::get_supported_color_spaces() {
+	ERR_FAIL_COND_V_MSG(!fb_color_space_ext, Array(), "XR_FB_color_space extension is not enabled");
+
+	if (!supported_color_spaces_set) {
+		populate_supported_color_spaces();
+	}
+
+	ERR_FAIL_COND_V(supported_color_spaces.size() == 0, Array());
+
+	Array ret;
+	ret.resize(supported_color_spaces.size());
+	int i = 0;
+	for (const XrColorSpaceFB &color_space : supported_color_spaces) {
+		ret[i++] = from_openxr_color_space(color_space);
+	}
+	return ret;
+}
+
+OpenXRFbColorSpaceExtensionWrapper::ColorSpace OpenXRFbColorSpaceExtensionWrapper::get_native_color_space() {
+	ERR_FAIL_COND_V_MSG(!fb_color_space_ext, COLOR_SPACE_RUNTIME_DEFAULT, "XR_FB_color_space extension is not enabled");
+	return from_openxr_color_space(system_color_space_properties.colorSpace);
+}
+
+void OpenXRFbColorSpaceExtensionWrapper::set_color_space(ColorSpace p_color_space) {
+	ERR_FAIL_COND_MSG(!fb_color_space_ext, "XR_FB_color_space extension is not enabled");
+	ERR_FAIL_COND_MSG(p_color_space == COLOR_SPACE_RUNTIME_DEFAULT, "Color space must be set to an explicit option when being set at runtime");
+
+	if (p_color_space == current_color_space) {
+		return;
+	}
+
+	if (!supported_color_spaces_set) {
+		populate_supported_color_spaces();
+	}
+
+	XrColorSpaceFB openxr_color_space = to_openxr_color_space(p_color_space);
+	ERR_FAIL_COND_MSG(!supported_color_spaces.has(openxr_color_space), vformat("Color space is not supported: %s", p_color_space));
+
+	XrResult result = xrSetColorSpaceFB(SESSION, openxr_color_space);
+	ERR_FAIL_COND_MSG(XR_FAILED(result), vformat("Failed to set color space [%s]", get_openxr_api()->get_error_string(result)));
+
+	current_color_space = p_color_space;
+}
+
+OpenXRFbColorSpaceExtensionWrapper::ColorSpace OpenXRFbColorSpaceExtensionWrapper::get_color_space() {
+	ERR_FAIL_COND_V_MSG(!fb_color_space_ext, COLOR_SPACE_RUNTIME_DEFAULT, "XR_FB_color_space extension is not enabled");
+	return current_color_space;
+}
+
+void OpenXRFbColorSpaceExtensionWrapper::_bind_methods() {
+	ClassDB::bind_method(D_METHOD("is_enabled"), &OpenXRFbColorSpaceExtensionWrapper::is_enabled);
+
+	ClassDB::bind_method(D_METHOD("get_supported_color_spaces"), &OpenXRFbColorSpaceExtensionWrapper::get_supported_color_spaces);
+
+	ClassDB::bind_method(D_METHOD("get_native_color_space"), &OpenXRFbColorSpaceExtensionWrapper::get_native_color_space);
+
+	ClassDB::bind_method(D_METHOD("set_color_space", "color_space"), &OpenXRFbColorSpaceExtensionWrapper::set_color_space);
+	ClassDB::bind_method(D_METHOD("get_color_space"), &OpenXRFbColorSpaceExtensionWrapper::get_color_space);
+
+	BIND_ENUM_CONSTANT(COLOR_SPACE_RUNTIME_DEFAULT)
+	BIND_ENUM_CONSTANT(COLOR_SPACE_UNMANAGED)
+	BIND_ENUM_CONSTANT(COLOR_SPACE_REC2020)
+	BIND_ENUM_CONSTANT(COLOR_SPACE_REC709)
+	BIND_ENUM_CONSTANT(COLOR_SPACE_RIFT_CV1)
+	BIND_ENUM_CONSTANT(COLOR_SPACE_RIFT_S)
+	BIND_ENUM_CONSTANT(COLOR_SPACE_QUEST)
+	BIND_ENUM_CONSTANT(COLOR_SPACE_P3)
+	BIND_ENUM_CONSTANT(COLOR_SPACE_ADOBE_RGB)
+}
+
+void OpenXRFbColorSpaceExtensionWrapper::cleanup() {
+	fb_color_space_ext = false;
+	supported_color_spaces_set = false;
+}
+
+void OpenXRFbColorSpaceExtensionWrapper::populate_supported_color_spaces() {
+	if (supported_color_spaces_set) {
+		return;
+	}
+
+	uint32_t num_color_spaces = 0;
+	XrResult result = xrEnumerateColorSpacesFB(SESSION, 0, &num_color_spaces, nullptr);
+	ERR_FAIL_COND_MSG(XR_FAILED(result), vformat("Failed to get supported color space count [%s]", get_openxr_api()->get_error_string(result)));
+
+	supported_color_spaces.resize(num_color_spaces);
+	result = xrEnumerateColorSpacesFB(SESSION, num_color_spaces, &num_color_spaces, supported_color_spaces.ptrw());
+	if (XR_FAILED(result)) {
+		supported_color_spaces.clear();
+		ERR_FAIL_MSG(vformat("Failed to get supported color spaces [%s]", get_openxr_api()->get_error_string(result)));
+	}
+
+	supported_color_spaces_set = true;
+}
+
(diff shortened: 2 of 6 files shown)


plugin/src/main/cpp/include/extensions/openxr_fb_color_space_extension_wrapper.h
```diff
@@ -0,0 +1,121 @@
...
+	};
+
+	uint64_t _set_system_properties_and_get_next_pointer(void *next_pointer) override;
+
+	godot::Dictionary _get_requested_extensions() override;
+
+	void _on_instance_created(uint64_t instance) override;
+	void _on_instance_destroyed() override;
+	void _on_session_created(uint64_t p_session);
+	void _on_state_ready();
```
dsnopek:
```suggestion
	void _on_session_created(uint64_t p_session) override;
	void _on_state_ready() override;
```

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
```diff
@@ -0,0 +1,274 @@
...
+	}
+}
+
+uint64_t OpenXRFbColorSpaceExtensionWrapper::_set_system_properties_and_get_next_pointer(void *p_next_pointer) {
+	if (fb_color_space_ext) {
+		system_color_space_properties.next = p_next_pointer;
+		p_next_pointer = &system_color_space_properties;
+	}
+
+	return reinterpret_cast<uint64_t>(&p_next_pointer);
```
dsnopek: This is returning `void **` rather than `void *`. It should be:

```suggestion
	return reinterpret_cast<uint64_t>(p_next_pointer);
```

plugin/src/main/cpp/include/extensions/openxr_fb_color_space_extension_wrapper.h
```diff
@@ -0,0 +1,121 @@
...
+	bool fb_color_space_ext = false;
+	bool supported_color_spaces_set = false;
+
+	XrSystemColorSpacePropertiesFB system_color_space_properties = {
+		XR_TYPE_SYSTEM_COLOR_SPACE_PROPERTIES_FB, // type
+		nullptr, // next
+	};
+
+	Vector<XrColorSpaceFB> supported_color_spaces;
+	ColorSpace current_color_space;
```
dsnopek:
```suggestion
	ColorSpace current_color_space = COLOR_SPACE_UNMANAGED;
```

Otherwise, this will contain garbage data until the OpenXR session starts.

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
```diff
@@ -0,0 +1,274 @@
...
+void OpenXRFbColorSpaceExtensionWrapper::_on_instance_destroyed() {
+	cleanup();
+}
+
+void OpenXRFbColorSpaceExtensionWrapper::_on_session_created(uint64_t p_session) {
+	if (!fb_color_space_ext) {
+		return;
+	}
+
+	current_color_space = from_openxr_color_space(system_color_space_properties.colorSpace);
```
dsnopek: Hm, personally, I would have expected `current_color_space` to be unmanaged until explicitly set by the user, rather using the native color space. How do we want this to work?

Also, I think we should probably clear `current_color_space` when the OpenXR session ends, because I don't think the value is retained between sessions

plugin/src/main/cpp/include/extensions/openxr_fb_color_space_extension_wrapper.h
```diff
@@ -0,0 +1,121 @@
...
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#ifndef OPENXR_FB_COLOR_SPACE_EXTENSION_WRAPPER_H
+#define OPENXR_FB_COLOR_SPACE_EXTENSION_WRAPPER_H
```
dsnopek: This is fine, but now that Godot and godot-cpp have switched to `#pragma once`, I've been trying to use that rather than traditional header guards in all new headers

dsnopek (commented):
Thanks!

Should we enable this extension in one (or maybe all?) of our sample projects? It is something that most folks should probably enable.

I did a little testing, and experienced a weird issue in the main demo project, where if I enabled the color space extension, then render models would stop working. I think it might be related to the `void **` vs `void *` thing that I pointed out.

Otherwise, it seems to be working! I made minor changes to the main demo project so I could change color spaces with a button press, and was able to notice minor differences.

doc_classes/OpenXRFbColorSpaceExtensionWrapper.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRFbColorSpaceExtensionWrapper" inherits="OpenXRExtensionWrapperExtension" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Wraps the [code]XR_FB_color_space[/code] extension.
+	</brief_description>
+	<description>
+		Wraps the [code]XR_FB_color_space[/code] extension.
+	</description>
```
m4gr3d, line 11: Let's expand on the description using the color space sample as reference: https://github.com/meta-quest/Meta-OpenXR-SDK/tree/main/Samples/XrSamples/XrColorSpaceFB.

We should also point to this documentation about color space management on Quest devices: https://developers.meta.com/horizon/resources/color-management-guide/

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
```diff
@@ -0,0 +1,274 @@
...
+	XrColorSpaceFB openxr_color_space = to_openxr_color_space(p_color_space);
+	ERR_FAIL_COND_MSG(!supported_color_spaces.has(openxr_color_space), vformat("Color space is not supported: %s", p_color_space));
+
+	XrResult result = xrSetColorSpaceFB(SESSION, openxr_color_space);
+	if (XR_FAILED(result)) {
+		UtilityFunctions::printerr("OpenXR: Failed to set color space [", get_openxr_api()->get_error_string(result), "]");
+		return;
+	}
+
+	current_color_space = p_color_space;
```
m4gr3d, line 99: We should return early if `p_color_space == current_color_space`.

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
```diff
@@ -0,0 +1,274 @@
...
+	supported_color_spaces.resize(num_color_spaces);
+
+	result = xrEnumerateColorSpacesFB(SESSION, num_color_spaces, &num_color_spaces, supported_color_spaces.ptrw());
+	if (XR_FAILED(result)) {
+		UtilityFunctions::printerr("OpenXR: Failed to get supported color spaces [", get_openxr_api()->get_error_string(result), "]");
+		supported_color_spaces.clear();
+		return;
+	}
+
+	supported_color_spaces_set = true;
```
m4gr3d, line 149: Do we want to return early is `supported_color_spaces_set` is already set?

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
m4gr3d: Yes, the [app or source color space](https://developers.meta.com/horizon/resources/color-management-guide/#source-color-space) is different from the [target device color space](https://developers.meta.com/horizon/resources/color-management-guide/#target-color-space), so this is incorrect. 

For backward compat (i.e: avoid changing the look of existing Godot XR apps), we should default to `unmanaged`, but we should add a warning that setting an explicit color space is recommended as otherwise it causes apps to look different when using different HMDs as mentioned in https://developers.meta.com/horizon/resources/color-management-guide/#source-color-spaces-we-offer-in-our-apis

plugin/src/main/cpp/register_types.cpp
```diff
@@ -394,6 +401,7 @@ void add_plugin_project_settings() {
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/anchor_api", false);
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/anchor_sharing", false);
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/scene_api", false);
+	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/color_space", false);
```
m4gr3d: I think this should be enabled by default as it allows us to recommend that apps explicitly set their desired color space as mentioned in previous comments.

Otherwise this is likely to not get noticed by devs (unless they explicitly know about it), leading to Godot XR apps to have an unmanaged color space by default and looking different across different devices.

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
m4gr3d: Looking at the Unity documentation, we may want to introduce an `unknown` value and use that as the default: https://developers.meta.com/horizon/documentation/unity/unity-color-space/

`unmanaged` seems to explicitly disable color correction, whereas the default app color space if unspecified seems to change per devices:
- Oculus Rift: Between Adobe RGB and DCI-P3
- Oculus Rift S, Meta Quest 2: Rec 709
- Meta Quest: Rec 2020

m4gr3d: I take my initial comment back, and @dsnopek it looks like we were both incorrect as to the default behavior..

The OpenXR documentation states:
> If this function is not called, the session will use the color space deemed appropriate by the runtime. Oculus HMDs for both PC and Mobile product lines default to XR_COLOR_SPACE_RIFT_CV1_FB.

So it looks like if the default is not specified, then `system_color_space_properties.colorSpace` is the color space used when displaying the app at runtime.

Since we won't know that value until runtime, we should introduce the `unknown` value as the default, with the documentation stating that this means the session will use the color space deemed appropriate by the runtime.

Note that an alternative to `unkown` could be `runtime_default` or `headset_default`, but then we should be explicit in the documentation (and linting) that these are not recommended values as it would cause the content to look different on different devices.

m4gr3d (changes requested):

plugin/src/main/cpp/include/extensions/openxr_fb_color_space_extension_wrapper.h
devloglogan: Updated this to `COLOR_SPACE_RUNTIME_DEFAULT`

doc_classes/OpenXRFbColorSpaceExtensionWrapper.xml
devloglogan: I've expanded upon it a bit!

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
devloglogan: I think that makes sense, I've added it.

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
devloglogan: I've added `COLOR_SPACE_RUNTIME_DEFAULT` as a value.

Despite what the OpenXR spec says, the runtime default (at least for my Quest 3) is Rec 709, not Rift CV1. I am thinking that the runtime default and the device native color space are one in the same (this appears to be the case for my Quest 3), but maybe I'm incorrect here?

plugin/src/main/cpp/register_types.cpp
devloglogan: I've enabled it by default. 🙂

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
```diff
@@ -0,0 +1,291 @@
...
+		} break;
+		case COLOR_SPACE_P3: {
+			return XR_COLOR_SPACE_P3_FB;
+		} break;
+		case COLOR_SPACE_ADOBE_RGB: {
+			return XR_COLOR_SPACE_ADOBE_RGB_FB;
+		} break;
+		case COLOR_SPACE_RUNTIME_DEFAULT:
+		default: {
+			ERR_FAIL_V_MSG(XR_COLOR_SPACE_UNMANAGED_FB, vformat("Unknown color space: %s", p_color_space));
```
m4gr3d: We shouldn't default to `XR_COLOR_SPACE_UNMANAGED_FB` since this applies different settings than the runtime default.
Can we just return `XR_COLOR_SPACE_MAX_ENUM_FB` instead?

plugin/src/main/cpp/register_types.cpp
```diff
@@ -417,6 +425,22 @@ void add_plugin_project_settings() {
 		project_settings->add_property_info(property_info);
 	}
 
+	{
+		String starting_color_space = "xr/openxr/extensions/meta/color_space/starting_color_space";
+		if (!project_settings->has_setting(starting_color_space)) {
+			project_settings->set_setting(starting_color_space, 2);
```
m4gr3d: The default value should be `Runtime Default`

plugin/src/main/cpp/register_types.cpp
```diff
@@ -417,6 +425,22 @@ void add_plugin_project_settings() {
...
+		}
+
+		project_settings->set_initial_value(starting_color_space, 2);
+		project_settings->set_as_basic(starting_color_space, false);
+		Dictionary property_info;
+		property_info["name"] = starting_color_space;
+		property_info["type"] = Variant::Type::INT;
+		property_info["hint"] = PROPERTY_HINT_ENUM;
+		property_info["hint_string"] = "Unmanaged:0,REC2020:1,REC709:2,Rift CV1:3,Rift S:4,Quest:5,P3:6,Adobe RGB:7,Runtime Default:8";
```
m4gr3d: `Runtime Default` should be zeroth index value.

plugin/src/main/cpp/register_types.cpp
devloglogan: @dsnopek and I had discussed this a while ago, since Godot works in sRGB / Rec 709 space, I had thought setting Rec 709 as the default would be desirable here.

dsnopek: This is a tricky one.

I do think Rec 709 is what most people should be using, because that's what Godot's import system + renderer is going to produce. No other option really makes sense, until Godot can render assets that were designed for other color spaces. So, this would be a good default.

However, this will subtly change colors for existing apps that may not expect it. So, making the default be "Runtime Default" could make sense because it means existing apps won't change at all.

I'm really not sure which is the right way to go

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
```diff
@@ -0,0 +1,291 @@
...
+void OpenXRFbColorSpaceExtensionWrapper::_on_instance_destroyed() {
+	cleanup();
+}
+
+void OpenXRFbColorSpaceExtensionWrapper::_on_session_created(uint64_t p_session) {
+	if (!fb_color_space_ext) {
+		return;
+	}
+
+	current_color_space = from_openxr_color_space(system_color_space_properties.colorSpace);
```
dsnopek: Wait, I'm still not sure this is right. Isn't `system_color_space_properties.colorSpace` the device's native color space and not the current color space?

The Quest 3 should be returning Rec 709 here, because that's its native colorspace. But we still see a subtle color difference when we set the colorspace to Rec 709. Which leads me to think that it doesn't necessary start on the native color space, but could be some other value.

Now that `set_color_space()` has this code:

```
	if (p_color_space == current_color_space) {
		return;
	}
```

... I think setting the color space to Rec 709 would fail on the Quest 3, because `current_color_space` is Rec 709.

Perhaps we shouldn't even have this line? Then it won't change the `current_color_space` on session start, and it would still be "runtime default" until we explicitly set it

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
dsnopek: > So it looks like if the default is not specified, then `system_color_space_properties.colorSpace` is the color space used when displaying the app at runtime.

If this was true, then we wouldn't see the subtle color difference we see when setting the color space to Rec 709 on Quest 3, because isn't its native color space Rec 709?

plugin/src/main/cpp/register_types.cpp
devloglogan: I've swapped the default to `Runtime Default`!

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
devloglogan: Huh I thought I'd maybe made an incorrect observation before, and confirmed that the native color space was the runtime default (at least on Quest 3), but you're definitely correct. I've removed the session created function override entirely, so I don't believe this is a problem anymore!

This would have been a lot simpler if OpenXR gave us a way to check what color space is active; but, alas. 🙂

doc_classes/OpenXRFbColorSpaceExtensionWrapper.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRFbColorSpaceExtensionWrapper" inherits="OpenXRExtensionWrapperExtension" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Wraps the [code]XR_FB_color_space[/code] extension.
+	</brief_description>
+	<description>
+		Wraps the [code]XR_FB_color_space[/code] extension.
+		A color space indicates the range of colors that a device can display. This extension allows applications to see the native color space of the XR device, see which color spaces are supported, and set the active color space.
```
dsnopek:
```suggestion
		A color space defines how digital color values are translated into actual visible colors. Specific displays have a native color space that they are capable of producing.
		This extension allows applications to tell the OpenXR runtime what color space we are rendering in, so that it can potentially translate from that into the native color space of the headset's display.
```

doc_classes/OpenXRFbColorSpaceExtensionWrapper.xml
```diff
@@ -0,0 +1,76 @@
...
+		<method name="get_native_color_space">
+			<return type="int" enum="OpenXRFbColorSpaceExtensionWrapper.ColorSpace" />
+			<description>
+				Returns the native color space of the XR device.
+			</description>
+		</method>
+		<method name="get_supported_color_spaces">
+			<return type="Array" />
+			<description>
+				Returns an array of color spaces supported by the XR device.
```
dsnopek: "supported by the XR device" is confusing here, because the display only has its one native color space. This is a list of color spaces that we can use for `set_color_space()`, which the device will then use to translate to the device's native color space.

```suggestion
				Returns an array of supported color spaces that we can pass to [method set_color_space].
```

doc_classes/OpenXRFbColorSpaceExtensionWrapper.xml
```diff
@@ -0,0 +1,76 @@
...
+			<return type="bool" />
+			<description>
+				Checks if the extension is enabled or not.
+			</description>
+		</method>
+		<method name="set_color_space">
+			<return type="void" />
+			<param index="0" name="color_space" type="int" enum="OpenXRFbColorSpaceExtensionWrapper.ColorSpace" />
+			<description>
+				Set the color space.
```
dsnopek:
```suggestion
				Sets the color space that the app is rendering in.
```

doc_classes/OpenXRFbColorSpaceExtensionWrapper.xml
```diff
@@ -0,0 +1,76 @@
...
+			</description>
+		</method>
+	</methods>
+	<constants>
+		<constant name="COLOR_SPACE_RUNTIME_DEFAULT" value="0" enum="ColorSpace">
+			The color space the OpenXR runtime selects by default, this may vary depending on the XR device.
+			This is not a valid option for [method set_color_space].
+		</constant>
+		<constant name="COLOR_SPACE_UNMANAGED" value="1" enum="ColorSpace">
+			Unmanaged color space.
```
dsnopek:
```suggestion
			Unmanaged color space. Using this will disable color correction.
```

dsnopek (commented):
The code for this is looking great to me now! I retested, all seems to be working as expected.

I've gone through the docs, and attempted to clarify them for my understanding of how this extension works (which I hope is correct :-)).

I've been debating whether or not it makes sense to specifically say that at the moment Godot can only render in Rec 709, so the only rational options are either (1) Rec 709 for correctness, or (2) "Runtime Default" for existing projects that don't like how using Rec 709 changes their colors.

But I'm not sure the best way to say that, and it also may not be true forever (Godot may one day be able to render Rec 2020 as part of HDR support?), so I didn't include it in my suggestions. Might be better for a tutorial? Or, if you have a good idea for how to integrate, feel free to

I also think we should update all our samples to use this extension with the color space set to Rec 709, but we don't necessarily have to do that in this PR

devloglogan:
@dsnopek thanks for the docs suggestions, I've applied them. 🙂

m4gr3d (approved):
Looks good! 

Quick reminder about adding the warning for 4.x

devloglogan:
Added a warning on startup if the project setting isn't set to Rec. 709, does this work for you @m4gr3d?

I also updated some of the other error printing.

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+void OpenXRFbColorSpaceExtensionWrapper::_on_state_ready() {
+	ERR_FAIL_COND_MSG(!fb_color_space_ext, "XR_FB_color_space extension is not enabled");
+
+	ProjectSettings *project_settings = ProjectSettings::get_singleton();
+	ERR_FAIL_NULL(project_settings);
+
+	ColorSpace starting_color_space = ColorSpace((int)project_settings->get_setting_with_override("xr/openxr/extensions/meta/color_space/starting_color_space"));
+	if (starting_color_space != COLOR_SPACE_REC709) {
+		WARN_PRINT("Recommended color space project setting is REC709");
```
m4gr3d, line 263: Looks good!

Alongside this warning, let's also include a warning in the export window by updating https://github.com/GodotVR/godot_openxr_vendors/blob/001d03c087c3761b96debceb037e7be3e366eaa7/plugin/src/main/cpp/export/export_plugin.cpp#L184

m4gr3d (changes requested):

plugin/src/main/cpp/extensions/openxr_fb_color_space_extension_wrapper.cpp
dsnopek: It could be a little confusing getting a warning about a project setting, when in the export settings area - folks could start looking in the wrong place

devloglogan: I agree, feels strange to have a warning about project settings in export settings.

m4gr3d: I disagree; with the switch to gating the extensions via the project settings, and moving related export options to project settings, the two are intrinsically linked. 
Taking a look at `meta_export_plugin.cpp#_get_export_option_warning(...)` most of them check a project setting in order to dispatch a warning. I see this as following a similar approach.

The export window is also the most effective place to surface the warning (up until we update the default in `v5`), because a warning in the logs can be easily missed, but here we prompt you, right before you are about to export your game, to ensure it looks correct by selecting the adequate color space.

If there is concern about folks looking in the wrong place, let's just be explicit about how they can address the warning. We do the same thing with the export option, where we specify the name of the export option that needs to be udpated.

dsnopek: > If there is concern about folks looking in the wrong place, let's just be explicit about how they can address the warning. We do the same thing with the export option, where we specify the name of the export option that needs to be udpated.

My worry is that we'll tell them to look for a particular named option, but they'll be looking for it in export settings, where they should really be in project settings

> Taking a look at `meta_export_plugin.cpp#_get_export_option_warning(...)` most of them check a project setting in order to dispatch a warning. I see this as following a similar approach.

Those messages are a little different. While the code is looking at the project settings, the warnings are instructing the user to change something on the export settings.

Whereas this would be a warning instructing the user to change a project setting.

If we could link to the setting, the way that Godot does for this export error:

Selection_352

... that would be great! There's no confusion if the user clicks and goes to the right place. But I don't know that we have an easy way to do that here

devloglogan: @m4gr3d @dsnopek ultimately I'm fine with either way, and  it's true that without adding another warning somewhere, it's entirely possible developers would miss the warning entirely.

I've added a warning to the export window as well. Let me know if you find this acceptable!

plugin/src/main/cpp/export/export_plugin.cpp
```diff
@@ -200,6 +200,16 @@ String OpenXREditorExportPlugin::_get_export_option_warning(const Ref<EditorExpo
 				return "\"Disable " + _vendor.capitalize() + " Plugin before enabling another. Multiple plugins are not supported!\"";
 			}
 		}
+
+		if (_vendor == "meta") {
```
m4gr3d: I was going to suggest removing this gate, but looking at the support matrix, [only meta runtimes support `XR_FB_color_space`](https://github.khronos.org/OpenXR-Inventory/extension_support.html#XR_FB_color_space). Given that, this logic should be moved within `meta_export_plugin.cpp` instead.

devloglogan: Makes sense to me. Change applied! I think this PR should be good to merge now.

PR #307, opened 2025-06-18
Make REC709 default for XR_FB_color_space

devloglogan:
This PR builds on top of #305.

Per discussion at the last XR meeting, the XR_FB_color_space implementation will initially use the runtime default color space in 4.x of the vendor plugin. When we move to 5.0 though, the default option will be updated to Rec. 709 and the warnings encouraging devs to use it will be removed.

This PR will remain as a draft until the XR_FB_color_space implementation is merged and until we begin merging changes for plugin version 5.0

### DIFF
--- a/plugin/src/main/cpp/register_types.cpp
+++ b/plugin/src/main/cpp/register_types.cpp
@@ -455,10 +455,10 @@ void add_plugin_project_settings() {
 	{
 		String starting_color_space = "xr/openxr/extensions/meta/color_space/starting_color_space";
 		if (!project_settings->has_setting(starting_color_space)) {
-			project_settings->set_setting(starting_color_space, 0);
+			project_settings->set_setting(starting_color_space, 3);
 		}
 
-		project_settings->set_initial_value(starting_color_space, 0);
+		project_settings->set_initial_value(starting_color_space, 3);
 		project_settings->set_as_basic(starting_color_space, false);
 		Dictionary property_info;
 		property_info["name"] = starting_color_space;


plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -289,12 +289,6 @@ String MetaEditorExportPlugin::_get_export_option_warning(const Ref<EditorExport
 		if (!openxr_enabled && _get_bool_option(option)) {
 			return "\"Instant splash screen\" is only valid when \"XR Mode\" is \"OpenXR\".\n";
 		}
-	} else if (option == "xr_features/enable_meta_plugin") {
-		if ((bool)project_settings->get_setting_with_override("xr/openxr/extensions/meta/color_space")) {
-			if ((int)project_settings->get_setting_with_override("xr/openxr/extensions/meta/color_space/starting_color_space") != 3) {
```
m4gr3d: We should keep the warning if the developer mistakenly sets the color space to Runtime Default.

This suggests to me that in the current 4.x logic, we should only show the warning if the setting is set to the default value of Runtime Default, so that if a developer explicitly sets the color space to a non-default value, we don't continuously pester them with this warning.

devloglogan: Just to make sure I understand before making another small PR, you're suggesting:
- Make a new PR updating the "use REC709" warnings to only appear when runtime default is selected
- Do not remove those warnings in this PR, just update the default to REC709

That makes sense to me!

m4gr3d: Yeah, that's correct!

devloglogan: I've applied the changes and opened #322 🙂

m4gr3d (approved):

dsnopek (approved):
Looks good to me!

PR #310, opened 2025-06-19
Update OpenXR to 1.1.49 release

devloglogan:
Updates OpenXR header files to 1.1.49 release!

Issue #304, opened 2025-06-03
Remove check for Meta headers when features become part of OpenXR spec


devloglogan:
This issue serves as a tracker for the removal of preprocessor checks for `META_HEADERS_ENABLED`, which is defined for building with Meta OpenXR extensions not yet a part of the OpenXR spec.

Above each instance of this I'll insert the following todo comment:
```
// @todo GH Issue 304: Remove check for meta headers when feature becomes part of OpenXR spec.
```

Right now only #302 adds this but I'll soon create another PR for the additional body tracking extensions that will also rely on this define. I'll update this issue to track those as well when I make that PR.

<ins>Extensions to have preprocessor checks removed</ins>
- [ ] XR_META_boundary_visibility

devloglogan:
Added META_body_tracking_fidelity and META_body_tracking_calibration to the list now that #311 has been merged.

PR #311, opened 2025-06-19
Implement additional Meta body tracking extensions

devloglogan:
~~Builds on top of #310, so this PR will remain as a draft until the first is merged.~~

This PR implements the following Meta body tracking extensions, adding functionality to the preexisting `OpenXRFbBodyTrackingExtensionWrapper`:
- META_body_tracking_full_body
- META_body_tracking_fidelity
- META_body_tracking_calibration

META_body_tracking_full_body provides tracking for the lower body joints on a Godot humanoid skeleton! You can test this with the Meta body tracking sample. In order to keep compatibility, not all full body joints are implemented, as this would require the additional joints in `XRBodyTracker` added in https://github.com/godotengine/godot/pull/107220. These additional joints can be added in a future PR.

META_body_tracking_fidelity and META_body_tracking_calibration are not yet a part of the OpenXR spec, so their implementations are in `META_HEADERS_ENABLED` preprocessor checks. Once this PR is merged I'll add them to the issue tracking these additions: #304

### DIFF
--- a/doc_classes/OpenXRFbBodyTrackingExtensionWrapper.xml
+++ b/doc_classes/OpenXRFbBodyTrackingExtensionWrapper.xml
@@ -8,4 +8,77 @@
 	</description>
 	<tutorials>
 	</tutorials>
+	<methods>
+		<method name="get_body_tracking_calibration_state">
+			<return type="int" enum="OpenXRFbBodyTrackingExtensionWrapper.BodyTrackingCalibrationState" />
+			<description>
+				Returns the body tracking calibration state.
+			</description>
+		</method>
+		<method name="get_body_tracking_fidelity_status">
+			<return type="int" enum="OpenXRFbBodyTrackingExtensionWrapper.BodyTrackingFidelity" />
+			<description>
+				Returns the body tracking fidelity status.
+			</description>
+		</method>
+		<method name="is_body_tracking_fidelity_supported">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the body tracking fidelity extension is supported.
+			</description>
+		</method>
+		<method name="is_body_tracking_height_override_supported">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if height override from the body tracking calibration extension is supported.
+			</description>
+		</method>
+		<method name="is_full_body_tracking_supported">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the body tracking full body extension is supported.
+			</description>
+		</method>
+		<method name="request_body_tracking_fidelity">
+			<return type="void" />
+			<param index="0" name="fidelity" type="int" enum="OpenXRFbBodyTrackingExtensionWrapper.BodyTrackingFidelity" />
+			<description>
+				Request that the OpenXR runtime use a specified body tracking fidelity.
+			</description>
+		</method>
+		<method name="reset_body_tracking_calibration">
+			<return type="void" />
+			<description>
+				Reset the body tracking calibration state.
+			</description>
+		</method>
+		<method name="suggest_body_tracking_height_override">
+			<return type="void" />
+			<param index="0" name="body_height" type="float" />
+			<description>
+				Suggest a height override in meters for body tracking calibration to the OpenXR runtime.
+				[b]Note:[/b] this value must be within the range of 0.5 and 3.0 meters.
+			</description>
+		</method>
+	</methods>
+	<constants>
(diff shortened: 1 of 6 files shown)


plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -255,28 +345,53 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
...
+	ERR_FAIL_COND_MSG(XR_FAILED(result), vformat("Failed to get body joint locations: ", get_openxr_api()->get_error_string(result)));
 
 	// Set the tracking active flag
 	xr_body_tracker->set_has_tracking_data(locations.isActive);
 
 	// Process all joints
 	for (const JointMapEntry &entry : joint_table) {
+		// Skip full body joints if extension is not supported.
+		if (!is_body_tracking_full_body_supported() && entry.fb_joint >= XR_BODY_JOINT_COUNT_FB) {
+			continue;
```
dsnopek: Since we don't want to loop anymore, this could be `break`:

```suggestion
			break;
```

plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -255,28 +345,53 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
...
+#endif // META_HEADERS_ENABLED
+
 	// Construct the locations struct.
-	XrBodyJointLocationFB fb_locations[XR_BODY_JOINT_COUNT_FB] = {};
+	int fb_joint_count = XR_BODY_JOINT_COUNT_FB;
+	if (meta_body_tracking_full_body_ext && is_body_tracking_full_body_supported()) {
+		fb_joint_count = XR_FULL_BODY_JOINT_COUNT_META;
+	}
+
+	LocalVector<XrBodyJointLocationFB> fb_locations;
```
dsnopek: Rather than using `LocalVector`, you could just make a normal array of the full body joint count (the bigger of the two) and then just tell OpenXR that we have the right number of elements for the current joint set. This would be somewhat faster since `LocalVector` allocates memory on the heap, and a normal array would be on the stack.

plugin/src/main/cpp/extensions/openxr_fb_body_tracking_extension_wrapper.cpp
```diff
@@ -255,28 +345,53 @@ void OpenXRFbBodyTrackingExtensionWrapper::_on_process() {
...
+	if (meta_body_tracking_calibration_ext) {
+		body_tracking_calibration_status.next = next_pointer;
+		next_pointer = &body_tracking_calibration_status;
+	}
+#endif // META_HEADERS_ENABLED
+
 	// Construct the locations struct.
-	XrBodyJointLocationFB fb_locations[XR_BODY_JOINT_COUNT_FB] = {};
+	int fb_joint_count = XR_BODY_JOINT_COUNT_FB;
+	if (meta_body_tracking_full_body_ext && is_body_tracking_full_body_supported()) {
```
dsnopek: I think we should cache `meta_body_tracking_full_body_ext && is_body_tracking_full_body_supported()` in a local variable in this function, so we can avoid calling `is_body_tracking_full_body_supported()` so many times.

This function is called every frame, so it'd be good to err on the side of efficiency.

devloglogan:
@dsnopek thanks for the review! I've applied all suggestions. I've also updated the vrs file so CI works again.

Seeing as the prerequisite PR has been merged, I'll open this PR now!

doc_classes/OpenXRFbBodyTrackingExtensionWrapper.xml
```diff
@@ -8,4 +8,77 @@
...
+				Returns the body tracking fidelity status.
+			</description>
+		</method>
+		<method name="is_body_tracking_fidelity_supported">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the body tracking fidelity extension is supported.
+			</description>
+		</method>
+		<method name="is_body_tracking_full_body_supported">
```
m4gr3d: nit: `is_full_body_tracking_supported` would read / 'sound' better

devloglogan: Works for me, I've applied your suggestion. 🙂

PR #312, opened 2025-06-23
[DRAFT] Allow switching hybrid mode on desktop for testing

dsnopek:
This PR allows switching hybrid mode between immersive and panel on desktop for testing. When paired with the Meta XR Simulator or Monado's QWERTY driver, this can allow developers to test out mode switching without putting on a headset.

When running from the editor, it should work with the embedded game view (only tested on Linux with Monado so far), and still have a working debugger and remote scene tree after the mode switch.

Marked as a DRAFT for now, because it depends:

- https://github.com/godotengine/godot/pull/107671
- https://github.com/GodotVR/godot_openxr_vendors/pull/308

Since this depends on `extension_api.json` changes that won't be available until Godot 4.6 at the earliest, we won't be able to merge this until this plugin is targeting Godot 4.6+


plugin/src/main/cpp/classes/openxr_hybrid_app.cpp
```diff
@@ -46,26 +52,34 @@ void OpenXRHybridApp::_bind_methods() {
...
+bool OpenXRHybridApp::is_hybrid_app_enabled() const {
+	return ProjectSettings::get_singleton()->get_setting_with_override("xr/hybrid_app/enabled");
+}
 
-	return os->has_feature("godot_openxr_hybrid_app");
+bool OpenXRHybridApp::is_hybrid_app() const {
+#ifdef ANDROID_ENABLED
+	return OS::get_singleton()->has_feature("godot_openxr_hybrid_app");
+#else
+	Ref<OpenXRInterface> xr_interface = XRServer::get_singleton()->find_interface("OpenXR");
```
m4gr3d, line 63: Is the `OpenXR` interface available when in panel mode?

On Android, we pass `--xr-mode off` when launching the panel, so I'd expect OpenXR to be unavailable. Is that different on desktop?

plugin/src/main/cpp/classes/openxr_hybrid_app.cpp
```diff
@@ -75,25 +89,109 @@ bool OpenXRHybridApp::switch_mode(HybridMode p_mode, const String &p_data) {
...
+		mode_args.push_back("off");
+		mode_args.push_back("--xr_mode_regular");
+	}
+
+	if (!p_data.is_empty()) {
+		mode_args.push_back("--openxr-hybrid-data");
+		mode_args.push_back(Marshalls::get_singleton()->utf8_to_base64(p_data));
+	}
+
+	// This path works when running from the editor.
```
m4gr3d, line 120: Should it be gated with `#ifdef TOOLS_ENABLED`?

plugin/src/main/cpp/classes/openxr_hybrid_app.cpp
dsnopek: > Is the `OpenXR` interface available when in panel mode?

Yes. If the engine was compiled with OpenXR support, then the interface will _exist_, so `find_interface("OpenXR")` will return it. But if `--xr-mode off` was passed, then it won't be initialized (ie `xr_interface->is_initialized()` will return `false`).

I'm using this as a way to determine if OpenXR is even a possibility on the given platform. If you were running on iOS or the Web, for example, then `find_interface("OpenXR")` would return a null reference.

plugin/src/main/cpp/classes/openxr_hybrid_app.cpp
dsnopek: We could, but we're not making distinct editor and debug builds for this extension - we're just making debug and release builds, so it wouldn't make any difference. I don't think we're currently using `TOOLS_ENABLED` anywhere, even in our editor plugins

PR #313, opened 2025-06-24
Prevent "XR_FB_color_space extension is not enabled" error on session start

dsnopek:
Presently, if you enable the `XR_FB_color_space` extension in project settings, and then you run on a runtime that doesn't support that extension, you'll always get this error when the OpenXR session starts.

I don't think this should be an error in `_state_ready()` - only in functions that the developer explicitly called.

In this case, I think we should just silently return (which is what this PR does).


devloglogan (approved):
Agreed!

m4gr3d (approved):

PR #314, opened 2025-06-24
Use the Meta OpenXR preview headers in CI builds

dsnopek:
This PR will download and use the Meta OpenXR preview headers in CI builds. This would also mean that our releases will include support for the features that are only available with those headers.

I am not a lawyer, but my personal interpretation of [the license](https://developers.meta.com/horizon/licenses/oculussdk/) is that this should be OK for us to do.

However, this is something we should discuss! (I'll add it to the agenda of the next XR team meeting.)

If we don't want to go this way, there are a number of possible alternative options:

- We could release _both_ Meta enabled and disabled builds
- We could build both with and without the headers (to test that both can compile), but only release the version without the headers here
- We could build and release the Meta enabled version elsewhere (maybe in a project under [godot-sdk-integrations](https://github.com/godot-sdk-integrations)?)

~~**UPDATE (2025-07-10):** The latest version builds _both_ Meta and non-Meta builds, including for release, but not for MavenCentral~~

**UPDATE (2025-07-11):** The latest version builds both Meta and non-Meta builds, but only releases the Meta ones, including to MavenCentral

### DIFF
--- a/.github/workflows/build-addon-on-push.yml
+++ b/.github/workflows/build-addon-on-push.yml
@@ -7,6 +7,8 @@ env:
   # Only used for the cache key. Increment version to force clean build.
   GODOT_BASE_BRANCH: master
   SCONS_CACHE: ${{ github.workspace }}/.scons-cache/
+  # This is ovr_openxr_mobile_sdk v77.
+  META_OPENXR_HEADERS_URL: 'https://securecdn-ord5-3.oculus.com/binaries/download/?id=23875289988749597'
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
Feedback from XR Meeting:
- we are all in agreement we want this
- would like confirmation from Meta that this type of distribution does not violate their license.

BastiaanOlij (approved):
One small change but otherwise this is good to go, discussing today that we want the version with meta headers to be published.

m4gr3d (approved):

PR #318, opened 2025-06-27
Configure all samples to use REC709 color space

dsnopek:
This builds on https://github.com/GodotVR/godot_openxr_vendors/pull/305 by setting all our samples to use the recommended color space of REC709


devloglogan (approved):
Should have done this in that PR; thanks!

PR #322, opened 2025-07-10
Only display color space warnings when runtime default is set

devloglogan:
Per the discussion in #307, this would make the warnings less intrusive, as currently if a user wants something other than Rec 709 they will _always_ receive warnings.


dsnopek (approved):
LGMT!

PR #327, opened 2025-07-23
Use stable version for 4.1.0 version

insanepure:
This should fix an issue where it is trying to find 4.1.0-dev on maven but there is none.

### DIFF
--- a/config.gradle
+++ b/config.gradle
@@ -21,7 +21,7 @@ ext {
 
     // Parse the release version from the gradle project properties (e.g: -Prelease_version=<version>)
     getReleaseVersion = { ->
-        final String defaultVersion = "4.1.0-dev" // Also update 'plugin/src/main/cpp/include/export/export_plugin.h#PLUGIN_VERSION'
+        final String defaultVersion = "4.1.0-stable" // Also update 'plugin/src/main/cpp/include/export/export_plugin.h#PLUGIN_VERSION'
 
         String releaseVersion = project.hasProperty("release_version") ? project.property("release_version") : defaultVersion
         if (releaseVersion == null || releaseVersion.isEmpty()) {
--- a/plugin/src/main/cpp/include/export/export_plugin.h
+++ b/plugin/src/main/cpp/include/export/export_plugin.h
@@ -40,7 +40,7 @@
 
 using namespace godot;
 
-static const char *PLUGIN_VERSION = "4.1.0-dev"; // Also update 'config.gradle#defaultVersion'
+static const char *PLUGIN_VERSION = "4.1.0-stable"; // Also update 'config.gradle#defaultVersion'
 
 // Set of supported vendors
 static const char *META_VENDOR_NAME = "meta";


config.gradle
```diff
@@ -21,7 +21,7 @@ ext {
 
     // Parse the release version from the gradle project properties (e.g: -Prelease_version=<version>)
     getReleaseVersion = { ->
-        final String defaultVersion = "4.1.0-dev" // Also update 'plugin/src/main/cpp/include/export/export_plugin.h#PLUGIN_VERSION'
+        final String defaultVersion = "4.1.0-stable" // Also update 'plugin/src/main/cpp/include/export/export_plugin.h#PLUGIN_VERSION'
```
dsnopek, line 24: Hm. When we actually push to MavenCentral from CI, we override this with `-Prelease_version=...`, so it should be pushed with `4.1.0-stable` like the GitHub tag.

I'm not sure what else this affects?

insanepure: Ah sorry, I posted this in the Discord.

I got this error with my github action that exports the project for all headsets based on the godot xr tools.

What went wrong:
Execution failed for task ':checkStandardDebugAarMetadata'.
Could not resolve all files for configuration ':standardDebugRuntimeClasspath'.
Could not find org.godotengine:godot-openxr-vendors-meta:4.1.0-dev.
     Searched in the following locations:
       
https://dl.google.com/dl/android/maven2/org/godotengine/godot-openxr-vendors-meta/4.1.0-dev/godot-openxr-vendors-meta-4.1.0-dev.pom
https://repo.maven.apache.org/maven2/org/godotengine/godot-openxr-vendors-meta/4.1.0-dev/godot-openxr-vendors-meta-4.1.0-dev.pom
https://plugins.gradle.org/m2/org/godotengine/godot-openxr-vendors-meta/4.1.0-dev/godot-openxr-vendors-meta-4.1.0-dev.pom
https://s01.oss.sonatype.org/content/repositories/snapshots/org/godotengine/godot-openxr-vendors-meta/4.1.0-dev/godot-openxr-vendors-meta-4.1.0-dev.pom
Required by:
   project :

Same error for pico (there it is godot-openxr-vendors-pico:4.1.0-dev)

insanepure: This happend after upgrading to 4.1.0 from 4.0.0.
And looking at the code, there used to be -stable, now it is only -dev so it matches the error message.

m4gr3d: @dsnopek The issue at play is that we forgot to update https://github.com/GodotVR/godot_openxr_vendors/blob/b7c312e37230ded260eb3a8bff4c74932765621b/plugin/src/main/cpp/include/export/export_plugin.h#L43 to the same version as the one we're pushing to maven central.

When the `aar` files are not available, the export plugin defaults to the maven central dependency based on https://github.com/GodotVR/godot_openxr_vendors/blob/b7c312e37230ded260eb3a8bff4c74932765621b/plugin/src/main/cpp/export/export_plugin.cpp#L75.

So in the current 4.1.0-stable release, the returned maven dependency is `org.godotengine:godot-openxr-vendors-<vendor>:4.1.0-dev` because `PLUGIN_VERSION` is set to `4.1.0-dev`.

In addition to the changes in this PR, we need to generate and push a 4.1.1 release to replace the current maven central release.
I'll create the PR for it.

dsnopek: Hm, ok. We need to document the release steps and/or setup some automation so can't forget them

We have the `-Prelease_version="..."` thing on CI, but maybe rather than doing that, it should confirm that the Git tag matches what we have in the code, and refuse to do the push if it doesn't match

PR #328, opened 2025-07-23
Implement XR_ML_marker_understanding

zenorbi:
This PR adds support for the XR_ML_marker_understanding extension. I also added a sample project.

Please note, that this is my first time writing C++ and Godot extension. I am mainly a C and C# developer, so please double check my memory management.


BastiaanOlij:
This is very cool, thanks for submitting it.

It will be hard for us to verify, hopefully there are other Magicleap users who can take this for a spin and report back on this.

From an implementation POV I would like you to have a look at [the Meta scene discovery implementation](https://godotvr.github.io/godot_openxr_vendors/manual/meta/scene_manager.html) and the upcoming [spatial entities implementation](https://github.com/godotengine/godot/pull/107391).

Both implementations are centred around XRTracker objects that are consumed by a manager script. This will hopefully limit the amount of work developers need to do to create a single implementation that targets different vendor/core implementations.

In essence, any marker that is found should be registered with the XRServer as an XRTracker (instance of XRPositionalTracker or a subclass thereof, with the tracker type set to anchor). This is a pure data class that tracks the location of the marker and which can store meta data such as the marker code.

On the Godot side a manager object reacts to new trackers being added or removed from the XRServer and adding child scenes as a result. These can simply have an XRNode3D as root with the tracker assigned. They will automatically update their positions as the data object is updated, and further implementation can react to other data held within the tracker object.

You can read the [draft of the spatial entities manual entry](https://github.com/godotengine/godot/pull/107391) for more clarity on this.

BastiaanOlij:
### XRTracker

The XRTrackers indeed need to have a unique name as they are reference by the name, but it can be anything. You can even generate a UID just within Godot or even go as simple as having just a counter that increments  by 1 every time a new marker is found and prefix it with something (e.g. `ml_marker/1`). There is no expectation the name survives over sessions.

This will also allow people to use the same marker multiple times, say you're making an AR card game, the marker code would indicate which card is being put on the table, a deck may have the same card multiple times.

The data is then simply exposed through a new method of property in your marker subclass.

### Manager

As for the manager, its just a manager that creates and removes the nodes that represent the trackers. So a new marker is detected, a new XRTracker is created, the manager reacts on the new tracker signal, and creates a new XRNode3D for that tracker.

You are continuously updating the position of the marker by updating the pose on the tracker, and that is all handled by the XRNode3D which reacts to position updates.

If the marker is removed when it's no longer in view, and recreated later, the manager simply removes the node, and adds a new node back in, probably with a new marker id. There is no way for us to know whether the marker is still the same one, probably Magic Leap can't know for sure either.

zenorbi:
Changed the implementation to inherit from `XRPositionalTracker` and add these to the `XRServer`. I am not completely sure how the manager should work. Should it only watch a specific marker detector, or should it watch the `XRServer` and create a node for every `/ml_marker/*` tracker, independent of which detector created that?

BastiaanOlij:
> Changed the implementation to inherit from `XRPositionalTracker` and add these to the `XRServer`. I am not completely sure how the manager should work. Should it only watch a specific marker detector, or should it watch the `XRServer` and create a node for every `/ml_marker/*` tracker, independent of which detector created that?

It's a shame that the spatial entities documentation isn't merged yet because it has a good example of a manager implementation (which you do in GDScript). But yeah, you basically connect to the [tracker_added](https://docs.godotengine.org/en/stable/classes/class_xrserver.html#class-xrserver-signal-tracker-added) and [tracker_removed](https://docs.godotengine.org/en/stable/classes/class_xrserver.html#class-xrserver-signal-tracker-removed) signals to see when a new ML marker becomes available or disappears.

The manager than instances a scene with an `XRNode3D` node at its root specific for ML markers, (which you can detect by their class being `OpenXRMlMarkerTracker`).

If we do this consistently over all the different implementations, that means the user can implement a single manager that can handle different tracker types so that you can create a cross platform solution.

BastiaanOlij:
Just for reference, this is the manager in the current spatial entities demo (which I haven't made public yet):
```
## Spatial entity manager handles adding subscenes as new spatial entities are discovered.
## Right now this is implemented on the assumption that we automatically discover entities
## in our OpenXR implementation. We may change way from this in which case we'll add info
## about our query parameters here and select nodes based on that.
class_name SpatialEntitiesManager
extends Node3D

#region Signals
## Signals a new spatial entity node was added.
signal added_spatial_entity(node : XRNode3D)

## Signals a spatial entity node is about to be removed.
signal removed_spatial_entity(node : XRNode3D)
#endregion

#region Export variables
## Scene to instantiate for spatial anchor entities.
@export var spatial_anchor_scene : PackedScene

## Scene to instantiate for plane tracking spatial entities.
@export var plane_tracker_scene : PackedScene

## Scene to instantiate for mark tracking spatial entities.
@export var marker_tracker_scene : PackedScene
#endregion

#region Private variables
# Our current active spacial entities manager (last one instantiated)
static var _current_manager : SpatialEntitiesManager

# Trackers we manage nodes for
var _managed_nodes : Dictionary[XRTracker, XRAnchor3D]
#endregion

#region Public methods
## Create a new spacial anchor with the associated child scene
## If persistent anchors are supported, this will be created as a persistent node
## and we will store the child scene path with the anchors UUID for future recreation.
static func create_spacial_anchor(p_transform : Transform3D, p_child_scene_path : String, p_new_anchor : bool = true) -> void:
	# Do we have anchor support?
	if not OpenXRSpatialAnchorCapability.is_spatial_anchor_supported():
		push_error("Spatial anchors are not supported or enabled on this device!")
		return

	if not _current_manager:
		push_error("No current spatial entities manager set!")
		return

	# Adjust our transform to local space
	var t : Transform3D = _current_manager.global_transform.inverse() * p_transform

	# Create anchor on our current manager.
	var new_anchor = OpenXRSpatialAnchorCapability.create_new_anchor(t, RID())
	if not new_anchor:
		push_error("Couldn't create an anchor for %s." % [ p_child_scene_path ])
		return

	# Creating a new anchor should have resulted in an XRAnchor being added to the scene
	var anchor_scene = _current_manager.get_tracked_scene(new_anchor)
	if not anchor_scene:
		push_error("Couldn't locate anchor scene for %s, has the manager been configured with an applicable anchor scene?" % [ new_anchor.name ])
		return
	if not anchor_scene is OpenXRSpatialAnchor3D:
		push_error("Anchor scene for %s is not a OpenXRSpatialAnchor3D scene, has the manager been configured with an applicable anchor scene?" % [ new_anchor.name ])
		return

	anchor_scene.set_child_scene(p_child_scene_path, p_new_anchor)


## Removes this spatial anchor from our scene.
## If the spatial anchor is persistant, the associated UUID will be cleared.
static func remove_spacial_anchor(p_anchor : XRAnchor3D) -> void:
	# Do we have anchor support?
	if not OpenXRSpatialAnchorCapability.is_spatial_anchor_supported():
		push_error("Spatial anchors are not supported on this device!")
		return

	var tracker : XRTracker = XRServer.get_tracker(p_anchor.tracker)
	if tracker and tracker is OpenXRAnchorTracker:
		var anchor_tracker : OpenXRAnchorTracker = tracker
		if anchor_tracker.has_uuid() and OpenXRSpatialAnchorCapability.is_spatial_persistence_supported():
			# If we have a UUID we should first make the anchor unpersistent
			# and then remove it on its callback.
			if SpatialUuidDb:
				SpatialUuidDb.remove_uuid(anchor_tracker.uuid)

			var future_result = OpenXRSpatialAnchorCapability.make_anchor_unpersistent(anchor_tracker)
			var success : bool = await future_result.completed

			if success:
				# Our tracker is now no longer persistent, we can remove it.
				OpenXRSpatialAnchorCapability.remove_anchor(anchor_tracker)
		else:
			# Else we can just remove it.
			# This will remove it from the XRServer, which in turn will trigger cleaning up our node.
			OpenXRSpatialAnchorCapability.remove_anchor(tracker)

## Retrieve the scene we've added for a given tracker (if any).
func get_tracked_scene(p_tracker : XRTracker) -> XRNode3D:
	for node in get_children():
		if node is XRNode3D and node.tracker == p_tracker.name:
			return node

	return null
#endregion


#region Build in private functions
# Enter tree is called whenever our node is added into our scene.
func _enter_tree():
	# This is now our current manager
	_current_manager = self

	# Connect to signals that inform us about tracker changes.
	XRServer.tracker_added.connect(_on_tracker_added)
	XRServer.tracker_updated.connect(_on_tracker_updated)
	XRServer.tracker_removed.connect(_on_tracker_removed)

	# Setup existing trackers
	var trackers : Dictionary = XRServer.get_trackers(XRServer.TRACKER_ANCHOR)
	for tracker_name in trackers:
		var tracker : XRTracker = trackers[tracker_name]
		if tracker:
			_add_tracker(tracker)

# Exit tree is called whenever our node is removed from out scene.
func _exit_tree():
	# If we are the current manager, bye!
	if _current_manager == self:
		_current_manager = null

	# Clean up our signals.
	XRServer.tracker_added.disconnect(_on_tracker_added)
	XRServer.tracker_updated.disconnect(_on_tracker_updated)
	XRServer.tracker_removed.disconnect(_on_tracker_removed)

	# Clean up
	for tracker in _managed_nodes:
		removed_spatial_entity.emit(_managed_nodes[tracker])
		remove_child(_managed_nodes[tracker])
		_managed_nodes[tracker].queue_free()
	_managed_nodes.clear()
#endregion

#region Private methods
# See if this tracker should be managed by us and add it
func _add_tracker(tracker : XRTracker):
	var new_node : XRAnchor3D

	if _managed_nodes.has(tracker):
		# Already being managed by us!
		return

	if tracker is OpenXRAnchorTracker:
		# Note, generally spatial anchors are controlled by the developer and
		# are unlikely to be handled by our manager.
		# But just for completion we'll add it in.
		if spatial_anchor_scene:
			var new_scene = spatial_anchor_scene.instantiate()
			if new_scene is XRAnchor3D:
				new_node = new_scene
			else:
				push_error("Spatial anchor scene doesn't have an XRAnchor3D as a root node and can't be used!")
				new_scene.free()
	elif tracker is OpenXRPlaneTracker:
		if plane_tracker_scene:
			var new_scene = plane_tracker_scene.instantiate()
			if new_scene is XRAnchor3D:
				new_node = new_scene
			else:
				push_error("Plane tracking scene doesn't have an XRAnchor3D as a root node and can't be used!")
				new_scene.free()
	elif tracker is OpenXRMarkerTracker:
		if marker_tracker_scene:
			var new_scene = marker_tracker_scene.instantiate()
			if new_scene is XRAnchor3D:
				new_node = new_scene
			else:
				push_error("Marker tracking scene doesn't have an XRAnchor3D as a root node and can't be used!")
				new_scene.free()
	elif tracker is OpenXRSpatialEntityTracker:
		# Type of spatial entity tracker we're not supporting?
		push_warning("OpenXR Spatial Entities: Unsupported anchor tracker " + tracker.get_name() + " of type " + tracker.get_class())
	else:
		# Not a type managed by us!
		return

	if not new_node:
		# No scene defined or able to be instantiated? We're done!
		return

	# Setup and add to our scene.
	new_node.tracker = tracker.name
	new_node.pose = "default"
	_managed_nodes[tracker] = new_node
	add_child(new_node)

	added_spatial_entity.emit(new_node)
#endregion

#region Signal handling
# A new tracker was added to our XRServer.
func _on_tracker_added(tracker_name: StringName, type: int):
	print("Added tracker ", tracker_name)
	if type == XRServer.TRACKER_ANCHOR:
		var tracker : XRTracker = XRServer.get_tracker(tracker_name)
		if tracker:
			_add_tracker(tracker)


# A tracked managed by XRServer was changed.
func _on_tracker_updated(_tracker_name: StringName, _type: int):
	# For now we ignore this, there aren't changes here we need to react
	# to and the instanced scene can react to this itself if needed.
	pass


# A tracker was removed from our XRServer.
func _on_tracker_removed(tracker_name: StringName, type: int):
	if type == XRServer.TRACKER_ANCHOR:
		var tracker : XRTracker = XRServer.get_tracker(tracker_name)
		if _managed_nodes.has(tracker):
			# We emit this right before we remove it!
			removed_spatial_entity.emit(_managed_nodes[tracker])

			# Remove the node.
			remove_child(_managed_nodes[tracker])

			# Queue free the node.
			_managed_nodes[tracker].queue_free()

			# And remove from our managed nodes.
			_managed_nodes.erase(tracker)
#endregion
```

The magic happens in the `_add_tracker` method.

BastiaanOlij:
Looking further at the implementation, from as far as I can tell from the code, I do think you got it all pretty much as it should be. 

One thing that isn't clear to me is how `OpenXRMlMarkerDetector` is supposed to work. I'm assuming you somehow instance this together with a configuration to ensure markers are  being tracked but it's a RefCounted object.

Are we supposed to just hang on to it in a variable?

My gut feeling says this should be a node so that you can configure it within the IDE and its lifecycle is clear where the detector starts working when a scene is loaded and is cleaned up when unloaded.

BastiaanOlij:
It's hard to confirm without hardware to test on, but purely from reading from the code and what we've discussed so far, this feels good.

zenorbi:
Rebased the project and fixed the `get_estimate_*_length` behaving inversely compared to what was expected. `true` now means the platform should estimate the length, `false` means the length is provided.

m4gr3d (approved):
The code looks good to me!

It's a pity we can't test and validate the implementation, but I'd be in favor of merging it as is.

@dsnopek Are we able to use the experimental tag on GDExtension APIs? If yes, I'd suggest we mark this implementation as experimental until it's tested and validated.

dsnopek:
> @dsnopek Are we able to use the experimental tag on GDExtension APIs? If yes, I'd suggest we mark this implementation as experimental until it's tested and validated.

Yes, this is just an attribute in the XML for the class docs. It doesn't looks like this PR has generated the XML docs? But if it did, it's just adding `experimental="Optional note here"` in all the `<class ...>` tags

PR #334, opened 2025-08-06
Implement XR_META_simultaneous_hands_and_controllers extension

devloglogan:
Adds implementation of the [`XR_META_simultaneous_hands_and_controllers`](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_META_simultaneous_hands_and_controllers) extension. Really simple extension wrapper. Just has a function to check if the feature is supported, and functions to resume/pause it.

Right now the only thing of note in testing is that upon pausing simultaneous tracking, a few of the following errors print out:
```
hand_controller.gd:33 @ _process(): Condition "!v.is_finite()" is true.
```

The errors are printed from `RenderSceneCull::instance_set_transform()`, not sure why at the moment (will try to look into more soon). I know I see this error printed elsewhere from time to time, I believe related to render models or hand tracking meshes. So, I don't think this extension does anything wrong and could still be merged. Perhaps an issue should be made to track this `!v.s_finite()` error? I'll possibly do that after looking into it a bit more.


m4gr3d (approved):
Looks good!

PR #339, opened 2025-08-12
Fix `XR_META_environment_depth` when rendering with "Separate" thread model

dsnopek:
Currently, with the meta-scene-sample project, if you change the rendering thread model to "Separate", it will crash!

This PR fixes the issues that made it unsafe.

For the most part, this is just about moving stuff to the render thread


BastiaanOlij:
This looks good to me but hard to test. My only remark would be that in our setup, where we're initializing our render state even though we're not on the rendering thread, we should probably have a remark that this is safe. Though that might be overkill.

I/We should find some time to properly document what we're doing both in core and here.
Seeing we have a bit of a weird relationship with the rendering server and rendering thread, it's not always easy to tell why we're doing this weirdness and what is allowed and what leads to issues.

PR #340, opened 2025-08-14
Add XR_META_hand_tracking_microgestures to hand tracking sample

devloglogan:
This PR relies on https://github.com/godotengine/godot/pull/109611 and will remain as a draft until that PR is merged.

The above mentioned PR enables hand tracking microgestures, adding some more actions to the hand interaction profile. This PR adds that functionality to the hand tracking sample!

~~Hand aim and microgestures require different trackers to be selected, so in the sample you're able to swap between the two. I'm now realizing you could probably just add additional controllers set to these other trackers though, and then there would be no need for switching. I'll probably update this PR to do that instead, but in the meantime it still works for testing.~~
**Edit 2025-08-15**: There is no tracker swapping now, I've just added extra nodes for the additional trackers so it all can be viewed at once.


dsnopek:
Thanks! This PR is working great in my testing

However, I agree with this statement:

> Hand aim and microgestures require different trackers to be selected, so in the sample you're able to swap between the two. I'm now realizing you could probably just add additional controllers set to these other trackers though, and then there would be no need for switching.

I think this would be better with the 2nd set of controllers always being available and listening for input, so there's no need to "swap" between them. I don't think the "swapping" reflects what a game developer would want to do in a real game

dsnopek (approved):
Looks great!

Of course, this'll need to wait until the Godot PR is merged, _and_ we've decided to start targeting Godot 4.6+

m4gr3d (approved):

dsnopek (approved):
Looks great to me! I just re-tested on Meta Quest 3, and the microgestures in the sample worked fine :-)

PR #342, opened 2025-08-19
Fix Meta passthrough when rendering with "Separate" thread model

dsnopek:
Similar to https://github.com/GodotVR/godot_openxr_vendors/pull/339 this moves a bunch of stuff to the render thread to make it fully safe to use with the "Separate" thread model

However, unlike that other PR, I don't have a way to reliably cause a crash without the PR.

In fact, the old code is _mostly_ safe for enabling passthrough initially. However, there's theoretically issues with changing passthrough settings mid-way, destroying `OpenXRFbPassthroughGeometry` and `OpenXRMetaPassthroughColorLut`, or when leaving passthrough or quitting the app. This PR should fix all those theoretical issues, despite not having a reliable way to trigger any crashes due to them.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
@@ -34,6 +34,7 @@
 #include <godot_cpp/classes/main_loop.hpp>
 #include <godot_cpp/classes/open_xrapi_extension.hpp>
 #include <godot_cpp/classes/os.hpp>
+#include <godot_cpp/classes/rendering_server.hpp>
 #include <godot_cpp/classes/scene_tree.hpp>
 #include <godot_cpp/classes/viewport.hpp>
 #include <godot_cpp/classes/xr_server.hpp>
@@ -196,10 +197,11 @@ void OpenXRFbPassthroughExtensionWrapper::_on_session_created(uint64_t p_session
 			0, // flags
 		};
 
-		XrResult result = xrCreatePassthroughFB(session, &passthrough_create_info, &passthrough_handle);
+		// Since OpenXR isn't yet involved in rendering yet, it's safe to write to `render_state`.
+		XrResult result = xrCreatePassthroughFB(session, &passthrough_create_info, &render_state.passthrough_handle);
 		if (XR_FAILED(result)) {
 			UtilityFunctions::print("Failed to create passthrough");
-			passthrough_handle = XR_NULL_HANDLE;
+			render_state.passthrough_handle = XR_NULL_HANDLE;
 			return;
 		}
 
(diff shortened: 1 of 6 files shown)


plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -265,15 +267,16 @@ bool OpenXRFbPassthroughExtensionWrapper::_on_event_polled(const void *p_event)
 
 void OpenXRFbPassthroughExtensionWrapper::_on_session_destroyed() {
 	if (fb_passthrough_ext) {
-		stop_passthrough();
+		// Since OpenXR isn't is no longer involved in rendering, it's safe to write to `render_state`.
```
devloglogan:
```suggestion
		// Since OpenXR is no longer involved in rendering, it's safe to write to `render_state`.
```

plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
```diff
@@ -321,71 +328,79 @@ void OpenXRFbPassthroughExtensionWrapper::start_passthrough_layer(LayerPurpose p
...
-			emit_signal("openxr_fb_projected_passthrough_layer_created");
+			// Use `call_deferred()` so the signal will be emitted on the main thread.
+			callable_mp(this, &OpenXRFbPassthroughExtensionWrapper::_emit_signal).bind("openxr_fb_projected_passthrough_layer_created").call_deferred();
 		}
 	} else { // Else resume already created layer.
-		XrResult result = xrPassthroughLayerResumeFB(passthrough_layer[p_layer_purpose]);
+		XrResult result = xrPassthroughLayerResumeFB(render_state.passthrough_layer[p_layer_purpose]);
 		if (XR_FAILED(result)) {
 			UtilityFunctions::print("Failed to resume passthrough layer ", p_layer_purpose, ", error code: ", result);
+			callable_mp(this, &OpenXRFbPassthroughExtensionWrapper::_set_current_passthrough_layer).bind(LAYER_PURPOSE_NONE).call_deferred();
```
devloglogan, line 383: Seems like every time you use `call_deferred()` you drop a comment above it's use, so I'd put one here as well!

devloglogan (approved):
Tested both thread models with the mobile renderer and seems to be working well! Code LGTM, just some little comment nits.

m4gr3d (approved):

plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
dsnopek: Thanks! Fixed in my latest push

plugin/src/main/cpp/extensions/openxr_fb_passthrough_extension_wrapper.cpp
dsnopek: Thanks! Added in my latest push

PR #344, opened 2025-08-25
Fix errors reported by `XrApiLayer_core_validation`

dsnopek:
This fixes a couple of minor bug fixes based on errors from running all of the sample projects with the OpenXR `XrApiLayer_core_validation` layer.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
@@ -78,6 +78,11 @@ uint64_t OpenXRFbCompositionLayerSecureContentExtensionWrapper::_set_viewport_co
 		return reinterpret_cast<uint64_t>(p_next_pointer);
 	}
 
+	ExternalOutput external_output = (ExternalOutput)(int)p_property_values.get(EXTERNAL_OUTPUT_PROPERTY_NAME, EXTERNAL_OUTPUT_DISPLAY);
+	if (external_output == EXTERNAL_OUTPUT_DISPLAY) {
+		return reinterpret_cast<uint64_t>(p_next_pointer);
+	}
+
 	const XrCompositionLayerBaseHeader *layer = reinterpret_cast<const XrCompositionLayerBaseHeader *>(p_layer);
 
 	if (!layer_structs.has(layer)) {
@@ -90,9 +95,9 @@ uint64_t OpenXRFbCompositionLayerSecureContentExtensionWrapper::_set_viewport_co
 
 	XrCompositionLayerSecureContentFB *secure_content = layer_structs.getptr(layer);
 
-	switch ((ExternalOutput)(int)p_property_values.get(EXTERNAL_OUTPUT_PROPERTY_NAME, EXTERNAL_OUTPUT_DISPLAY)) {
+	switch (external_output) {
 		case EXTERNAL_OUTPUT_DISPLAY: {
-			secure_content->flags = 0;
+			// We'll never reach this - it would have been handled above.
 		} break;
 		case EXTERNAL_OUTPUT_EXCLUDE: {
 			secure_content->flags = XR_COMPOSITION_LAYER_SECURE_CONTENT_EXCLUDE_LAYER_BIT_FB;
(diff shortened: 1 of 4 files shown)


plugin/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
```diff
@@ -92,7 +97,7 @@ uint64_t OpenXRFbCompositionLayerSecureContentExtensionWrapper::_set_viewport_co
 
 	switch ((ExternalOutput)(int)p_property_values.get(EXTERNAL_OUTPUT_PROPERTY_NAME, EXTERNAL_OUTPUT_DISPLAY)) {
```
m4gr3d: Can you reuse the `external_output` variable from above.

plugin/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
```diff
@@ -92,7 +97,7 @@ uint64_t OpenXRFbCompositionLayerSecureContentExtensionWrapper::_set_viewport_co
...
 		case EXTERNAL_OUTPUT_DISPLAY: {
-			secure_content->flags = 0;
+			// Skip.
```
m4gr3d: I'd mention something to the effect that we should not reach this code since it should have already been handled above.

plugin/src/main/cpp/extensions/openxr_fb_spatial_entity_query_extension_wrapper.cpp
```diff
@@ -146,14 +146,13 @@ void OpenXRFbSpatialEntityQueryExtensionWrapper::on_space_query_results(const Xr
 
 	QueryInfo *query = queries.getptr(event->requestId);
 
-	Vector<XrSpaceQueryResultFB> results;
 	query->results.resize(queryResults.resultCountOutput);
-	queryResults.resultCapacityInput = query->results.size();
-	queryResults.resultCountOutput = 0;
+	queryResults.resultCapacityInput = queryResults.resultCountOutput;
```
m4gr3d, line 150: Should we check the value of `resultCountOutput` here and return early if it's `0` as done in https://github.com/GodotVR/godot_openxr_vendors/pull/341?

dsnopek: Hm, logically, that makes sense. However, the spec has this:

> This function must only retrieve each query result once. After the application has used this function to retrieve a query result, the runtime frees its copy.

If we return early and don't do the second call to attempt to get the result data, will the runtime ever free this query result?

If the query id wasn't valid, we'd be getting `XR_ERROR_VALIDATION_FAILURE` on this 2nd call, and we don't get that error if the query returns zero results (at least as far as I remember). So, that leads me to believe that the runtime is keeping the query data alive until we do the 2nd call.

Here's the relevant part of the spec: https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#xrRetrieveSpaceQueryResultsFB

plugin/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
dsnopek: Ah, yes, we can! Updated in my latest push

plugin/src/main/cpp/extensions/openxr_fb_composition_layer_secure_content_extension_wrapper.cpp
dsnopek: Good idea! Comment updated in my latest push

plugin/src/main/cpp/extensions/openxr_fb_spatial_entity_query_extension_wrapper.cpp
dsnopek: I just tested this by purposefully making a query that would return zero results:

```gdscript
	var query := OpenXRFbSpatialEntityQuery.new()
	query.openxr_fb_spatial_entity_query_completed.connect(func(x): print("results: ", x))
	# A totally made up UUID.
	query.query_by_uuid(['aa988e62-96c5-4328-a66a-72e3692085eb'], OpenXRFbSpatialEntity.STORAGE_LOCAL)
	query.execute()
```

It does _not_ return `XR_ERROR_VALIDATION_FAILURE` in the 2nd call, which means the result is still valid at that point, so I can reasonably suspect that if we didn't make the 2nd call (even if it's got `0` results) that the runtime wouldn't free the result.

So, I think the code I already have in this PR is correct.

PR #345, opened 2025-08-26
Implement XR_META_performance_metrics extension

devloglogan:
Adds an extension wrapper implementation for [XR_META_performance_metrics](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_META_performance_metrics).

Performance metrics can be enabled via the `set_performance_metrics_enabled` function assuming the extension itself has been enabled in project settings. Once enabled the available metrics to measure can be obtained via the `get_performance_metrics_counter_paths` function, and those can be individually queried to receive a `Dictionary` of data with the `query_performance_metrics_counter` function.

The query function can be used with the `Performance` singleton to add custom monitors for debugging (screenshot below), I will in the near future add a doc page showing users how to do this. 🙂

Clipboard - August 26, 2025 12_34 PM


devloglogan:
After chatting with @dsnopek, having the extension wrapper automatically add the metrics as custom monitors for the user seemed like the correct way to go for this extension. When enabling the performance metrics system, that now happens! There's also another project setting (`performance_metrics_enable_on_startup`) that will make it so users don't have to enable the performance metrics system by hand.

Here's what the custom monitors look like now testing with Quest 3:
image

plugin/src/main/cpp/register_types.cpp
```diff
@@ -441,6 +448,8 @@ void add_plugin_project_settings() {
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/composition_layer_settings", true);
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/dynamic_resolution", true);
 	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/headset_id", false);
+	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/performance_metrics", false);
+	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/performance_metrics_enable_on_startup", false);
```
dsnopek: Should the "enable on startup" option be `true` by default? If we think this is the most common way that the extension will be used, it seems like it should be? It won't do anything unless the extension itself is enabled, but this would mean one click instead of two

Also, I'm starting to wonder if we shouldn't use a different word than "enable" for this since it is confusing with the idea of enabling the OpenXR module? Maybe the word should be "capture" and so "capture on startup"? And perhaps this should be a suboption, similar to the color space and scene API options?

So, something like:

```suggestion
	_add_bool_project_setting(project_settings, "xr/openxr/extensions/meta/performance_metrics/capture_on_startup", true);
```

plugin/src/main/cpp/include/extensions/openxr_meta_performance_metrics_extension_wrapper.h
```diff
@@ -0,0 +1,137 @@
...
+	godot::Dictionary _get_requested_extensions() override;
+
+	void _on_instance_created(uint64_t instance) override;
+	void _on_instance_destroyed() override;
+	void _on_state_ready() override;
+
+	bool is_enabled() { return meta_performance_metrics_ext; }
+
+	void set_performance_metrics_enabled(bool p_enabled);
+	bool is_performance_metrics_enabled();
```
dsnopek: Again, maybe we could use the word "capture" rather than the more generic "enable"?

Maybe:

```suggestion
	void set_capture_performance_metrics(bool p_enabled);
	bool is_capturing_performance_metrics();
```

plugin/src/main/cpp/register_types.cpp
devloglogan: I agree with these suggestions, and I've applied them in the latest push. 🙂

plugin/src/main/cpp/include/extensions/openxr_meta_performance_metrics_extension_wrapper.h
devloglogan: Applied!

doc_classes/OpenXRMetaPerformanceMetricsExtensionWrapper.xml
```diff
@@ -0,0 +1,76 @@
...
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the extension is enabled; otherwise [code]false[/code].
+			</description>
+		</method>
+		<method name="query_performance_metrics_counter">
+			<return type="Dictionary" />
+			<param index="0" name="counter_path" type="String" />
+			<description>
+				Returns a [Dictionary] containing the following keys:
```
dsnopek: This can also return an empty dictionary. It should probably explain the conditions that can lead to that

devloglogan: Added a line detailing this!

devloglogan:
Pushed some changes updating the implementation to a more vendor-neutral API.

`OpenXRVendorPerformanceMetrics` is now the singleton that exposes the exact same API as before, it just requires an `OpenXRVendorPerformanceMetricsProvider` to be set (which `OpenXRMetaPerformanceMetrics` now inherits from).

On `OpenXRVendorPerformanceMetrics` I've also added a `capture_performance_metrics` property which can be used instead of the setter/getter.

doc_classes/OpenXRVendorPerformanceMetrics.xml
```diff
@@ -0,0 +1,69 @@
...
+		<method name="get_performance_metrics_counter_paths">
+			<return type="PackedStringArray" />
+			<description>
+				Returns a [code]PackedStringArray[/code] of counter paths that may be passed to [method query_performance_metrics_counter].
+			</description>
+		</method>
+		<method name="is_enabled">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if the extension is enabled; otherwise [code]false[/code].
```
m4gr3d:
```suggestion
				Returns [code]true[/code] if the singleton is enabled; otherwise [code]false[/code].
```

plugin/src/main/cpp/extensions/openxr_meta_performance_metrics_extension_wrapper.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+	performance_metrics_counter_xr_paths.resize(num_counter_paths);
+	result = xrEnumeratePerformanceMetricsCounterPathsMETA(xr_instance, num_counter_paths, &num_counter_paths, performance_metrics_counter_xr_paths.ptrw());
+	if (XR_FAILED(result)) {
+		UtilityFunctions::print_verbose(vformat("Failed to get performance metrics counter paths [%s]", get_openxr_api()->get_error_string(result)));
+		performance_metrics_counter_xr_paths.clear();
+		return;
+	}
+
+	performance_metrics_counter_paths.resize(num_counter_paths);
```
m4gr3d, line 165: We should check if `num_counter_paths` is greater than 0 and return early if not.

plugin/src/main/cpp/include/classes/openxr_vendor_performance_metrics.h
```diff
@@ -0,0 +1,88 @@
...
+	};
+
+	OpenXRVendorPerformanceMetrics();
+	~OpenXRVendorPerformanceMetrics();
+
+	static OpenXRVendorPerformanceMetrics *get_singleton();
+
+	bool is_enabled();
+
+	void set_vendor_performance_metrics_provider(OpenXRVendorPerformanceMetricsProvider *p_provider);
```
m4gr3d, line 64: Should this and the getter be friend functions only exposed to class extending `OpenXRVendorPerformanceMetricsProvider`?

plugin/src/main/cpp/extensions/openxr_meta_performance_metrics_extension_wrapper.cpp
devloglogan: Done!

plugin/src/main/cpp/include/classes/openxr_vendor_performance_metrics.h
devloglogan: Could you expand on this a little bit? I understand friend classes, but looking up `friend` in use with functions, I'm not sure I see the advantage. From my understanding it seems like we could make these functions private, but then we'd have to make public functions (that would be the friend functions) in `OpenXRVendorPerformanceMetricsProvider`, which feels like unnecessary complexity. Perhaps I misunderstand though!

m4gr3d: Nevermind; I forgot that unless we expose these methods via `_bind_methods`, they're not public to the `gdscript` code.

PR #353, opened 2025-09-03
Allow embedding OpenXR validation layers on Android

dsnopek:
This PR will allow building and including the OpenXR validation layers (the core validation and best practices ones) into Android builds and automatically enable them.

**UPDATE (2025-12-11):** This now builds the validation layers into its own AAR file, which can then be included via an export option. It no longer has any issues with OpenGL, and godot_openxr_vendors has been updated for the earlier validation issues, but you still need to use it with Godot 4.6 which includes the fixes for the validation issues there. So, a lot of the original description no longer applies.

## Original description

But I'm not sure this is exposed to developers in the best way.

Presently, you need to pass `-PincludeOpenxrValidationLayers=true` to Gradle when building, and they'll be included in all the debug builds, and so using a debug build means you get the validation layers.

However, I wonder if it would be better to have an export option that can enable/disable adding the JSON files that are necessary to automatically enable the layers, so developers can easily turn them on and off at export time?

Another option is including them as "explicit" layers (rather than "implicit" layers, as it's currently doing) and then add a configuration option in Godot that can enable the layers if present (and adding them as "explicit" layers would make them present on Android)

@m4gr3d You also mentioned an idea of including them in a "dev" build (so, we'd have "debug", "release" and "dev" builds) but I'm not entirely sure how that would work? How does a developer choose to use a "dev" build?

If anyone decides to test actually test this:

1. You need to use Vulkan Mobile; it'll crash if you use OpenGL. I think this has to do with the options used when building the validation layers: I think they need to be built differently to support OpenGL, and I'm not sure if they can support both OpenGL and Vulkan in the same build. Since it works with Vulkan, that's what I've been testing with
2. To at least run the demo project, you need to also include the changes from PRs https://github.com/GodotVR/godot_openxr_vendors/pull/344 and https://github.com/godotengine/godot/pull/109969. Without those, the demo will crash!
3. There are some bugs in the upstream validation layers that will cause problems (for example, hand tracking may not work, depending on which additional extensions you have enabled) or even crash the app. For now, I recommend testing with the simplest app possible, with the fewest extensions enabled


plugin/build.gradle
```diff
@@ -217,6 +237,76 @@ dependencies {
     implementation "androidx.core:core-splashscreen:$versions.splashscreenVersion"
 }
 
+task buildOpenxrValidationLayersAndroid {
+    for (abi in ['arm64-v8a', 'x86_64']) {
+        def suffix = (abi == 'arm64-v8a') ? 'Arm64' : (abi == 'x86_64') ? 'X86_64' : abi
+
+        def libraryTargetDir = file("${khronosOpenxrValidationLayersAndroidLibraryTargetDir}/${abi}/${abi}")
```
m4gr3d: Why does the `libraryTargetDir` need `${abi}` twice as suffix?

plugin/build.gradle
```diff
@@ -217,6 +237,76 @@ dependencies {
...
+            into jsonTargetDir
+            dependsOn "buildOpenxrValidationLayersAndroid${suffix}"
+        }
+    }
+
+    dependsOn "copyOpenxrValidationLayersAndroidArm64Library"
+    dependsOn "copyOpenxrValidationLayersAndroidArm64JSON"
+
+    dependsOn "copyOpenxrValidationLayersAndroidX86_64Library"
+    dependsOn "copyOpenxrValidationLayersAndroidX86_64JSON"
```
m4gr3d: These could also be in the `for` loop.

plugin/build.gradle
dsnopek: It's because we are conditionally adding the path for the specific ABIs. So, for example, when we want to include x86_64 binaries, we tell Gradle to use a particular path, ("${khronosOpenxrValidationLayersAndroidLibraryTargetDir}/${abi}") and then it wants that path to also contain a directory named with the ABI.

We're already doing the same thing for the Khronos OpenXR Loader - this is just repeating that for the validation layers.

m4gr3d:
> But I'm not sure this is exposed to developers in the best way.

We can categorize the mechanism to expose the functionality to developers in two groups:
- Unbundled: the validation layers are not bundled in the default debug / release binaries we generate for the plugin, a developer has to either make their own build which include the validation layers, or grab an alternate build (e.g: dev build) (that we could provide in the github download page) that includes the validation layers.
- Bundled: the validation layers are bundled in the default debug binaries we generate for the plugin, and developers can turn them on / off either via an export option, or another mechanism.

The bundled option would be the more user-friendly option, however the increase in binary size might be a major cons. Do you know the size for debug build without validation layers vs one with validation layers?

dsnopek:
> Unbundled: the validation layers are not bundled in the default debug / release binaries we generate for the plugin, a developer has to either make their own build which include the validation layers, or grab an alternate build

I think it's worth pointing that it's also possible to use the OpenXR validation layers without any changes to the godot_openxr_vendors extension.

Like I said before creating this PR, an alternative to building this into godot_openxr_vendors would just be a short bit of documentation explaining how to enable the validation layers in a Godot app (you just have to put the files in very specific places it your project). The main inconvenience is that Khronos doesn't make Android binaries available. But this doesn't _need_ to be built into godot_openxr_vendors at all.

> The bundled option would be the more user-friendly option, however the increase in binary size might be a major cons. Do you know the size for debug build without validation layers vs one with validation layers?

It's more user-friendly, but I doubt many normal users will want to use the validation layers. The errors aren't actionable for them (other than to make an issue for us, I guess), and if a validation error is encountered, it'll prevent the given feature from working, whereas it may have worked just fine even though we were doing something wrong (or there is a bug in the validation layers). All the cases so far where we fixed something based on the validation errors worked fine, even if we were using the API incorrectly.

Here's the size of the `.bin/android/debug/godotopenxr-meta-debug.aar`  file (all the AARs have a similar size):

- `master`: 2.6mb
- This PR: 5.6mb

I'm actually surprised by how small the difference is - when I previously got the 81mb, I must have had debug symbols enabled? Checking now, the `libXrApiLayer_core_validation.so` decompressed from the AAR file is only 11.4mb in total for both architectures. My most recent builds were without debug symbols

This will have a bigger impact on our total build size, though, because these .so files are duplicated in every AAR

validation_layers/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+        targetCompatibility versions.javaVersion
+    }
+
+    kotlinOptions {
+        jvmTarget = versions.javaVersion
+    }
+}
+
+dependencies {
+    implementation "androidx.fragment:fragment:$versions.fragmentVersion"
```
m4gr3d: I don't think we need this since we have no java / kotlin code.

validation_layers/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+
+    sourceSets {
+        main {
+            jniLibs.srcDirs += [
+                "${khronosOpenxrValidationLayersAndroidLibraryTargetDir}/arm64-v8a",
+                "${khronosOpenxrValidationLayersAndroidLibraryTargetDir}/x86_64",
+            ]
+
+            // We don't need to give separate values for arm64-v8a and x86_64 - just one of them is fine.
+            assets.srcDirs += ["${khronosOpenxrValidationLayersAndroidAssetsTargetDir}/arm64-v8a"]
```
m4gr3d: Let's just use the regular `jnilibs` directory (e.g: `src/main/jniLibs`) and regular assets directory (e.g: `src/main/assets`)

validation_layers/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    compileOptions {
+        sourceCompatibility versions.javaVersion
+        targetCompatibility versions.javaVersion
+    }
+
+    kotlinOptions {
+        jvmTarget = versions.javaVersion
+    }
```
m4gr3d, line 97: You should include the `publishing` block to configure for publication to maven central

https://github.com/GodotVR/godot_openxr_vendors/blob/d0b90451a19d18b1ab66fa99fc2e501d7461920b/plugin/build.gradle#L156

validation_layers/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+def khronosOpenxrValidationLayersAndroidBuildDir = file("${buildDir}/temp/openxr_validation_layers_build")
+
+android {
+    compileSdk versions.compileSdk
+    ndkVersion versions.ndkVersion
+
+    defaultConfig {
+        minSdk versions.minSdk
+        targetSdk versions.targetSdk
+        versionName getReleaseVersion()
```
m4gr3d: Do the validation layers have their own version?

plugin/src/main/cpp/include/export/validation_layers_export_plugin.h
```diff
@@ -0,0 +1,56 @@
...
+	OpenXRValidationLayersEditorExportPlugin();
+
+	String _get_name() const override;
+
+	bool _supports_platform(const Ref<EditorExportPlatform> &platform) const override;
+
+	String _get_export_option_warning(const Ref<EditorExportPlatform> &platform, const String &option) const override;
+	TypedArray<Dictionary> _get_export_options(const Ref<EditorExportPlatform> &platform) const override;
+
+	PackedStringArray _get_android_libraries(const Ref<EditorExportPlatform> &platform, bool debug) const;
```
m4gr3d: Should also include the override for `_get_android_dependencies` after the `publishing` configuration mentioned below is done.

validation_layers/build.gradle
dsnopek: They are versioned with together with OpenXR. So, with the submodule that's checked out, this would be 1.1.53

validation_layers/build.gradle
dsnopek: I usually try to reserve the regular directories for things we're actually committing to the repo. We can certainly put them there and setup `.gitignore`, but since these are "temporary" things that are created in the build process, that's why I have them using a "temp" directory

validation_layers/build.gradle
dsnopek: Removed!

validation_layers/build.gradle
m4gr3d: Sounds good, let's keep this approach then.

validation_layers/build.gradle
dsnopek: I'm pretty confused about how to do this. For normal vendor plugin, we've got `scripts/publish-module.gradle` and `scripts/publish-root.gradle`. Do I need to recreate this for the validation layers? Or, can I somehow include the validation layers in that?

dsnopek:
Aha, it appears my earlier attempts to fix OpenGL worked! However, using the validation layers with Godot 4.5 will still crash, because it doesn't have https://github.com/godotengine/godot/pull/109969

So, maybe this should actually be saved for version 5.x?

m4gr3d (approved):
Looks good!

As discussed, I'll follow up this PR with the addition to push the new library to maven central.

validation_layers/build.gradle
```diff
@@ -0,0 +1,147 @@
...
+        def jsonTargetDir = file("${khronosOpenxrValidationLayersAndroidAssetsTargetDir}/${abi}/openxr/1/api_layers/implicit.d")
+        def buildDir = file("${khronosOpenxrValidationLayersAndroidBuildDir}/${abi}")
+
+        tasks.create(name: "configureOpenxrValidationLayersAndroid${suffix}", type: Exec) {
+            doFirst {
+                buildDir.mkdirs()
+            }
+            workingDir buildDir
+            commandLine [
+                "cmake",
```
m4gr3d: @dsnopek This fails on Windows if `cmake` is not on the `PATH` env variable.

I'm wondering if instead we can use the `cmake` binary provided by Android Studio..

m4gr3d: I was able to provide a valid `cmake` path but now I'm running into a `Python3 not found` issue, even though I have `python3` installed and on the `PATH` env variable.

```
> Task :validation_layers:configureOpenxrValidationLayersAndroidArm64
-- Building for: Visual Studio 17 2022
-- Android: Targeting API '35' with architecture 'arm64', ABI 'arm64-v8a', and processor 'aarch64'
-- Android: Selected unified Clang toolchain
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Users/fhuya/AppData/Local/Android/Sdk/ndk/28.1.13356709/toolchains/llvm/prebuilt/windows-x86_64/bin/clang.exe - skipped
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Users/fhuya/AppData/Local/Android/Sdk/ndk/28.1.13356709/toolchains/llvm/prebuilt/windows-x86_64/bin/clang++.exe - skipped
-- Could NOT find Python3 (missing: Python3_EXECUTABLE Interpreter) 
-- Performing Test HAVE_FILESYSTEM_IN_STD
-- Performing Test HAVE_FILESYSTEM_IN_STD - Failed
-- Performing Test HAVE_FILESYSTEM_IN_STDEXPERIMENTAL
-- Performing Test HAVE_FILESYSTEM_IN_STDEXPERIMENTAL - Failed
-- Performing Test HAVE_FILESYSTEM_IN_STD_17
-- Performing Test HAVE_FILESYSTEM_IN_STD_17 - Failed
-- Performing Test HAVE_FILESYSTEM_WITHOUT_LIB
-- Performing Test HAVE_FILESYSTEM_WITHOUT_LIB - Failed
-- Performing Test HAVE_FILESYSTEM_NEEDING_LIBSTDCXXFS
-- Performing Test HAVE_FILESYSTEM_NEEDING_LIBSTDCXXFS - Failed
-- Performing Test HAVE_FILESYSTEM_NEEDING_LIBCXXFS
-- Performing Test HAVE_FILESYSTEM_NEEDING_LIBCXXFS - Failed
-- Found Vulkan: C:/Users/fhuya/AppData/Local/Android/Sdk/ndk/28.1.13356709/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/lib/aarch64-linux-android/35/libvulkan.so  
-- Enabling Vulkan support
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Threads: TRUE  
-- Could NOT find JsonCpp (missing: JsonCpp_INCLUDE_DIR JsonCpp_LIBRARY) 
-- Could NOT find glslc, using precompiled .spv files
-- Looking for secure_getenv
-- Looking for secure_getenv - not found
-- Looking for __secure_getenv
-- Looking for __secure_getenv - not found
-- Looking for clock_gettime
-- Looking for clock_gettime - found
-- OpenXR 1.1.53
CMake Error at src/CMakeLists.txt:443 (message):
  Python 3 not found, but pre-generated
  D:/playground/godotengine/v4x/godot_openxr_vendors/thirdparty/openxr-source/src/xr_generated_dispatch_table.h
  not found
Call Stack (most recent call first):
  src/CMakeLists.txt:510 (run_xr_xml_generate)
```

validation_layers/build.gradle
m4gr3d: So should this be set to the OpenXR version instead of setting it to the vendors plugin version?

dsnopek: We could go either way.

The OpenXR version will probably match whatever OpenXR version we use in godot_openxr_vendors, so versioning it like godot_openxr_vendors means it's the matching version. I'd probably lean towards that actually!

But since it doesn't really _depend_ on the other stuff in the plugin, using the "true" version could also make sense

validation_layers/build.gradle
dsnopek: Ugh, Windows is the worst!

Using the cmake from the Android build chain is a pretty good idea, because we know folks have it. I'm surprised it can't find Python, because you also need to have Python in order to run scons, and that somehow works :-/

When I have a chance to boot up Windows, I'll see what I can figure out

dsnopek: Poking around the internet, apparently, you can pass in `-DPython3_EXECUTABLE="C:/path/to/python.exe"` on the cmake configuration command. That may work for you?

It's not the ideal solution, though. Supposedly, cmake should be able to find it automatically

m4gr3d: Thanks, I'll give it a try!

Things seems to work as expected on macos, so Windows is the only outlier.

m4gr3d: I managed to solve the build issues on windows, so this PR should be good to go once you merge the changes in https://github.com/dsnopek/godot_openxr_vendors/pull/1

dsnopek: Awesome, thanks! I've pulled in your changes from that PR (and added you as a `Co-authored-by` on the commit)

Issue #351, opened 2025-09-01
Add an occlusion setting when using the Meta Depth API


plrepin:
Meta Depth API support has recently been added to the plugin, however, the Hard Occlusion shader is used by default. To improve the user experience and simplify development, you can add an option in the settings between hard occlusion and smoothed. Since the hard version looks terrible by default, and the smoothed version is more acceptable.

dsnopek:
The [`XR_META_environment_depth`](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_META_environment_depth) OpenXR extension doesn't provide anything like "Hard Occlusion" or "Soft Occlusion". Is this a feature of the Unity or Unreal plugin?

Possibly changes could be made to [the shader](https://github.com/GodotVR/godot_openxr_vendors/blob/master/plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension_wrapper.cpp#L72) we have that reprojects the depth data into the Godot render's depth buffer? Suggestions for how to improve that shader are welcome!

Otherwise, we just make the data from the OpenXR extension available, and developers are free to use that in their own shaders how they see fit.

This PR will add some more documentation about what this extension is doing, which may be useful:

https://github.com/GodotVR/godot_openxr_vendors/pull/347

dsnopek:
Thanks!

The temporal stability part makes it seem like it may be calculated over multiple frames, which could be a little tricky to figure out how to do

PR #354, opened 2025-09-05
Allow using bilinear filtering in reprojection and provide example for smoothing in a shader

dsnopek:
This attempts to address #351 somewhat, although, it doesn't do exactly what "Soft Occlusion" in the Unity plugin does.

The Unity plugin preprocesses the depth texture, and then uses that to do some additional math that ultimately leads to smoothly changing the alpha on the material, so there are softer edges rather than a hard cutoff. This works there because developers have to update their shaders to use some macros provided by the plugin, which will inject the right math.

However, in godot_openxr_vendors, we're promoting the reprojection into the depth buffer as the primary way to use the environment depth map, and that doesn't allow us to have any effect on the alpha on the material.

So, instead, this PR adds optional bilinear filtering in the reprojection, which improves this somewhat by making for straighter (rather than jagged) edges. And, additionally, it gives more examples for how developers could make even smoother edges by adjusting `ALPHA` in their own custom shaders. The examples don't do the same technique as the Unity plugin, but do provide a fairly decent result

Closes https://github.com/GodotVR/godot_openxr_vendors/issues/351


docs/manual/meta/environment_depth.rst
```diff
@@ -118,20 +120,82 @@ For example, if you **don't** use an ``OpenXRMetaEnvironmentDepth`` node, you ca
...
 	global uniform highp mat4 META_ENVIRONMENT_DEPTH_PROJECTION_VIEW_LEFT;
 	global uniform highp mat4 META_ENVIRONMENT_DEPTH_PROJECTION_VIEW_RIGHT;
 
 	uniform vec3 albedo : source_color;
 
 	varying vec4 world_position;
 
+	// 0 = Hard occlusion
+	// 1 = Soft occlusion via bilinear filtering
+	// 2 = Soft occulsion via PCF 3x3
```
m4gr3d:
```suggestion
	// 2 = Soft occlusion via PCF 3x3
```

dsnopek: Thanks! Fixed in my latest push

dsnopek:
@devloglogan Thanks!

> When switching between setting the `bilinear_filtering` to `true` and `false` at runtime I do see this error print out 10 times

Hm, this seems like a Godot bug of some kind. We're just calling `Shader::set_code()` with new code that has (or doesn't have) a `#define` - if that's having some issue making the appropriate OpenGL calls, I don't think there's anything we can do from here.

PR #355, opened 2025-09-08
Fix vendor performance metrics setter logic

devloglogan:
Dumb typo on my end slipped in and went uncaught because I wasn't initializing `provider` as `nullptr`. 🙂 This should be correct now.

### DIFF
--- a/plugin/src/main/cpp/classes/openxr_vendor_performance_metrics.cpp
+++ b/plugin/src/main/cpp/classes/openxr_vendor_performance_metrics.cpp
@@ -66,15 +66,12 @@ bool OpenXRVendorPerformanceMetrics::is_enabled() {
 }
 
 void OpenXRVendorPerformanceMetrics::set_vendor_performance_metrics_provider(OpenXRVendorPerformanceMetricsProvider *p_provider) {
-	if (provider == p_provider) {
+	if (provider != nullptr && provider->is_enabled()) {
+		WARN_PRINT("An OpenXRVendorPerformanceMetricsProvider is already set; ignoring new provider");
 		return;
 	}
 
 	provider = p_provider;
-}
-
-OpenXRVendorPerformanceMetricsProvider *OpenXRVendorPerformanceMetrics::get_vendor_performance_metrics_provider() {
-	return provider;
 }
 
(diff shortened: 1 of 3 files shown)


plugin/src/main/cpp/classes/openxr_vendor_performance_metrics.cpp
```diff
@@ -66,7 +66,8 @@ bool OpenXRVendorPerformanceMetrics::is_enabled() {
 }
 
 void OpenXRVendorPerformanceMetrics::set_vendor_performance_metrics_provider(OpenXRVendorPerformanceMetricsProvider *p_provider) {
-	if (provider == p_provider) {
+	if (provider != nullptr && provider->is_enabled()) {
+		UtilityFunctions::print_verbose("An OpenXRVendorPerformanceMetricsProvider is already set; ignoring new provider");
```
m4gr3d: Should this be a warning? 
More than one provider being active means that the runtime supports multiple performance metrics extensions, so we may want to explicitly call this out to the developer.

plugin/src/main/cpp/include/classes/openxr_vendor_performance_metrics.h
```diff
@@ -77,7 +77,7 @@ class OpenXRVendorPerformanceMetrics : public Object {
 private:
 	static OpenXRVendorPerformanceMetrics *singleton;
 
-	OpenXRVendorPerformanceMetricsProvider *provider;
+	OpenXRVendorPerformanceMetricsProvider *provider = nullptr;
```
m4gr3d, line 79: Do we still need the getter for this field?

plugin/src/main/cpp/classes/openxr_vendor_performance_metrics.cpp
devloglogan: Ultimately I was on the fence on if it should be a warning or not; but I'm OK with either. I've made it a warning in the latest push!

plugin/src/main/cpp/include/classes/openxr_vendor_performance_metrics.h
devloglogan: Ha, good point! I've removed the getter.

PR #362, opened 2025-09-16
Generate native debug symbols for Android

dsnopek:
This adds some steps to the `mavencentral-publish.yml` workflow that will zip up the native .so files with debug symbols, which we need to get backtraces with symbols on the stores. It should be fine to enable debug symbols there, because they'll get stripped by Gradle

This meant removing the blanket `doNotStrip` we had before, which means if you are making builds with debug symbols for use during development you'll need to pass `-PdoNotStrip=true` when running Gradle. But since we didn't even provide a nice way of doing that until super recently (in https://github.com/GodotVR/godot_openxr_vendors/pull/356) I think that's fine

I tested this as much as I could on my fork by making some fake releases. But we won't know if it works for sure until we run it here on a real release


dsnopek:
> Can you update the comments at the top of the mavencentral-publish workflow to indicate it also generates the native debug symbols files

Sure! I've updated the comment at the top of the file in my latest push

Issue #352, opened 2025-09-03
Hand Meshes disappear after scene change


JonasPaule:
I have an issue developing a game for my Meta Quest 3 on Godot 4.4.1 using the Vendors Plugin v4.
So, in my project I want to switch between multiple 3D scenes. In order not to have to add my XROrigin, camera and handtracking to each scene individually, I have created a scene called base_view which does this. It is structured like this: 

<img width="251" height="256" alt="Image" src="https://github.com/user-attachments/assets/e0f0c7e8-25e3-4b78-b2c3-0741eaaeb8e4" />

Then I just add this base_view scene to my game scenes and have all the basic XR functionalities working. When I start the game in my main scene everything is working fine, the hands are tracked and the OpenXRFbHandTrackingMeshes are shown accordingly. Yet, when I change my scene using `get_tree().call_deferred("change_scene_to_file", "res://scene2.tscn")`  I switch to my new scene but my hands are not shown anymore. In my minimal project the 2 scenes i switch between look like this: 

<img width="1344" height="754" alt="Image" src="https://github.com/user-attachments/assets/a79edae6-968b-47ca-b4de-75af22fc3d8a" />

<img width="276" height="146" alt="Image" src="https://github.com/user-attachments/assets/e353b8ef-4cad-44d8-b077-0f172ef61430" />

From my bigger project I know that the hands are still being correctly tracked as I use them to Interact with a GUI, utilizing the capsule feature. So after switching the scene everything is still working fine, I just do not see the Meshes for whatever reason. Does anybody know how to fix this?

PR #367, opened 2025-09-19
Fix `OpenXRFbHandTrackingMesh` changes after startup and detecting when it's not supported

dsnopek:
This aims to fix some issues I found with using `OpenXRFbHandTrackingMesh` in my recent [jam game](https://dsnopek.itch.io/rock-paper-infinity):

- If you have some `OpenXRFbHandTrackingMesh`s added at startup, but then add another `OpenXRFbHandTrackingMesh` later on, it'll never get its mesh populated
- If you change the `hand` property after its mesh has already been loaded, it won't switch to the other hand mesh
- There's no _good_ way to detect if this extension is enabled and loading the hash mesh has succeeded. This makes it hard to automatically use a fallback hand mesh when run on platforms that don't support the extension (it is possible to do in a hacky way, but there should be an "official" way)

Fixes https://github.com/GodotVR/godot_openxr_vendors/issues/352

### DIFF
--- a/doc_classes/OpenXRFbHandTrackingMeshExtensionWrapper.xml
+++ b/doc_classes/OpenXRFbHandTrackingMeshExtensionWrapper.xml
@@ -8,12 +8,4 @@
 	</description>
 	<tutorials>
 	</tutorials>
-	<signals>
-		<signal name="openxr_fb_hand_tracking_mesh_data_fetched">
-			<param index="0" name="hand_index" type="int" />
-			<description>
-				Emitted when the mesh data has been fetched for the given hand ([code]0[/code] is left, and [code]1[/code] is right).
-			</description>
-		</signal>
-	</signals>
 </class>
--- a/plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
+++ b/plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
@@ -39,19 +39,28 @@
 
 using namespace godot;
 
-void OpenXRFbHandTrackingMesh::setup_hand_mesh(Hand p_hand) {
-	if (p_hand != hand) {
+void OpenXRFbHandTrackingMesh::setup_hand_mesh(const Ref<Mesh> &p_mesh) {
+	if (p_mesh.is_null()) {
+		emit_signal("openxr_fb_hand_tracking_mesh_unavailable");
 		return;
 	}
 
-	ERR_FAIL_COND_MSG(mesh_instance != nullptr, "OpenXRFbHandTrackingMesh has already been set up.");
+	OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->reset_skeleton_pose(hand, this);
 
-	OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->reset_skeleton_pose(p_hand, this);
+	if (!mesh_instance) {
+		mesh_instance = memnew(MeshInstance3D);
+		mesh_instance->set_material_override(material);
+		mesh_instance->set_mesh(p_mesh);
+		add_child(mesh_instance);
+	} else {
+		// We need to remove/add the mesh instance for the mesh and skeleton to become connected.
+		// @todo Investigate if this is a Godot bug!
+		remove_child(mesh_instance);
+		mesh_instance->set_mesh(p_mesh);
+		add_child(mesh_instance);
+	}
 
-	mesh_instance = memnew(MeshInstance3D);
-	mesh_instance->set_mesh(OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->get_mesh(p_hand));
-	mesh_instance->set_material_override(material);
-	add_child(mesh_instance);
+	emit_signal("openxr_fb_hand_tracking_mesh_ready");
 }
 
 MeshInstance3D *OpenXRFbHandTrackingMesh::get_mesh_instance() const {
@@ -76,6 +85,12 @@ void OpenXRFbHandTrackingMesh::set_hand(Hand p_hand) {
 			set_bone_name(i, bone_name);
 		}
 	}
+
+	// If we already have mesh data, but now we're changing hands, we need to request it again.
+	if (is_node_ready() && ProjectSettings::get_singleton()->get_setting_with_override("xr/openxr/extensions/meta/hand_tracking_mesh")) {
+		OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->request_hand_mesh_data(hand, callable_mp(this, &OpenXRFbHandTrackingMesh::setup_hand_mesh));
+	}
+
 	notify_property_list_changed();
 }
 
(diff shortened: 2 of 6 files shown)


plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
```diff
@@ -39,19 +39,26 @@
...
-	OpenXRFbHandTrackingMeshExtensionWrapper::get_singleton()->reset_skeleton_pose(p_hand, this);
+	if (!mesh_instance) {
+		mesh_instance = memnew(MeshInstance3D);
+		mesh_instance->set_material_override(material);
+		mesh_instance->set_mesh(p_mesh);
+		add_child(mesh_instance);
+	} else {
+		remove_child(mesh_instance);
+		mesh_instance->set_mesh(p_mesh);
+		add_child(mesh_instance);
```
devloglogan, line 60: What's the reasoning for removing/re-adding the child? Can we not just `set_mesh()`?

plugin/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -120,35 +119,34 @@ uint64_t OpenXRFbHandTrackingMeshExtensionWrapper::_set_hand_joint_locations_and
...
+	if (!should_fetch_hand_mesh_data) {
 		return;
 	}
 
-	if (!should_fetch_hand_mesh_data) {
+	if (!is_enabled() || get_openxr_api().is_null()) {
 		return;
 	}
 
+	XrHandTrackerEXT hand_trackers[2];
```
devloglogan:
```suggestion
	XrHandTrackerEXT hand_trackers[Hand::HAND_MAX];
```

devloglogan: Other arrays use `Hand::HAND_MAX` instead of `2`, just feels more consistent.

plugin/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
```diff
@@ -120,35 +119,34 @@ uint64_t OpenXRFbHandTrackingMeshExtensionWrapper::_set_hand_joint_locations_and
...
-				}
-			}
+		hand_trackers[i] = reinterpret_cast<XrHandTrackerEXT>(get_openxr_api()->get_hand_tracker(i));
+		if (hand_trackers[i] == XR_NULL_HANDLE) {
+			// If we're missing a hand tracker, we can't fetch the data.
+			return;
 		}
 	}
 
+	bool success[2];
```
devloglogan:
```suggestion
	bool success[Hand::HAND_MAX];
```

plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
dsnopek: Otherwise the new mesh and the skeleton don't get connected. This seems like a Godot bug, but one I haven't had a chance to dig into yet

plugin/src/main/cpp/extensions/openxr_fb_hand_tracking_mesh_extension_wrapper.cpp
dsnopek: Ah, yes, changed in my latest push!

dsnopek:
> Could you fix the [binding of the hand enums](https://github.com/GodotVR/godot_openxr_vendors/blob/55d2a3ef1acee702adcc1e465c322322130092a6/plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp#L159) as a part of this PR as well?

Sure! Changed in my latest push (along with a few of the other suggestions)

doc_classes/OpenXRFbHandTrackingMeshExtensionWrapper.xml
```diff
@@ -8,12 +8,4 @@
...
 	<tutorials>
 	</tutorials>
-	<signals>
-		<signal name="openxr_fb_hand_tracking_mesh_data_fetched">
-			<param index="0" name="hand_index" type="int" />
-			<description>
-				Emitted when the mesh data has been fetched for the given hand ([code]0[/code] is left, and [code]1[/code] is right).
-			</description>
-		</signal>
-	</signals>
```
m4gr3d, line 18: Does this warrant a major version bump given that functionality is being removed?

plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
m4gr3d: Can you add a comment to describe that for future reference.

plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
```diff
@@ -39,19 +39,26 @@
 
 using namespace godot;
 
-void OpenXRFbHandTrackingMesh::setup_hand_mesh(Hand p_hand) {
-	if (p_hand != hand) {
+void OpenXRFbHandTrackingMesh::setup_hand_mesh(bool p_success, const Ref<Mesh> &p_mesh) {
+	if (!p_success) {
+		emit_signal("openxr_fb_hand_tracking_mesh_unavailable");
 		return;
 	}
```
m4gr3d: Seems like we can remove the `p_success` argument either by checking if `p_mesh.is_valid()` or by emitting the signal when hand mesh fetch fails outside of this method.

doc_classes/OpenXRFbHandTrackingMeshExtensionWrapper.xml
dsnopek: No, that signal only existed as a back channel for `OpenXRFbHandTrackingMeshExtensionWrapper` to communicate with `OpenXRFbHandTrackingMesh`. Since the singleton for `OpenXRFbHandTrackingMeshExtensionWrapper` isn't registered, there's no way a developer could have ever connected to it.

(Arguably all the extension wrappers we don't expose the singletons for should be registered with `GDREGISTER_INTERNAL_CLASS()` so we don't have to deal with docs for them, but that's something we should address elsewhere)

plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
dsnopek: Comment added!

plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp
dsnopek: Sure! I've removed `p_success` and it's now just checking if `p_mesh` is null or valid

m4gr3d (approved):

PR #369, opened 2025-09-23
Use `gdformat` to ensure consistent GDScript coding standards in CI

dsnopek:
This uses gdformat (from [gdtoolkit](https://github.com/Scony/godot-gdscript-toolkit)) in CI: if things look good tests should pass; if not, then it'll show a patch with the necessary changes

~~NOTE: This first push should fail - I want to test that failure works :-) but I'll fix in a subsequent push~~

### DIFF
--- a/.github/workflows/static_checks.yml
+++ b/.github/workflows/static_checks.yml
@@ -3,15 +3,22 @@ on: [push, pull_request]
 
 jobs:
   static-checks:
-    name: Formatting (clang-format, file format)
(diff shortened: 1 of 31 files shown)


BastiaanOlij:
XR Meeting: we agreed on disabling the line limit.

dsnopek:
I've disabled the line limit in my latest push! This makes the changes to our pre-existing GDScript files much smaller

PR #375, opened 2025-10-06
Build with NDK version 28.1.13356709

dsnopek:
Per @m4gr3d on [this comment](https://github.com/GodotVR/godot_openxr_vendors/pull/374#issuecomment-3369192851), this is needed for compatibility with 16kb page support


build.gradle
```diff
@@ -89,6 +89,7 @@ def createSconsTasks(File sconsExecutableFile, boolean clean) {
         "--directory=.",
         "custom_api_file=thirdparty/godot_cpp_gdextension_api/extension_api.json",
         "build_profile=thirdparty/godot_cpp_build_profile/build_profile.json",
+        "ndk_version=28.1.13356709",
```
m4gr3d: You should be able to use the `ndkVersion` constant defined and `config.gradle` and do `"ndk_version=$versions.ndkVersion"`

dsnopek: Ah, great idea! Updated in my latest push

devloglogan (approved):
Works great now, thanks!

PR #382, opened 2025-10-28
Add OpenXR META colocation discovery extension support

Clancey:
Introduces the OpenXRMetaColocationDiscovery class to support the XR_META_colocation_discovery extension, including registration, project setting, and Android permission handling. This enables colocation advertisement and discovery features for supported devices and exposes related methods and signals to Godot XR.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_meta_colocation_discovery_extension_wrapper.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_meta_colocation_discovery_extension_wrapper.cpp
@@ -1,0 +1,343 @@
+/**************************************************************************/
+/*  openxr_meta_colocation_discovery.cpp                                 */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "extensions/openxr_meta_colocation_discovery_extension_wrapper.h"
+
+#include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+using namespace godot;
+
+OpenXRMetaColocationDiscoveryExtensionWrapper *OpenXRMetaColocationDiscoveryExtensionWrapper::singleton = nullptr;
+
+OpenXRMetaColocationDiscoveryExtensionWrapper *OpenXRMetaColocationDiscoveryExtensionWrapper::get_singleton() {
+	if (singleton == nullptr) {
+		singleton = memnew(OpenXRMetaColocationDiscoveryExtensionWrapper());
+	}
+	return singleton;
+}
+
+OpenXRMetaColocationDiscoveryExtensionWrapper::OpenXRMetaColocationDiscoveryExtensionWrapper() :
+		OpenXRExtensionWrapperExtension() {
+	ERR_FAIL_COND_MSG(singleton != nullptr, "An OpenXRMetaColocationDiscoveryExtensionWrapper singleton already exists.");
+
+	request_extensions[XR_META_COLOCATION_DISCOVERY_EXTENSION_NAME] = &meta_colocation_discovery_ext;
+	singleton = this;
+}
+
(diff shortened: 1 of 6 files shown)


doc_classes/OpenXRMetaColocationDiscovery.xml
```diff
@@ -0,0 +1,126 @@
...
+		<method name="stop_discovery">
+			<return type="bool" />
+			<description>
+				Stops the current discovery operation. Returns [code]true[/code] if the stop request was successfully initiated, [code]false[/code] otherwise.
+				Listen to the [signal discovery_stopped] signal to know when discovery actually stops.
+			</description>
+		</method>
+	</methods>
+	<signals>
+		<signal name="advertisement_started">
```
m4gr3d: For vendor extension signals the convention is to use the class name as prefix.
See https://godotvr.github.io/godot_openxr_vendors/api/class_openxrfbscenemanager.html#signals for an example.

plugin/src/main/cpp/include/extensions/openxr_meta_colocation_discovery.h
```diff
@@ -0,0 +1,120 @@
...
+	void on_discovery_result(const XrEventDataColocationDiscoveryResultMETA *event);
+	void on_discovery_complete(const XrEventDataColocationDiscoveryCompleteMETA *event);
+	void on_stop_discovery_complete(const XrEventDataStopColocationDiscoveryCompleteMETA *event);
+
+	static OpenXRMetaColocationDiscovery *singleton;
+
+	std::map<godot::String, bool *> request_extensions;
+	bool meta_colocation_discovery_ext = false;
+
+	XrSession openxr_session = XR_NULL_HANDLE;
```
m4gr3d: You can use [SESSION](https://github.com/GodotVR/godot_openxr_vendors/blob/2dbc90fc788d1cd16ed9141ff8ec6942d1696c39/plugin/src/main/cpp/include/util.h#L126) to get access to the session.

plugin/src/main/cpp/extensions/openxr_meta_colocation_discovery.cpp
```diff
@@ -0,0 +1,350 @@
...
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
+#include "extensions/openxr_meta_colocation_discovery.h"
+
+#include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+#include "extensions/openxr_fb_spatial_entity_extension_wrapper.h"
```
devloglogan: I think this include can be removed

Clancey: I still need ```#include <godot_cpp/classes/open_xrapi_extension.hpp>
#include <godot_cpp/variant/utility_functions.hpp>``` But I can remove `#include "extensions/openxr_fb_spatial_entity_extension_wrapper.h"`

doc_classes/OpenXRMetaColocationDiscovery.xml
Clancey: Fixed

plugin/src/main/cpp/include/extensions/openxr_meta_colocation_discovery.h
Clancey: Fixed

doc_classes/OpenXRMetaColocationDiscovery.xml
m4gr3d: Thanks for the update!

Note that the prefix doesn't have to strictly duplicate the class name, just provide enough to distinguish so we can probably drop the `discovery` portion from the signals name.

m4gr3d (commented):
Btw, the [documentation](https://developers.meta.com/horizon/documentation/native/android/openxr-colocation-discovery-api-ref/) states that internet access may be required; have you found that to be the case, and if so should this be noted in the documentation?

plugin/src/main/cpp/include/extensions/openxr_meta_colocation_discovery.h
```diff
@@ -0,0 +1,117 @@
...
+		nullptr, // next
+		XR_FALSE // supportsColocationDiscovery
+	};
+
+	// Track current advertisement request
+	XrAsyncRequestIdFB current_advertisement_request_id = 0;
+	bool is_advertising = false;
+
+	// Track current discovery request
+	XrAsyncRequestIdFB current_discovery_request_id = 0;
```
m4gr3d: The request IDs don't seem to be used, should they be removed?

dsnopek:
Thanks! I've wanted to implement this extension for a long time, but just haven't gotten around to it :-)

I've got a co-located multiplayer game (which is currently based on exchanging codes with the other player), which I'll test this with as soon as I can

plugin/src/main/cpp/include/extensions/openxr_meta_colocation_discovery.h
dsnopek: Yeah, given that we can only start advertising or discovering once, the request IDs don't seem very useful. I agree that it would be better to just not store them in member variables because that could be confusing for folks reading this code later

plugin/src/main/cpp/include/extensions/openxr_meta_colocation_discovery.h
```diff
@@ -0,0 +1,117 @@
...
+#include <godot_cpp/classes/open_xr_extension_wrapper_extension.hpp>
+#include <godot_cpp/templates/hash_map.hpp>
+#include <map>
+
+#include "util.h"
+
+using namespace godot;
+
+// Wrapper for the XR_META_colocation_discovery extension.
+class OpenXRMetaColocationDiscovery : public OpenXRExtensionWrapperExtension {
```
dsnopek: Ah, one thing I just noticed: this doesn't follow our usual naming convention of `*ExtensionWrapper` for all our extension wrappers. I'm not the biggest fan of that convention (really just the `*Wrapper` part of it), but we should still use it for consistency

```suggestion
class OpenXRMetaColocationDiscoveryExtensionWrapper : public OpenXRExtensionWrapperExtension {
```

dsnopek (approved):
To expedite things, I used maintainer powers to update the PR for the requested changes. I hope you don't mind!

plugin/src/main/cpp/extensions/openxr_meta_colocation_discovery_extension_wrapper.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+void OpenXRMetaColocationDiscoveryExtensionWrapper::_on_instance_created(uint64_t instance) {
+	if (!meta_colocation_discovery_ext) {
+		return;
+	}
+
+	GDEXTENSION_INIT_XR_FUNC(xrStartColocationAdvertisementMETA);
+	GDEXTENSION_INIT_XR_FUNC(xrStopColocationAdvertisementMETA);
+	GDEXTENSION_INIT_XR_FUNC(xrStartColocationDiscoveryMETA);
+	GDEXTENSION_INIT_XR_FUNC(xrStopColocationDiscoveryMETA);
```
m4gr3d: We should reset `meta_colocation_discovery_ext` if initializing the function pointers fails.

dsnopek: Fixed in latest push!

PR #385, opened 2025-11-05
Add loading screen passthrough documentation

devloglogan:
This PR adds a section to the meta passthrough doc page detailing what settings are required to make an app have a passthrough loading screen, allowing users to open an MR app from their home environment without ever exiting passthrough. I've also applied these settings to the passthrough sample.

I believe the way this _should_ work is that the loading screen should be in passthrough (or not) depending on if the users home environment is in passthrough. However, these settings will make the app __always__ load in passthrough. This is ultimately determined entirely by [this line](https://github.com/GodotVR/godot_openxr_vendors/blob/d0b90451a19d18b1ab66fa99fc2e501d7461920b/plugin/src/main/cpp/export/meta_export_plugin.cpp#L500) in the manifest and I believe is out of our control.

I don't think a note in the doc about this potential bug is warranted since the doc is saying "this will make you load in passthrough", and it does! Rather, if it ever gets fixed, adding a note in the doc that it will depend on the passthrough state of the home environment makes sense to me.


docs/manual/meta/passthrough.rst
```diff
@@ -88,6 +83,20 @@ This is because hole-punching requires the viewport's ``transparent_bg`` propert
...
+--------------------------
+
+When making an app that utilizes passthrough, having the app load with passthrough active might be desirable to create a smoother startup experience.
+To accomplish this, some project and export settings must be set.
+
+In **Project Settings** under the **OpenXR** section, ensure that **Environment Blend Mode** is set to a value other than opaque.
+
+.. image:: img/passthrough/passthrough_loading_screen_project_settings.png
+
+In **Export Settings** under the **Meta XR Features** section, ensure that **Boundary Mode** is set to **Disabled**, and that **Instant Splash Screen** is turned on.
```
m4gr3d, line 96: Do we need `Boundary Mode` to be disabled? I believe this option subjects the app to an additional layer of scrutiny when it's submitted to the Horizon store given that safety implications.

devloglogan: I do think this is required; I was unable to get passthrough loading working without this setting.

PR #393, opened 2025-11-17
Implement XR_ANDROID_recommended_resolution

dhoverb:
Depends on https://github.com/godotengine/godot/pull/112890 and https://github.com/GodotVR/godot_openxr_vendors/pull/407


thirdparty/godot_cpp_gdextension_api/extension_api.json
```diff
@@ -174001,6 +174001,14 @@
...
 						"type": "enum::OpenXRAPIExtension.OpenXRAlphaBlendModeSupport"
 					}
+				},
+				{
+					"name": "update_main_swapchain_size",
+					"is_const": false,
+					"is_vararg": false,
+					"is_static": false,
+					"is_virtual": false,
+					"hash": 3218959716
```
dsnopek: These changes are fine for now, but we shouldn't merge them with this PR. Eventually, we'll have a PR that updates this whole file for Godot 4.6, and this PR can be rebased on that

dsnopek (commented):
Thanks!

The code looks good, and it worked great in my testing on a Samsung Galaxy XR.

However, we won't be able to merge this until after we start working on version 5.x which depends on Godot 4.6

thirdparty/godot_cpp_gdextension_api/extension_api.json
dhoverb: Done.

dsnopek:
Given that this extension provides the same high-level functionality as Meta's `XR_META_recommended_layer_resolution`, and we call that "Dynamic Resolution" in the settings and docs - should we also call this one "Dynamic Resolution" publicly?

And, should it be enabled by default like Meta's one is? Similar to the Meta one, it's basically an unambiguous win, because if you perform well, there are no changes.

This also begs the question of if we should have only one setting that enables both and uses which ever is supported?

The potential problem with that is if both are supported by an OpenXR runtime, then really weird things could happen, because they work through different mechanisms on the Godot side: Android XR's will recreate the swapchain, and Meta's will change the render region. Having both things happen at once wouldn't be ideal :-)

Although, we could bake something in to favor one over the other if both are present? So, one could "turn itself off" if the other is present (basically, we'd just refrain from using it, since once an extension is enabled, we can't disable it without re-initializing OpenXR)

My personal votes on the 3 questions:

- Rename "Dynamic Resolution": **yes**
- Enable by default: **yes**
- Merge into one setting: **no**, although, I could probably go either way. Given that there's a bit of uncertainty here, leaving this in the hands of developers with a setting seems good to me

devloglogan:
@dsnopek I'm inclined to agree on all three fronts. :slightly_smiling_face: For single vs. multiple settings, better to be safe than sorry and go with multiple IMO.

dsnopek (approved):

m4gr3d (approved):

PR #399, opened 2025-11-20
Fix background opacity on AndroidXR when using GL Compatibilty mode

ashwinvbs:


demo/addons/common/start_xr.gd
```diff
@@ -39,6 +39,10 @@ func _ready():
 		xr_interface.session_focussed.connect(_on_openxr_focused_state)
 		xr_interface.session_stopping.connect(_on_openxr_stopping)
 		xr_interface.pose_recentered.connect(_on_openxr_pose_recentered)
+
+		# If blend mode is additive or alpha blended, make viewport background transparent.
```
devloglogan: I've only ever actually used the alpha blend mode, but in the [docs](https://docs.godotengine.org/en/stable/tutorials/xr/ar_passthrough.html) we have:

>  When setting the blend mode to `XR_ENV_BLEND_MODE_ALPHA_BLEND` you must set the transparent_bg property on [Viewport](https://docs.godotengine.org/en/stable/classes/class_viewport.html#class-viewport) to `true`. When using the `XR_ENV_BLEND_MODE_ADDITIVE` blend mode you should set your background color to black.

Should we be setting it to transparent for additive as well?

ashwinvbs: Additive blending ignores the alpha channel, so setting to transparent would have no effect.
Agreed that its better to leave it default.

I have pushed an updated patch for this.

PR #400, opened 2025-11-25
Implement XR_ANDROID_depth_texture

dhoverb:
Implements [XR_ANDROID_depth_texture](https://developer.android.com/develop/xr/openxr/extensions/XR_ANDROID_depth_texture)


plugin/src/main/cpp/extensions/openxr_android_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,821 @@
...
+				  nullptr, // next
+				  {}, // fov
+				  {}, // pose
+		  },
+		  {
+				  XR_TYPE_DEPTH_VIEW_ANDROID, // type
+				  nullptr, // next
+				  {}, // fov
+				  {}, // pose
+		  } }, // views[2]
```
dsnopek: I suspect this would make the CI happy:

```suggestion
		{
			{
				XR_TYPE_DEPTH_VIEW_ANDROID, // type
				nullptr, // next
				{}, // fov
				{}, // pose
			},
			{
				XR_TYPE_DEPTH_VIEW_ANDROID, // type
				nullptr, // next
				{}, // fov
				{}, // pose
			}
		}, // views[2]
```

dsnopek (approved):
Thanks!

The code looks great to me (aside from the code style issues CI pointed out, but should be easy to fix). And the sample is working fine in my testing on Samsung Galaxy XR

plugin/src/main/cpp/extensions/openxr_android_environment_depth_extension_wrapper.cpp
dhoverb: Hmm, I ran clang-format and that file did not change.
I'll manually make this change if CI fails again.

dhoverb: Apparently I had spaces here, and clang-format wasn't converting to tabs.
Anyway, I copied your suggestion in the next patchset.  Hopefully it'll pass this time.

dhoverb: Looks like it's passing now.

samples/androidxr-depth-texture-sample/project.godot
```diff
@@ -0,0 +1,99 @@
...
+"value": Projection(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1)
+}
+META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_LEFT={
+"type": "mat4",
+"value": Projection(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1)
+}
+META_ENVIRONMENT_DEPTH_TO_CAMERA_PROJECTION_RIGHT={
+"type": "mat4",
+"value": Projection(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1)
+}
```
m4gr3d, line 64: @dsnopek @dhoverb I thought those globals were only added when `xr/openxr/extensions/meta/environment_depth` is enabled.

dsnopek: Ah, yeah, they should be, These might have just gotten left over from before that change?

plugin/src/main/cpp/extensions/openxr_android_environment_depth_extension_wrapper.cpp
```diff
@@ -0,0 +1,822 @@
...
+	already_setup_global_uniforms = true;
+
+	create_shader_global_uniform(ANDROID_ENVIRONMENT_DEPTH_AVAILABLE_NAME, RenderingServer::GLOBAL_VAR_TYPE_BOOL, false, rs, project_settings, is_editor);
+	create_shader_global_uniform(ANDROID_ENVIRONMENT_DEPTH_TEXTURE_NAME, RenderingServer::GLOBAL_VAR_TYPE_SAMPLER2DARRAY, Variant(), rs, project_settings, is_editor);
+	create_shader_global_uniform(ANDROID_ENVIRONMENT_DEPTH_RESOLUTION_NAME, RenderingServer::GLOBAL_VAR_TYPE_INT, 0, rs, project_settings, is_editor);
+	create_shader_global_uniform(ANDROID_ENVIRONMENT_DEPTH_TANFOV_LEFT_NAME, RenderingServer::GLOBAL_VAR_TYPE_VEC4, Vector4(), rs, project_settings, is_editor);
+	create_shader_global_uniform(ANDROID_ENVIRONMENT_DEPTH_TANFOV_RIGHT_NAME, RenderingServer::GLOBAL_VAR_TYPE_VEC4, Vector4(), rs, project_settings, is_editor);
+	create_shader_global_uniform(ANDROID_ENVIRONMENT_DEPTH_CAMERA_TO_WORLD_LEFT_NAME, RenderingServer::GLOBAL_VAR_TYPE_MAT4, Transform3D(), rs, project_settings, is_editor);
+	create_shader_global_uniform(ANDROID_ENVIRONMENT_DEPTH_CAMERA_TO_WORLD_RIGHT_NAME, RenderingServer::GLOBAL_VAR_TYPE_MAT4, Transform3D(), rs, project_settings, is_editor);
+}
```
m4gr3d, line 631: @dhoverb This should follow the approach in https://github.com/GodotVR/godot_openxr_vendors/pull/377 and only create the shader global uniforms when the project setting is enabled.

cc @dsnopek

samples/androidxr-depth-texture-sample/project.godot
dhoverb: I remember these being created regardless of the project setting.

This still seems to be the case looking at the code here:
https://github.com/GodotVR/godot_openxr_vendors/blob/master/plugin/src/main/cpp/register_types.cpp#L412
(which calls https://github.com/GodotVR/godot_openxr_vendors/blob/master/plugin/src/main/cpp/extensions/openxr_android_environment_depth_extension_wrapper.cpp#L605)

plugin/src/main/cpp/extensions/openxr_android_environment_depth_extension_wrapper.cpp
dhoverb: @m4gr3d I don't see the changes at tip

[377](https://github.com/GodotVR/godot_openxr_vendors/pull/377/files#diff-3a1e1566edafa0c777d24b9121ae2c50836eba2ed25ed55949f21e4b7a5e574aR556)
[tip](https://github.com/GodotVR/godot_openxr_vendors/blob/master/plugin/src/main/cpp/extensions/openxr_android_environment_depth_extension_wrapper.cpp#L605)

Also, I think the conditionals should swap places, like:

```
if (enabled) {
  if (already_setup_global_uniforms) {
    already_setup_global_uniforms = false;
    // ... undo
  }
  return;
}
```
Otherwise you could re-setup the globals after just undoing them.

samples/androidxr-depth-texture-sample/project.godot
dsnopek: Yeah, I think I messed up on https://github.com/GodotVR/godot_openxr_vendors/pull/377

I'll make a PR to fix both extensions

plugin/src/main/cpp/extensions/openxr_android_environment_depth_extension_wrapper.cpp
m4gr3d: > @m4gr3d I don't see the changes at tip
> 
> [377](https://github.com/GodotVR/godot_openxr_vendors/pull/377/files#diff-3a1e1566edafa0c777d24b9121ae2c50836eba2ed25ed55949f21e4b7a5e574aR556) [tip](https://github.com/GodotVR/godot_openxr_vendors/blob/master/plugin/src/main/cpp/extensions/openxr_android_environment_depth_extension_wrapper.cpp#L605)
> 

[377](https://github.com/GodotVR/godot_openxr_vendors/pull/377/files#diff-3a1e1566edafa0c777d24b9121ae2c50836eba2ed25ed55949f21e4b7a5e574aR556) is pointing to the logic in the Meta version of the depth extension wrapper. The Android version doesn't have that logic so my suggestion was to add it to the Android depth extension wrapper.


> Also, I think the conditionals should swap places, like:
> 
> ```
> if (enabled) {
>   if (already_setup_global_uniforms) {
>     already_setup_global_uniforms = false;
>     // ... undo
>   }
>   return;
> }
> ```
> 
> Otherwise you could re-setup the globals after just undoing them.

Yeah I noticed the same thing. @dsnopek is looking into it!

samples/androidxr-depth-texture-sample/project.godot
dsnopek: I think this will fix it: https://github.com/GodotVR/godot_openxr_vendors/pull/410

dhoverb: Looks good to me.

PR #406, opened 2025-12-04
Add support for XR_ANDROID_face_tracking

ashwinvbs:
Implement https://developer.android.com/develop/xr/openxr/extensions/XR_ANDROID_face_tracking.

Modify meta-body-tracking application to be a sample app for Android face tracking.


plugin/src/main/cpp/extensions/openxr_android_face_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	// We currently ignore the following members of face_state.
+	// faceTrackingState
+	// sampleTime
+	// regionConfidences
+
+	// Rest of this function deals with unpacking `parameters` into Godot XRFaceTracker structure
+
+	// Map Meta weights to Godot weights.
```
dsnopek:
```suggestion
	// Map Android XR weights to Godot weights.
```

samples/meta-body-tracking-sample/export_presets.cfg
```diff
@@ -248,3 +248,255 @@ khronos_xr_features/htc/hand_tracking=0
...
+gradle_build/export_format=0
+gradle_build/min_sdk=""
+gradle_build/target_sdk=""
+architectures/armeabi-v7a=false
+architectures/arm64-v8a=true
+architectures/x86=false
+architectures/x86_64=false
+version/code=1
+version/name="1.0"
+package/unique_name="com.godotopenxrvendors.androidxrbodytrackingsample"
```
dsnopek: This should have the same unique name as the Meta Quest preset

(Now that this sample supports multiple vendors body tracking extensions, we should probably rename the whole directory to be "body-tracking-sample", without the "meta-", but you don't need to do that part in this PR - we can do that in a follow-up)

samples/meta-body-tracking-sample/export_presets.cfg
```diff
@@ -248,3 +248,255 @@ khronos_xr_features/htc/hand_tracking=0
 khronos_xr_features/htc/eye_tracking=0
 pico_xr_features/eye_tracking=0
 magicleap_xr_features/hand_tracking=0
+
+[preset.1]
+
+name="Android XR"
+platform="Android"
+runnable=true
```
dsnopek: We've standardized on having the Meta Quest preset as the sole runnable Android preset in the repo, because currently it's the most popular AOSP-compatible headset

```suggestion
runnable=false
```

plugin/src/main/cpp/extensions/openxr_android_face_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+OpenXRAndroidFaceTrackingExtensionWrapper::CalibrationState OpenXRAndroidFaceTrackingExtensionWrapper::get_face_calibration_state() const {
+	return calibration_state;
+}
+
+void OpenXRAndroidFaceTrackingExtensionWrapper::_on_process() {
+	if (!available && face_tracking_properties.supportsFaceTracking) {
+		MainLoop *main_loop = Engine::get_singleton()->get_main_loop();
+		if (main_loop && !main_loop->is_connected("on_request_permissions_result", on_request_permissions_result_callable)) {
+			main_loop->connect("on_request_permissions_result", on_request_permissions_result_callable);
```
dsnopek: Checking this every frame is not ideal, and shouldn't be necessary.

Can you do this in some one-off place? The `Engine` singleton is available pretty early

plugin/src/main/cpp/extensions/openxr_android_face_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,402 @@
...
+		nullptr, // next
+	};
+
+	XrResult result = xrCreateFaceTrackerANDROID(SESSION, &create_info, &face_tracker);
+	if (result != XR_SUCCESS) {
+		UtilityFunctions::printerr("OpenXR: Failed to create face tracker; ", get_openxr_api()->get_error_string(result));
+		available = false;
+		face_tracker = XR_NULL_HANDLE;
+	}
+}
```
dsnopek: Given that this is the same code as in `_on_request_permissions_result()`, maybe this should be moved to its own method?

samples/meta-body-tracking-sample/export_presets.cfg
BastiaanOlij: I long for the day we can have just a single profile for all AOSP devices, but I doubt that will ever happen.

plugin/src/main/cpp/extensions/openxr_android_face_tracking_extension_wrapper.cpp
BastiaanOlij: On start session would be a good place, with unregistering it on stop session (though total overkill that one). 

While the OpenXR instance is created very early on, the session isn't started until most of Godot is initialised and you should have access to the engine singleton at that point in time.

plugin/src/main/cpp/extensions/openxr_android_face_tracking_extension_wrapper.cpp
ashwinvbs: Oops :D
Fixed it

samples/meta-body-tracking-sample/export_presets.cfg
ashwinvbs: Done

samples/meta-body-tracking-sample/export_presets.cfg
ashwinvbs: Done

plugin/src/main/cpp/extensions/openxr_android_face_tracking_extension_wrapper.cpp
ashwinvbs: Ive moved it to `on_state_ready`. It seems like the most appropriate place for this.

The sequence of calls are
```ExtensionWrapper::on_session_created
OS::set_main_loop
ExtensionWrapper::on_state_idle
ExtensionWrapper::on_state_ready
ExtensionWrapper::on_process
```

plugin/src/main/cpp/extensions/openxr_android_face_tracking_extension_wrapper.cpp
ashwinvbs: Done

plugin/src/main/cpp/extensions/openxr_android_face_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,399 @@
...
+	if (!xr_face_tracker.is_null()) {
+		XRServer *xr_server = XRServer::get_singleton();
+		if (xr_server) {
+			xr_server->remove_tracker(xr_face_tracker);
+		}
+	}
+}
+
+bool OpenXRAndroidFaceTrackingExtensionWrapper::_initialize_openxr_android_face_tracking_extension() {
+	GDEXTENSION_INIT_XR_FUNC_V(xrCreateFaceTrackerANDROID);
```
ashwinvbs, line 147: I found a logic bug.

`available` is set to false if `xrCreateFaceTrackerANDROID` function cannot be initialized here.
But there is a path where we end up calling this function through the `on_request_permissions_result` event.

Please do not merge until I push a fix for this.

ashwinvbs: done..

PR #410, opened 2025-12-10
Only add shader globals when the environment depth extensions are enabled

dsnopek:
It looks like I didn't get https://github.com/GodotVR/godot_openxr_vendors/pull/377 quite right, and the Meta depth extension wrapper can still add the global uniforms

This PR fixes the Meta extension wrapper, and then applies the same logic to the Android one as well


m4gr3d (approved):
Looks good!

PR #412, opened 2025-12-12
Implement XR_ANDROID_scene_meshing

dhoverb:
Implements [XR_ANDROID_scene_meshing](https://developer.android.com/develop/xr/openxr/extensions/XR_ANDROID_scene_meshing)


doc_classes/OpenXRAndroidSceneMeshing.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidSceneMeshing" inherits="RefCounted" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		An object for retrieving submesh data.
```
dsnopek:
```suggestion
		An object for retrieving submesh data from [code]XR_ANDROID_scene_meshing[/code] extension.
```

doc_classes/OpenXRAndroidSceneSubmeshData.xml
```diff
@@ -0,0 +1,107 @@
...
+				The [Array] contains data at the [constant Mesh.ARRAY_INDEX] index.
+				The [Array] may contain data at the [constant Mesh.ARRAY_NORMAL] index if [code]enable_normals[/code] was [b]true[/b] in [method OpenXRAndroidSceneMeshing.initialize].
+				The [Array] may contain data at the [constant Mesh.ARRAY_CUSTOM0] index if [code]semantic_label_set[/code] was [b]true[/b] in [method OpenXRAndroidSceneMeshing.initialize].
+			</description>
+		</method>
+		<method name="get_indexed_vertex_semantics" qualifiers="const">
+			<return type="Array" />
+			<description>
+				An [Array] of [enum SemanticLabel]s, which is identical to indexing the [Array] from [method get_indexed_arrays] at [constant Mesh.ARRAY_CUSTOM0].
+				[b]NOTE[/b] This [Array] is always empty when [method OpenXRAndroidSceneMeshing.initialize] was called with [constant OpenXRAndroidSceneMeshing.SEMANTIC_LABEL_SET_NONE].
```
dsnopek:
```suggestion
				[b]NOTE:[/b] This [Array] is always empty when [method OpenXRAndroidSceneMeshing.initialize] was called with [constant OpenXRAndroidSceneMeshing.SEMANTIC_LABEL_SET_NONE].
```

doc_classes/OpenXRAndroidSceneSubmeshData.xml
```diff
@@ -0,0 +1,107 @@
...
+			<return type="int" enum="OpenXRAndroidSceneSubmeshData.UpdateState" />
+			<description>
+				Get the state of the mesh.
+			</description>
+		</method>
+		<method name="get_vertex_semantics" qualifiers="const">
+			<return type="Array" />
+			<description>
+				An [Array] of [enum SemanticLabel]s, which is identical to indexing the [Array] from [method get_arrays] at [constant Mesh.ARRAY_CUSTOM0].
+				[b]NOTE[/b] This [Array] is always empty when [method OpenXRAndroidSceneMeshing.initialize] was called with [constant OpenXRAndroidSceneMeshing.SEMANTIC_LABEL_SET_NONE].
```
dsnopek:
```suggestion
				[b]NOTE:[/b] This [Array] is always empty when [method OpenXRAndroidSceneMeshing.initialize] was called with [constant OpenXRAndroidSceneMeshing.SEMANTIC_LABEL_SET_NONE].
```

doc_classes/OpenXRAndroidSceneSubmeshData.xml
```diff
@@ -0,0 +1,107 @@
...
+			</description>
+		</method>
+	</methods>
+	<constants>
+		<constant name="UPDATE_STATE_CREATED" value="0" enum="UpdateState">
+			The submesh was just created.
+		</constant>
+		<constant name="UPDATE_STATE_UNCHANGED" value="1" enum="UpdateState">
+			The submesh is unchanged since the last call to [method OpenXRAndroidSceneMeshing.get_submesh_data].
+			[b]NOTE[/b] [method get_transform] may have changed; the array data (vertices, indices, etc) are not changed.
```
dsnopek:
```suggestion
			[b]NOTE:[/b] [method get_transform] may have changed; the array data (vertices, indices, etc) are not changed.
```

dsnopek (approved):
Thanks! The code looks good to me, and the sample works great on Samsung Galaxy XR in my testing.

I have only a handful of docs nitpicks

doc_classes/OpenXRAndroidSceneMeshing.xml
dhoverb: Done.

doc_classes/OpenXRAndroidSceneSubmeshData.xml
dhoverb: Done.

doc_classes/OpenXRAndroidSceneSubmeshData.xml
dhoverb: Done.

doc_classes/OpenXRAndroidSceneSubmeshData.xml
dhoverb: Done.

plugin/src/main/cpp/extensions/openxr_android_scene_meshing_extension_wrapper.cpp
```diff
@@ -0,0 +1,172 @@
...
+	XrResult result = xrEnumerateSupportedSemanticLabelSetsANDROID(xr_instance, xr_system_id, 0, &supported_semantic_label_sets_output_count, nullptr);
+	if (result != XR_SUCCESS) {
+		UtilityFunctions::printerr("OpenXR: Failed to enumerate semantic label sets; ", get_openxr_api()->get_error_string(result));
+		available = false;
+		return ret;
+	}
+
+	if (supported_semantic_label_sets_output_count == 0) {
+		WARN_PRINT("OpenXR: enumerate semantic label sets returned zero");
+		available = false;
```
m4gr3d: So here and below, if the count is `0` is that considered an error, and is that why `available` is reset to `false`?

plugin/src/main/cpp/classes/openxr_android_scene_submesh_data.cpp
```diff
@@ -0,0 +1,384 @@
...
+namespace {
+constexpr bool IS_INDEXED_ARRAY = true;
+constexpr bool COPY_NORMALS = true;
+constexpr bool COPY_SEMANTICS = true;
+
+_FORCE_INLINE_ void xr_to_godot(Vector3 &r_godot_vector3, const XrVector3f &p_xr_vector3) {
+	r_godot_vector3.x = p_xr_vector3.x;
+	r_godot_vector3.y = p_xr_vector3.y;
+	r_godot_vector3.z = p_xr_vector3.z;
+}
```
m4gr3d: Let's match the pattern in https://github.com/GodotVR/godot_openxr_vendors/pull/416/changes#diff-f7ebc19f640578d041567d317e657154ccb47f48e6f1111efc818c54d456fd0a and move this method to `util.h`

plugin/src/main/cpp/classes/openxr_android_scene_submesh_data.cpp
```diff
@@ -0,0 +1,384 @@
...
+uint64_t OpenXRAndroidSceneSubmeshData::get_update_idx() const {
+	return update_idx;
+}
+
+void OpenXRAndroidSceneSubmeshData::update(UpdateState p_update_state, uint64_t p_update_idx, const StringName &p_submesh_uuid, const XrSceneSubmeshStateANDROID &p_submesh_state, XrSceneMeshSnapshotANDROID p_snapshot, bool p_enable_normals, OpenXRAndroidSceneMeshing::SemanticLabelSet p_semantic_label_set) {
+	if (!_update(p_update_state, p_update_idx, p_submesh_uuid, p_submesh_state, p_snapshot, p_enable_normals, p_semantic_label_set)) {
+		WARN_PRINT("Update submesh data failed; assuming UpdateState is unchanged");
+		update_state = UPDATE_STATE_UNCHANGED;
+	}
+}
```
m4gr3d, line 169: Is there a reason for splitting the `update` methods?

plugin/src/main/cpp/extensions/openxr_android_scene_meshing_extension_wrapper.cpp
dsnopek: Hm, yeah, this does seem sort of odd. I could understand the count being zero (and the rest below) meaning that we can't get the semantic labels, but setting `available = false` means we can't do anything at all if can't get them

plugin/src/main/cpp/classes/openxr_android_scene_submesh_data.cpp
```diff
@@ -0,0 +1,384 @@
...
+		0, // vertexCountOutput
+		nullptr, // vertexPositions
+		nullptr, // vertexNormals
+		nullptr, // vertexSemantics
+		0, // indexCapacityInput
+		0, // indexCountOutput
+		nullptr, // indices
+	};
+
+	XrResult result = wrapper->xrGetSubmeshDataANDROID(p_snapshot, 1, &xrsubmesh_data);
```
dsnopek: This doesn't appear to be checking if the extension is actually enabled before calling this function. We should exit early and show an error if it's not enabled

plugin/src/main/cpp/extensions/openxr_android_scene_meshing_extension_wrapper.cpp
devloglogan: I've removed this line! I've also renamed `available` to `androidxr_scene_meshing_ext` as that is more in line with our typical convention.

plugin/src/main/cpp/classes/openxr_android_scene_submesh_data.cpp
devloglogan: Suggestion applied!

plugin/src/main/cpp/classes/openxr_android_scene_submesh_data.cpp
devloglogan: I've removed `_update()` and put the logic here instead. There's a tiny amount of code duplication (the two blocks where this warning gets printed) but I prefer it this way.

plugin/src/main/cpp/classes/openxr_android_scene_submesh_data.cpp
devloglogan: Check added!

devloglogan:
@m4gr3d @dsnopek I've made some adjustments per your reviews and some other minor tweaks. When running the sample, after a little while performance would start to be pretty bad. For the time being I've added an `OS.delay_msec()` call that seems to fix that. I will rework the sample a bit in a future PR!

plugin/src/main/cpp/extensions/openxr_android_scene_meshing_extension_wrapper.cpp
dsnopek: > I've removed this line!

There's more than this one line - through out this function, it's setting `androidxr_scene_meshing_ext = false` if there is an error, which doesn't seem quite right

plugin/src/main/cpp/extensions/openxr_android_scene_meshing_extension_wrapper.cpp
```diff
@@ -0,0 +1,181 @@
...
+		return ret;
+	}
+
+	uint32_t supported_semantic_label_sets_output_count = 0;
+	XrInstance xr_instance = (XrInstance)get_openxr_api()->get_instance();
+	XrSystemId xr_system_id = (XrSystemId)get_openxr_api()->get_system_id();
+	XrResult result = xrEnumerateSupportedSemanticLabelSetsANDROID(xr_instance, xr_system_id, 0, &supported_semantic_label_sets_output_count, nullptr);
+	if (result != XR_SUCCESS) {
+		UtilityFunctions::printerr("OpenXR: Failed to enumerate semantic label sets; ", get_openxr_api()->get_error_string(result));
+		androidxr_scene_meshing_ext = false;
```
dsnopek: Here's another - seems weird to act like the whole extension is disabled if we can't get the semantic labels

plugin/src/main/cpp/extensions/openxr_android_scene_meshing_extension_wrapper.cpp
```diff
@@ -0,0 +1,181 @@
...
+		WARN_PRINT("OpenXR: enumerate semantic label sets returned zero");
+		return ret;
+	}
+
+	LocalVector<XrSceneMeshSemanticLabelSetANDROID> semantic_label_sets;
+	semantic_label_sets.resize(supported_semantic_label_sets_output_count);
+	result = xrEnumerateSupportedSemanticLabelSetsANDROID(xr_instance, xr_system_id, supported_semantic_label_sets_output_count, &supported_semantic_label_sets_output_count, semantic_label_sets.ptr());
+	if (result != XR_SUCCESS) {
+		UtilityFunctions::printerr("OpenXR: Failed to enumerate semantic label sets; ", get_openxr_api()->get_error_string(result));
+		androidxr_scene_meshing_ext = false;
```
dsnopek: Here too

plugin/src/main/cpp/extensions/openxr_android_scene_meshing_extension_wrapper.cpp
```diff
@@ -0,0 +1,181 @@
...
+		return ret;
+	}
+
+	if (semantic_label_sets.size() != supported_semantic_label_sets_output_count) {
+		WARN_PRINT("OpenXR: enumerate semantic label sets returned a different count");
+		if (semantic_label_sets.size() < supported_semantic_label_sets_output_count) {
+			UtilityFunctions::printerr("OpenXR: somehow received more semantic label sets on the second query; ", get_openxr_api()->get_error_string(result));
+			androidxr_scene_meshing_ext = false;
```
dsnopek: Here too

plugin/src/main/cpp/extensions/openxr_android_scene_meshing_extension_wrapper.cpp
```diff
@@ -0,0 +1,181 @@
...
+			return ret;
+		}
+
+		if (supported_semantic_label_sets_output_count == 0) {
+			UtilityFunctions::printerr("OpenXR: second enumerate semantic label sets returned zero");
+			androidxr_scene_meshing_ext = false;
```
dsnopek: Here too

dsnopek (commented):
Thanks!

I tested on the Samsung Galaxy XR, and performance was good throughout - staying at 90 fps.

The only remaining note I have is that there's a bunch of other times in `get_supported_semantic_label_sets()` where it's doing `androidxr_scene_meshing_ext = false` which doesn't seem quite right. I've pointed them out below

devloglogan:
@dsnopek I've removed the other ones as well! I mistakenly thought from your and Fredia's initial comments that you only took issue with the one where it wasn't an OpenXR error, removing all of them does make the most sense to me.

plugin/src/main/cpp/classes/openxr_android_scene_meshing.cpp
```diff
@@ -0,0 +1,263 @@
...
+			return false;
+	}
+
+	XrSceneMeshingTrackerCreateInfoANDROID create_info = {
+		XR_TYPE_SCENE_MESHING_TRACKER_CREATE_INFO_ANDROID, // type
+		nullptr, // next
+		xrsemantic_label_set, // semanticLabelSet
+		(XrBool32)(p_enable_normals ? XR_TRUE : XR_FALSE) // enableNormals
+	};
+	XrResult result = wrapper->xrCreateSceneMeshingTrackerANDROID((XrSession)wrapper->get_openxr_api()->get_session(), &create_info, &tracker);
```
m4gr3d, line 82:
```suggestion
	XrResult result = wrapper->xrCreateSceneMeshingTrackerANDROID(SESSION, &create_info, &tracker);
```

m4gr3d: nvm, that macro is only usable by `OpenXRExtensionWrapperExtension` instances.

m4gr3d (approved):

PR #415, opened 2025-12-18
Fix Android XR manifest entries for face tracking

ashwinvbs:
Face tracking permission was incorrectly setup to depend on face tracking feature flag when generating manifest.

This PR fixes this issue and removes the face tracking feature flag as it is not necessary.


dsnopek (approved):
Thanks!

The code looks good to me. I just tested it on Samsung Galaxy XR with the meta-body-tracking-sample, and it seems to work fine :-)

PR #416, opened 2025-12-19
Add support for XR_ANDROID_eye_tracking extension

ashwinvbs:
Uses XRControllerTracker to track eye positions.
Eye input precision (fine/coarse) and blink are hooked up as button inputs in the controller.

A project level setting `xr/openxr/extensions/androidxr/eye_tracking/require_high_precision` allows game developers to choose between precision levels.

If set to false, the tracker will use best available precision based on permissions granted by the user.
If set to true, only the fine eye tracking info will be requested and no tracking information will be available if the fine eye tracking query fails or if the corresponding permission is not granted by the user.


plugin/src/main/cpp/export/android_xr_export_plugin.cpp
```diff
@@ -270,11 +270,20 @@ String AndroidXREditorExportPlugin::_get_android_manifest_element_contents(const
...
 
-		int eye_tracking_value = _get_int_option("android_xr_features/eye_tracking", EYE_TRACKING_OPTIONAL_VALUE);
-		if (eye_tracking_value == EYE_TRACKING_REQUIRED_VALUE) {
+	if (eye_gaze_interaction_enabled || eye_tracking_enabled) {
+		// XR_EXT_eye_gaze_interaction always requires EYE_TRACKING_FINE.
+		contents += "    <uses-permission android:name=\"android.permission.EYE_TRACKING_FINE\" />\n";
+	}
+	if (eye_tracking_enabled && !eye_tracking_fine_precision_required) {
+		// Request coarse tracking permission only if developer makes fine precision optional.
+		contents += "    <uses-permission android:name=\"android.permission.EYE_TRACKING_COARSE\" />\n";
```
m4gr3d, line 284: If `eye_tracking_enabled` and `eye_tracking_fine_precision_required` is false, both permissions are added; is that the intent?

plugin/src/main/cpp/include/util.h
```diff
@@ -127,4 +129,5 @@ struct XrMatrix4x4f;
 namespace OpenXRUtilities {
 godot::StringName uuid_to_string_name(const XrUuid &p_uuid);
 void xrMatrix4x4f_to_godot_projection(XrMatrix4x4f *m, godot::Projection &p);
+godot::Transform3D xr_to_godot(const XrPosef &pose);
```
m4gr3d: Let's be explicit with the function name 

```suggestion
godot::Transform3D xrPosef_to_godot_transform3d(const XrPosef &pose);
```

plugin/src/main/cpp/extensions/openxr_android_eye_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,251 @@
...
+	};
+
+	if (have_fine_tracking_permission || have_fine_tracking_permission_requres_refresh) {
+		precision_fine = true;
+		result = xrGetFineTrackingEyesInfoANDROID(eye_tracker, &get_info, &eyes);
+		have_fine_tracking_permission_requres_refresh = false;
+		have_fine_tracking_permission = (result != XR_ERROR_PERMISSION_INSUFFICIENT);
+	}
+
+	if (!have_fine_tracking_permission && !project_requires_fine_tracking) {
```
m4gr3d: > If set to true, only the fine eye tracking info will be requested and no tracking information will be available if the fine eye tracking query fails or if the corresponding permission is not granted by the user.

Are there scenarios where fine tracking is required? I wonder if we should always default to using coarse tracking if we don't get the fine tracking permission regardless of whether the project requires fine tracking.

plugin/src/main/cpp/export/android_xr_export_plugin.cpp
ashwinvbs: yes.
If eye_tracking_enabled is true and eye_tracking_fine_precision_required is false, the extension requests for both types of data and uses the one user has granted permission for.

BastiaanOlij: Discussing this with the group, we think it should be either or, so if fine is requested, and the user doesn't provide permission, we just don't have permission.
This puts a bit more on the developer to choose whether to enable fine permissions or not, but we feel only a few developers will need to enable this, and if they need this then a user not giving permission makes it impossible to do what needs to be done.

We probably should make this clear in documentation somewhere and advise developers not to enable fine unless they need to.

ashwinvbs: Due to inherent complexity and limited use cases, I have removed fine eye tracking from the scope of this implementation.
Based on user feedback, fine eye tracking can be added at a later point.

plugin/src/main/cpp/include/util.h
ashwinvbs: Done.

plugin/src/main/cpp/extensions/openxr_android_eye_tracking_extension_wrapper.cpp
ashwinvbs: Due to inherent complexity and limited use cases, I have removed fine eye tracking from the scope of this implementation.

ashwinvbs:
Due to inherent complexity and limited use cases, I have removed fine eye tracking from the scope of this implementation.
Based on user feedback, fine eye tracking can be added at a later point.

The PR is ready for review.

doc_classes/OpenXRAndroidEyeTrackingExtensionWrapper.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidEyeTrackingExtensionWrapper" inherits="OpenXRExtensionWrapperExtension" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Wraps the [code]XR_ANDROID_eye_tracking[/code] extension.
+	</brief_description>
+	<description>
+		Wraps the [code]XR_ANDROID_eye_tracking[/code] extension.
+		This class is not meant to be used directly.
+		To use coarse eye tracking data, instantiate [XRController3D] with [code]"/user/eye_tracker_android/left"[/code] to track left eye or [code]"/user/eye_tracker_android/right"[/code] to track right eye.
```
dsnopek: The bit about "coarse eye tracking" probably doesn't make sense any more:

```suggestion
		Instantiate a [XRController3D] with [code]"/user/eye_tracker_android/left"[/code] to track left eye or [code]"/user/eye_tracker_android/right"[/code] to track right eye.
```

doc_classes/OpenXRAndroidEyeTrackingExtensionWrapper.xml
```diff
@@ -0,0 +1,15 @@
...
+		Blink state is exposed as a input button named [code]"button"[/code]. On eyelid close, the button indicates pressed and on eyelid open, the button indicates released.
```
dsnopek:
```suggestion
		Blink state is exposed as a input button named [code]"blink"[/code]. On eyelid close, the button indicates pressed and on eyelid open, the button indicates released.
```

plugin/src/main/cpp/extensions/openxr_android_eye_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,222 @@
...
+		result[ext.key] = (Variant)value;
+	}
+	return result;
+}
+
+uint64_t OpenXRAndroidEyeTrackingExtensionWrapper::_set_system_properties_and_get_next_pointer(void *p_next_pointer) {
+	eye_tracking_properties.type = XR_TYPE_SYSTEM_EYE_TRACKING_PROPERTIES_ANDROID;
+	eye_tracking_properties.next = p_next_pointer;
+	eye_tracking_properties.supportsEyeTracking = XR_FALSE;
+	return reinterpret_cast<uint64_t>(&eye_tracking_properties);
```
dsnopek: This should be wrapped in `if (available) { ... }`:

```suggestion
	if (available) {
		eye_tracking_properties.type = XR_TYPE_SYSTEM_EYE_TRACKING_PROPERTIES_ANDROID;
		eye_tracking_properties.next = p_next_pointer;
		eye_tracking_properties.supportsEyeTracking = XR_FALSE;
		return reinterpret_cast<uint64_t>(&eye_tracking_properties);
	}
	return nullptr;
```

plugin/src/main/cpp/extensions/openxr_android_eye_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+void OpenXRAndroidEyeTrackingExtensionWrapper::_on_instance_created(uint64_t p_instance) {
+	if (!_initialize_openxr_android_eye_tracking_extension()) {
+		UtilityFunctions::print("Failed to initialize eye tracking extension");
+		available = false;
+	}
```
dsnopek: Here too:

```suggestion
	if (available) {
		if (!_initialize_openxr_android_eye_tracking_extension()) {
			UtilityFunctions::print("Failed to initialize eye tracking extension");
			available = false;
		}
	}
```

plugin/src/main/cpp/extensions/openxr_android_eye_tracking_extension_wrapper.cpp
```diff
@@ -0,0 +1,222 @@
...
+void OpenXRAndroidEyeTrackingExtensionWrapper::_on_session_created(uint64_t instance) {
+	if (!available || !eye_tracking_properties.supportsEyeTracking) {
+		return;
+	}
+
+	_try_create_eye_tracker();
+}
+
+void OpenXRAndroidEyeTrackingExtensionWrapper::_on_request_permissions_result(const String &p_permission, bool p_granted) {
+	if (p_permission != "android.permission.EYE_TRACKING_COARSE" || !p_granted) {
```
dsnopek: Here too:

```suggestion
	if (!available || p_permission != "android.permission.EYE_TRACKING_COARSE" || !p_granted) {
```

dsnopek (commented):
Thanks!

This worked great in my testing on the Samsung Galaxy XR. I only have a new notes below

samples/androidxr-eye-tracking-sample/main.tscn
```diff
@@ -0,0 +1,30 @@
...
+[node name="XRCamera3D" type="XRCamera3D" parent="XROrigin3D"]
+
+[node name="Hud" type="Label3D" parent="XROrigin3D"]
+script = ExtResource("3_lquwl")
+
+[node name="LeftEyePose" type="XRController3D" parent="XROrigin3D"]
+tracker = &"/user/eye_tracker_android/left"
+show_when_tracked = true
+
+[node name="CSGSphere3D" type="CSGSphere3D" parent="XROrigin3D/LeftEyePose"]
```
dsnopek: This is fine, because using `CSGSphere3D` doesn't really hurt anything (there's a performance difference but it won't matter in this sample), but there also isn't any need for this to be using the CSG node. I think this would work just fine using a `MeshInstance3D` with a `SphereMesh`

doc_classes/OpenXRAndroidEyeTrackingExtensionWrapper.xml
ashwinvbs: Done

doc_classes/OpenXRAndroidEyeTrackingExtensionWrapper.xml
ashwinvbs: Done

plugin/src/main/cpp/extensions/openxr_android_eye_tracking_extension_wrapper.cpp
ashwinvbs: Done

plugin/src/main/cpp/extensions/openxr_android_eye_tracking_extension_wrapper.cpp
ashwinvbs: Done

plugin/src/main/cpp/extensions/openxr_android_eye_tracking_extension_wrapper.cpp
ashwinvbs: Done

samples/androidxr-eye-tracking-sample/main.tscn
ashwinvbs: Done

dsnopek (approved):
Thanks! Looks good to me :-)

m4gr3d (approved):
Looks good!

PR #418, opened 2025-12-22
Implement XR_EXTX2_stationary_reference_space extension

devloglogan:
This PR builds on top of https://github.com/GodotVR/godot_openxr_vendors/pull/407 and will remain in draft until that PR is merged, at which point this could be merged into the 5.x branch.

This is another extension that relies on Meta OpenXR SDK headers, it's not currently in the OpenXR spec. When merged, it should be added to the issue used for tracking these: https://github.com/GodotVR/godot_openxr_vendors/issues/304

**Note:** In order for stationary reference space to work, you currently have to enable experimental features on your headset using `adb`:
```
adb shell setprop debug.oculus.experimentalEnabled 1
```

~~I've enabled the use of the stationary reference space in the Meta Scene Sample.~~ Seeing that the extension is not Meta-specific, I've made the project setting vendor neutral. I only added the manifest entry for it (enabling experimental features) in `meta_export_plugin.cpp`, I'm not aware if any other runtimes support it in this experimental stage.

**EDIT 2026-01-26:** I've removed enabling stationary reference space from the scene sample.

### DIFF
--- a/plugin/src/main/cpp/export/meta_export_plugin.cpp
+++ b/plugin/src/main/cpp/export/meta_export_plugin.cpp
@@ -434,6 +434,11 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
 	if ((bool)project_settings->get_setting_with_override("xr/openxr/extensions/meta/boundary_visibility")) {
 		contents += "    <uses-permission android:name=\"com.oculus.permission.BOUNDARY_VISIBILITY\" />\n";
 	}
+
+	// Check for stationary reference space
+	if ((bool)project_settings->get_setting_with_override("xr/openxr/extensions/stationary_reference_space")) {
+		contents += "    <uses-feature android:name=\"com.oculus.experimental.enabled\" />\n";
+	}
 #endif // META_HEADERS_ENABLED
 
 	// Check for overlay keyboard
--- a/plugin/src/main/cpp/include/extensions/openxr_stationary_reference_space_extension.h
+++ b/plugin/src/main/cpp/include/extensions/openxr_stationary_reference_space_extension.h
@@ -1,0 +1,93 @@
+/**************************************************************************/
+/*  openxr_stationary_reference_space_extension.h                         */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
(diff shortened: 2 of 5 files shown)


plugin/src/main/cpp/include/extensions/openxr_stationary_reference_space_extension.h
```diff
@@ -0,0 +1,91 @@
...
+	OpenXRStationaryReferenceSpaceExtension();
+	~OpenXRStationaryReferenceSpaceExtension();
+
+	godot::Dictionary _get_requested_extensions(uint64_t p_xr_version) override;
+
+	void _on_instance_created(uint64_t p_instance) override;
+	void _on_instance_destroyed() override;
+	void _on_session_created(uint64_t p_instance) override;
+
+	void set_stationary_reference_space_enabled(bool p_enable);
```
dsnopek, line 58: I think we need a way to check if this extension is enabled and the `stationary_space` was successfully created, otherwise developers will have no way to know if enabling the space even did anything. And they may want to do things a little differently if they have the stationary reference and when they don't

devloglogan: I've added an `is_stationary_reference_space_available()` method

plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -434,6 +434,11 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
 	if ((bool)project_settings->get_setting_with_override("xr/openxr/extensions/meta/boundary_visibility")) {
 		contents += "    <uses-permission android:name=\"com.oculus.permission.BOUNDARY_VISIBILITY\" />\n";
 	}
+
+	// Check for stationary reference space
+	if ((bool)project_settings->get_setting_with_override("xr/openxr/extensions/stationary_reference_space")) {
+		contents += "    <uses-feature android:name=\"com.oculus.experimental.enabled\" />\n";
```
m4gr3d, line 440: @devloglogan This feature is also added below when `_get_bool_option("meta_xr_features/use_experimental_features") == true`, so those two blocks should be merged to avoid adding duplicate feature tags in the manifest.

PR #420, opened 2025-12-23
Add Meta Body Tracking doc

devloglogan:
Adds a doc page about Meta body tracking features.

It's worth mentioning that while the body tracking fidelity extension seems to be working fine (I can request low fidelity without issue), I've never actually seen that request be applied by the runtime. Without this extension being in the spec and no sample for it in the Meta OpenXR SDK, I am just assuming we're at the mercy of the runtime and there's nothing more that can be done on our end. But if anyone has ideas on how to test this further, or thinks this extension should be withheld from the doc page for the time being, let me know!


dsnopek (approved):
Thanks! Looks good to me :-)

PR #421, opened 2025-12-23
Add input handling to `viewport_2d_in_3d` and use fixed viewport size

dsnopek:
In our samples, traditionally, we've used controller buttons to toggle settings.

However, on Samsung Galaxy XR, folks may not have controllers at all!

In order to make settings easier in that context, I thought it would be good to have input on our implementation of `viewport_2d_in_3d`, so we can pinch select UI controls there (or use controllers too). So, I stole some code from [my personal version of `viewport_2d_in_3d`](https://gitlab.com/snopek-games/sgxr/-/blob/main/addons/sgxr/ui_layer/ui_layer.gd?ref_type=heads) and some other bits and pieces to implement this. (I didn't bring everything over - I removed the use of `OpenXRCompositionLayer`s and the support for WebXR.)

Also, this PR removes all the stuff that tried to have automatic viewport sizing and replaces it with an property for viewport size.

The previous implementation had some "circular problems": it was grabbing the size of the control it added to the viewport, and then resizing the viewport to that - but the size of the control _depends_ on the size of its parent viewport! I don't think there's a way to truly do automatic sizing without doing it "iteratively" somehow, and that seems like way too much - we can just give it an explicit size and be done with it :-)

### DIFF
--- a/demo/addons/common/view_pivot/view_pivot.tscn
+++ b/demo/addons/common/view_pivot/view_pivot.tscn
@@ -1,0 +1,12 @@
+[gd_scene load_steps=2 format=3 uid="uid://xga880omsrpm"]
+
+[ext_resource type="Script" uid="uid://1nit8ovimkj4" path="res://addons/common/view_pivot/view_pivot.gd" id="1_tpea8"]
+
+[node name="ViewPivot" type="Node3D"]
+script = ExtResource("1_tpea8")
+
+[node name="Timer" type="Timer" parent="."]
+unique_name_in_owner = true
+one_shot = true
+
+[connection signal="timeout" from="Timer" to="." method="_on_timer_timeout"]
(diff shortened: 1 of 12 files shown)


demo/addons/common/view_pivot/view_pivot.tscn
```diff
@@ -0,0 +1,12 @@
+[gd_scene load_steps=2 format=3 uid="uid://xga880omsrpm"]
+
+[ext_resource type="Script" uid="uid://1nit8ovimkj4" path="res://addons/sgxr/view_pivot.gd" id="1_tpea8"]
```
m4gr3d: Is that path correct?

m4gr3d (approved):

demo/addons/common/view_pivot/view_pivot.tscn
dsnopek: Ah, no, it's not! it probably still worked because of the UID. I've fixed it in my latest push

PR #422, opened 2025-12-23
Add support for `XR_ANDROID_light_estimation`

dsnopek:
This adds support for `XR_ANDROID_light_estimation`

It works in a similar way to Meta and Android XR's environment depth: it provides raw access to the data so that developers can use this in their own shaders, as well as provides a node (`OpenXRAndroidLightEstimation`) that tries to do a reasonable default that works more-or-less automatically.

It does this by taking the path to a `DirectionalLight3D` and a `WorldEnvironment` and applies changes to them. In order to automatically support the spherical harmonics data (and eventually cubemaps once that's available), it uses a sky shader to update Godot's radiance map.

The sample uses the changes from https://github.com/GodotVR/godot_openxr_vendors/pull/421 to provide some settings, so developers can play with them and see what they do. This includes switching to a custom shader, so we can demonstrate how that works

~~However, I'm marking this as a DRAFT for now, because:~~

- ~~I'm not sure about any of the math. :-) The result is "plausible" but there's a bunch of stuff I'm unsure about, so I'm going to continue checking and experimenting with it for a bit~~
- ~~I still need to add API docs~~
- ~~This depends on https://github.com/GodotVR/godot_openxr_vendors/pull/421~~
- ~~This depends on https://github.com/GodotVR/godot_openxr_vendors/pull/426~~


demo/addons/common/function_pointer/function_pointer.gd
```diff
@@ -0,0 +1,180 @@
...
+
+func _exit_tree() -> void:
+	if _controller:
+		_controller.input_float_changed.disconnect(_on_controller_input_float_changed)
+		_controller.button_pressed.disconnect(_on_controller_button_pressed)
+		_controller.button_released.disconnect(_on_controller_button_released)
+
+
+func _process(_delta: float) -> void:
+	if enabled:
```
dhoverb: Should we instead use `set_process` when `_process` is dependent on one variable?

For example:

```gdscript
func set_enabled(p_enabled: bool) -> void:
  # ... current code
  set_process(p_enabled)


func _process(_de_delta: float) -> void:
  for layer in _get_ui_layers():
     # ...
```

Though I can see this pattern being harder to read (and not sure about benefits in performance, if anything).

Another suggestion (though subjective), you could reduce one level of indentation with an early return:

```gdscript
if !enabled:
  return
  
for layer in _get_ui_layers():
  # ...
```

demo/addons/common/function_pointer/function_pointer.gd
```diff
@@ -0,0 +1,180 @@
...
+func _get_ui_layers() -> Array:
+	if not is_inside_tree():
+		return []
+
+	var nodes := get_tree().get_nodes_in_group("ui_layer")
+
+	# Sort them closest to furthest.
+	if nodes.size() > 1:
+		var gp := global_position
+		nodes.sort_custom(func(a, b): return gp.distance_squared_to(a.global_position) < gp.distance_squared_to(b.global_position))
```
dhoverb, line 112: This looks like it is sorting from closest to furthest from the origin.  Should it be from the camera? (or better, from the source of the ray cast)

demo/addons/common/function_pointer/function_pointer.tscn
```diff
@@ -0,0 +1,29 @@
...
+[ext_resource type="Script" uid="uid://qe5cwrdl7ri2" path="res://addons/common/function_pointer/function_pointer.gd" id="1_pjipa"]
+
+[sub_resource type="StandardMaterial3D" id="StandardMaterial3D_pjipa"]
+shading_mode = 0
+disable_receive_shadows = true
+
+[sub_resource type="StandardMaterial3D" id="StandardMaterial3D_g1l7f"]
+shading_mode = 0
+albedo_color = Color(0, 0.478431, 1, 1)
+disable_receive_shadows = true
```
dhoverb, line 12: Assuming `shading_mode = 0` means `unshaded`: is this required for unshaded materials?

demo/addons/common/function_pointer/function_pointer.tscn
```diff
@@ -0,0 +1,29 @@
...
+resource_local_to_scene = true
+size = Vector3(0.005, 0.005, 5)
+
+[node name="FunctionPointer" type="Node3D" groups=["function_pointer"]]
+script = ExtResource("1_pjipa")
+select_action = "trigger"
+released_material = SubResource("StandardMaterial3D_pjipa")
+pressed_material = SubResource("StandardMaterial3D_g1l7f")
+
+[node name="MeshInstance3D" type="MeshInstance3D" parent="."]
```
dhoverb: Can these two nodes be merged?  (why are they separate?)

demo/addons/common/function_pointer/function_pointer.tscn
```diff
@@ -0,0 +1,29 @@
...
+unique_name_in_owner = true
+transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, -2.57)
+mesh = SubResource("BoxMesh_6h6mo")
+skeleton = NodePath("../../LeftHand")
```
dhoverb: Should this be deleted?

demo/addons/common/function_pointer/function_pointer.gd
dhoverb, line 1: subjective: this name is a little confusing because before I reached function_pointer.tscn, I thought this was for drawing the cursor in the viewport and wasn't expecting a 3d mesh.

Maybe rename to clarify? (some ideas: `raycast_visual`, `pointer_ray_mesh`, `cursor_cast_mesh`)

demo/addons/common/function_pointer/function_pointer.gd
```diff
@@ -0,0 +1,180 @@
...
+	set(v):
+		released_material = v
+		_update_pointer_material()
+
+@export var pressed_material: Material:
+	set(v):
+		pressed_material = v
+		_update_pointer_material()
+
+@export var pointer_length := 5.0
```
dhoverb: Should this also call `_update_pointer_length`?

demo/addons/common/function_pointer/function_pointer.gd
```diff
@@ -0,0 +1,180 @@
...
+@export var pointer_offset := 0.07
```
dhoverb: same for this since it is part of the pointer length logic

demo/addons/common/function_pointer/function_pointer.gd
```diff
@@ -0,0 +1,180 @@
...
+		length = pointer_length
+	_update_pointer_length(length)
+
+
+func _update_pointer_material() -> void:
+	if _mesh_instance:
+		_mesh_instance.set_surface_override_material(0, pressed_material if _pressed else released_material)
+
+
+func _on_controller_input_float_changed(p_name: String, p_value: float) -> void:
```
dhoverb: Wonder if hand_pinch_detector could be used?  Though might be a little awkward since it is also a Node, unless if some refactoring is done to separate the business logic from the Node.
(it's fine to address this in a separate PR, or do nothing)

Also, you'd probably end up keeping this (or refactor hand_pinch_detector even more) if you wanted the cursor to react to how close/far the fingers are from pinching (sort of like a visual indication to the user that they are interacting with the cursor).

demo/addons/common/function_pointer/function_pointer.gd
```diff
@@ -0,0 +1,180 @@
...
+	if p_name == select_action:
```
dhoverb: subjective: similar to `_process()`, you could remove one level of indentation with an early return

demo/addons/common/function_pointer/function_pointer.gd
```diff
@@ -0,0 +1,180 @@
...
+			if _cur_layer:
+				_cur_layer.pointer_set_pressed(self, true)
+		elif _pressed and p_value < select_release_threshold:
+			_pressed = false
+			_update_pointer_material()
+			if _cur_layer:
+				_cur_layer.pointer_set_pressed(self, false)
+
+
+func _do_select_press() -> void:
```
dhoverb: This and `_do_select_release` are called in one place (and are fairly short); could merge functions

demo/addons/common/function_pointer/function_pointer.tscn
dhoverb, line 24: Can these two nodes be merged?  (why are they separate?)

demo/addons/common/function_pointer/function_pointer.tscn
```diff
@@ -0,0 +1,29 @@
...
+
+[node name="FunctionPointer" type="Node3D" groups=["function_pointer"]]
+script = ExtResource("1_pjipa")
+select_action = "trigger"
+released_material = SubResource("StandardMaterial3D_pjipa")
+pressed_material = SubResource("StandardMaterial3D_g1l7f")
+
+[node name="MeshInstance3D" type="MeshInstance3D" parent="."]
+unique_name_in_owner = true
+transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, -2.57)
```
dhoverb, line 26: should this be deleted? the initial transform doesn't matter right?

demo/addons/common/function_pointer/function_pointer.tscn
dhoverb: Should the skeleton be deleted?

demo/addons/common/view_pivot/view_pivot.gd
```diff
@@ -0,0 +1,124 @@
+extends Node3D
+
+## The minimum threshold that the dot product between the camera direction
+## and desired direction must fall below before the pivot is rotated.
+## Higher values will rotate more often, and lower values will be looser.
+@export var direction_threshold := 0.9
```
dhoverb, line 6: Shouldn't this have a setter too?
The reason if if you set a higher threshold than the current (like, from 0.0 to 0.9).  `_process()` will eventually update it, but it would be necessary if there are other queries that depend on the result of this change now.

demo/addons/common/view_pivot/view_pivot.gd
```diff
@@ -0,0 +1,124 @@
...
+	if camera == null:
+		return _current_direction
+
+	return camera.global_position
+
+
+func update_transform_immediately() -> void:
+	_current_direction = _get_camera_direction()
+	_current_position = _get_camera_position()
+	_pivot_transform.origin = Vector3(_current_position.x, 0.0, _current_position.z)
```
dhoverb: Why is the origin always y == 0?

demo/addons/common/view_pivot/view_pivot.gd
```diff
@@ -0,0 +1,124 @@
...
+		if _current_direction.dot(camera_direction) < direction_threshold or camera_distance > distance_threshold:
+			if _timer.is_stopped():
+				_timer.start()
+		else:
+			_timer.stop()
+
+
+func _notification(p_what: int) -> void:
+	match p_what:
+		NOTIFICATION_VISIBILITY_CHANGED:
```
dhoverb, line 98: Maybe `if p_what == NOTIFICATION_VISIBILITY_CHANGED:` since there's only one match case?

demo/addons/common/view_pivot/view_pivot.gd
```diff
@@ -0,0 +1,124 @@
...
+	if camera == null or not camera is XRCamera3D:
+		return null
+
+	return camera as XRCamera3D
+
+
+func _get_camera_direction() -> Vector3:
+	var camera: XRCamera3D = _get_camera()
+	if camera == null:
+		return _current_direction
```
dhoverb, line 52: this case is kind of undefined, so, returning anything is "wrong".  Maybe this should just be `Vector3(0, 0, -1)`?
Similar for `_get_camera_position()` (though return `Vector3.ZERO`)

demo/addons/common/view_pivot/view_pivot.gd
```diff
@@ -0,0 +1,124 @@
...
+		NOTIFICATION_VISIBILITY_CHANGED:
+			if is_visible_in_tree():
+				update_transform_immediately()
+				set_process(true)
+			else:
+				set_process(false)
+
+
+func _on_timer_timeout() -> void:
+	_current_position = _get_camera_position()
```
dhoverb, line 107: `_get_camera_position()` is assigned to `camera_direction` on line 88, but to `_current_position` here.  It's hard to tell if `_current_position` refers to the camera's current position or to this node's current position.
Renaming `_current_position` to `_camera_current_position` would help, assuming it is for the camera's current position.

demo/addons/common/view_pivot/view_pivot.gd
```diff
@@ -0,0 +1,124 @@
...
+	_current_direction = _get_camera_direction()
+	var new_basis = Basis.looking_at(_current_direction)
+
+	var pivot_tween = get_tree().create_tween()
```
dhoverb: can another tween be currently running?

demo/addons/common/view_pivot/view_pivot.gd
```diff
@@ -0,0 +1,124 @@
...
+	pivot_tween.tween_method(self._tween_pivot_transform_basis, _pivot_transform.basis.get_rotation_quaternion(), new_basis.get_rotation_quaternion(), 1.0).set_trans(Tween.TRANS_EXPO).set_ease(Tween.EASE_OUT)
```
dhoverb: Should this tween from the current rotation, rather than _pivot_transform? (I'm confused on why _pivot_transform is needed, since it seems like in every place it is updated, global_transform is updated immediately after)

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
```diff
@@ -5,24 +5,27 @@ extends Node3D
...
+const DOUBLE_CLICK_DIST = 5.0
 
 ## The current [PackedScene] rendered by the quad.[br]
 ## The [PackedScene]'s root node must be a [Control] node.[br]
 ## Can be [code]null[/code].
 @export var scene: PackedScene:
 	set = set_scene
 
+## The viewport size in pixels.
+@export var viewport_size := Vector2i(1152, 648):
```
dhoverb, line 21: Explicitly setting the viewport size means ever use of viewport-2d-in-3d is now responsible for reacting to changes in the underlying gui3d's size, to ensure the viewport is large or small enough.

I saw the comment in the (now merged) pr, but I didn't understand how it was circular logic (it was: gui3d size changes due to added/removed control node(s) -> viewport-2d-in-3d changes size of viewport).  Without automatically updating the size, the burden to update the size is put on every use of the viewport-2d-in-3d (where it'll end up doing what viewport-2d-in-3d used to do, but can be error prone since every use of viewport-2d-in-3d must implement the same logic).

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
```diff
@@ -124,26 +140,158 @@ func get_scene_root() -> Control:
...
+
+func intersects_ray(p_origin: Vector3, p_direction: Vector3) -> Vector2:
+	if not visible:
+		return NO_INTERSECTION
+
+	var quad_transform: Transform3D = get_global_transform()
+	var quad_normal: Vector3 = quad_transform.basis.z
+
+	var denom: float = quad_normal.dot(p_direction)
+	if absf(denom) > 0.0001:
```
dhoverb: I think this is equivalent?

```gdscript
if denum < -0.0001:
```

This is because the ray's direction is always in the "opposite" direction as the quad's normal.  As in, if the ray's direction is the same as the quad's normal, then the ray is pointing away from the quad (or the ray is pointing at the back of the quad)

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
```diff
@@ -124,26 +140,158 @@ func get_scene_root() -> Control:
...
+
+func _intersect_to_viewport_pos(p_intersection: Vector2) -> Vector2:
+	if _viewport and p_intersection != NO_INTERSECTION:
+		var pos: Vector2 = p_intersection * Vector2(_viewport.size)
+		return Vector2(pos)
+
+	return NO_INTERSECTION
+
+
+func intersects_ray(p_origin: Vector3, p_direction: Vector3) -> Vector2:
```
dhoverb, line 172: Maybe add doccomment saying the returned Vector2 is a normalized point in 2D space (or something similar)?

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
```diff
@@ -124,26 +140,158 @@ func get_scene_root() -> Control:
...
+func pointer_intersects(p_pointer: Node3D) -> bool:
+	var pt := p_pointer.global_transform
+	var intersection := intersects_ray(pt.origin, -pt.basis.z)
+	if intersection != NO_INTERSECTION:
+		# If there was no current pointer, let's take this one.
+		if _pointer == null:
+			_pointer = p_pointer
+			_cursor.visible = true
+
+		var cursor_position = _intersect_to_global_pos(intersection, CURSOR_DISTANCE)
```
dhoverb, line 214: Wonder if it'd be useful to have two different `intersects_ray()`functions, since `intersects_ray` already calculated the global pos.

Something like:

```gdscript

## origin and direction are in global space
## returns the normalized intersection point in 2D space
func intersects_ray_2d(origin: Vector3, direction: Vector3) -> Vector2:
  var projected_point := _intersects_ray_projected(origin, direction)
  if projected_point == NO_INTERSECTION:
    return NO_INTERSECTION
  
  var u: float = 0.5 + (projected_point.x / quad_size.x)
  var v: float = 1.0 - (0.5 + (projected_point.y / quad_size.y))

  return Vector2(u, v)


## origin and direction are in global space
## returns the point in global 3D space
func intersects_ray_global(origin: Vector3, direction: Vector3) -> Vector3:
  var projected_point := _intersects_ray_projected(origin, direction)
  if projected_point == NO_INTERSECTION:
    return NO_INTERSECTION
  return get_global_transform() * projected_point


func _intersects_ray_projected(origin: Vector3, direction: Vector3) -> Vector3:
  # ...
  var projected_point := Vector3(relative_point.dot(quad_transform.basis.x), relative_point.dot(quad_transform.basis.y), relative_point.dot(quad_transform.basis.z))

  var quad_size := get_size()
  if absf(projected_point.x) > quad_size.x / 2.0:
    return NO_INTERSECTION
   if absf(projected_point.y) > quad_size.y / 2.0:
    return NO_INTERSECTION

  return projected_point
```

So then here you'd have:

```gdscript
func pointer_intersects(p_pointer: Node3D) -> bool:
  var pt := p_pointer.global_transform
  var cursor_position := intersects_ray_global(pt.origin, -pt.basis.z)
  if cursor_position != NO_INTERSECTION:
    # ...
```

Also, at a glance `CURSOR_DISTANCE` seems wrong since the distance from the ray to the intersection point can change.

plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
```diff
@@ -0,0 +1,398 @@
...
+		if (is_processing_internal()) {
+			configure_light_estimate_types();
+		}
+	} else {
+		directional_light_id = ObjectID();
+	}
+}
+
+DirectionalLight3D *OpenXRAndroidLightEstimation::get_directional_light() const {
+	return Object::cast_to<DirectionalLight3D>(ObjectDB::get_instance(directional_light_id));
```
dhoverb, line 185: Is this here in case if the DirectionalLight3D was free'd without telling us?
Not sure if this is better; but the alternative is to connect to a signal that tells us when the DirectionalLight3D is being destroyed (and just use the pointer instead of the id), though don't know if that exists.

plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
```diff
@@ -0,0 +1,398 @@
...
+	WorldEnvironment *world_environment = get_world_environment();
+	Ref<Environment> env;
+	if (world_environment) {
+		env = world_environment->get_environment();
+	}
+
+	if (env.is_valid() && ambient_light_mode != AMBIENT_LIGHT_MODE_DISABLED) {
+		if (ambient_light_mode == AMBIENT_LIGHT_MODE_COLOR && light_estimation_extension->is_ambient_light_valid()) {
+			Color intensity = light_estimation_extension->get_directional_light_intensity();
+			float luminance = (0.2126 * intensity.r) + (0.7152 * intensity.g) + (0.0722 * intensity.b);
```
dhoverb: what are these hardcoded values?

plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
```diff
@@ -0,0 +1,398 @@
...
+		configure_light_estimate_types();
+	}
+}
+
+OpenXRAndroidLightEstimation::AmbientLightMode OpenXRAndroidLightEstimation::get_ambient_light_mode() const {
+	return ambient_light_mode;
+}
+
+void OpenXRAndroidLightEstimation::set_spherical_harmonics_degree(SphericalHarmonicsDegree p_sh_degree) {
+	sh_degree = p_sh_degree;
```
dhoverb: maybe check that this is a valid number since it could be SPHERICAL_HARMONICS_DEGREE_MAX? (or if called from C++, it could be anything)

samples/androidxr-light-estimation-sample/custom_ambient_material.gdshader
dhoverb, line 1: at a glance this shader looks the same as the shader defined plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp

samples/androidxr-light-estimation-sample/main.gd
```diff
@@ -0,0 +1,92 @@
...
+
+func _on_openxr_session_begun() -> void:
+	start_light_estimation()
+
+
+func start_light_estimation() -> void:
+	if OpenXRAndroidLightEstimationExtensionWrapper.start_light_estimation():
+		print("Light estimation started")
+	else:
+		print("Unable to start light estimation")
```
dhoverb: `push_error` ?

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
```diff
@@ -124,26 +140,158 @@ func get_scene_root() -> Control:
...
-	set_pixel_size(pixel_size)
+	var event := InputEventMouseButton.new()
+	event.button_index = MOUSE_BUTTON_LEFT
+	event.button_mask = MOUSE_BUTTON_MASK_LEFT
+	event.pressed = p_pressed
+	event.position = _intersect_to_viewport_pos(_prev_intersection)
+
+	if p_pressed:
+		var time := Time.get_ticks_msec()
+		if time - _prev_pressed_time < DOUBLE_CLICK_TIME and (event.position - _prev_pressed_pos).length() < DOUBLE_CLICK_DIST:
```
dhoverb, line 259: Wonder if the logic in hand pinch detector could be generalized to include double clicking too.
Overall I see overlap with the hand pinch detector and this PR; could be addressed in a followup (if it makes sense to).

demo/addons/common/function_pointer/function_pointer.gd
dsnopek: Yeah, the source of the ray is the FunctionPointer node, so it is measuring the distance from the ray origin

demo/addons/common/function_pointer/function_pointer.gd
dsnopek: `Function*` and `function_*` are somewhat commonly used names for "component style" nodes. And anyone who's used XR Tools would be familiar with this naming convention, which would likely include many developers looking at these samples

demo/addons/common/function_pointer/function_pointer.gd
dsnopek: There is a performance difference, but it's not very big. I'll update it to use `set_process()` anyway!

demo/addons/common/function_pointer/function_pointer.tscn
dsnopek: Ah, yeah, I don't think `disable_receive_shadows` does anything with unshaded materials. It doesn't really hurt anything either, though

demo/addons/common/function_pointer/function_pointer.gd
dsnopek: Yes, I'll update it!

demo/addons/common/function_pointer/function_pointer.gd
dsnopek: Personally, I think it makes sense to keep them separate - they have very different use cases

demo/addons/common/function_pointer/function_pointer.tscn
dsnopek: > Can these two nodes be merged? (why are they separate?)

If they were merged, then when the node was instanced in a scene, all the properties for `MeshInstance3D` would be shown. But, we don't want the developer to interact with the `MeshInstance3D`, that's all hidden behind the `FunctionPointer` node

demo/addons/common/function_pointer/function_pointer.tscn
dsnopek: It makes for a nice visual representation in the editor.

I suppose the alternative is to make this a `@tool` script, and then the visual representation will get updated by the `pointer_*` properties. But even if we did that, this value would still get saved :-)

demo/addons/common/function_pointer/function_pointer.tscn
dsnopek: Hm, yeah, this shouldn't be here. I'm not sure how that got there - I'll remove it and see if the editor tries to add it back

demo/addons/common/view_pivot/view_pivot.gd
dsnopek: I don't think that's needed. This whole node is just to animate something around a pivot based on where the user is looking, and this just decides when to start animating. I don't think anything's going to need to query this value - any other code will just want to look at the global positions of this node's children

demo/addons/common/view_pivot/view_pivot.gd
dsnopek: The Y position of this node doesn't really matter. I suppose we could have it be like:

```suggestion
	_pivot_transform.origin = Vector3(_current_position.x, global_position.y, _current_position.z)
```

So, then we're maintaining whatever original Y position the user supplied?

demo/addons/common/view_pivot/view_pivot.gd
dsnopek: Sure, but I'm not sure it matters too much. I usually just write these as `match` in GDScript and `switch` in C++, since adding more stuff is sort of inevitable

demo/addons/common/view_pivot/view_pivot.gd
dsnopek: No, although, thanks for pointing this out, because it made me realize the tween's duration needs to be the same as the `movement_delay` which it currently isn't

demo/addons/common/view_pivot/view_pivot.gd
dsnopek: Ah, yeah, `_pivot_transform` might not be needed. This was originally ported from a game that did things a little differently, and I think it might just be left over from that

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
dsnopek: I think the whole idea of the viewport resizing to match the controls is backwards. Generally, UI's are built in a "size agnostic" way, that changes based on the viewport.

This is how, for example, you build a UI that'll work on any smart phone even with the differences in size and aspect ratio, or as the user resizes the window in a desktop app.

In this case, the `Viewport2Din3D` is the screen, and it should be made a size that makes sense in the given context, and the UI will adapt to it.

plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
dsnopek: This is the standard equation for luminance.

That said, I don't know if its the appropriate equation in this context for separating the color from the intensity - as noted in the description, this is something I need to spend some more time on

samples/androidxr-light-estimation-sample/custom_ambient_material.gdshader
dsnopek: It's very similar, but they're coming from different directions. The one in C++ is for generating a radiance map, and this one is calculating the color for the pixels of an object in the scene. Anyway, I don't think either is doing exactly what it should - I need to spend some more time on this

samples/androidxr-light-estimation-sample/main.gd
dsnopek: Makes sense!

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
dsnopek: Will do!

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
dsnopek: Makes sense!

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
dsnopek: Sure, having both methods could be useful! Maybe something for a follow-up?

> Also, at a glance `CURSOR_DISTANCE` seems wrong since the distance from the ray to the intersection point can change.

I don't understand what you mean by this

The `CURSOR_DISTANCE` is the distance to put the cursor is away from the quad in the quad's local +Z axis. It's there to put it far enough away that there's no Z fighting, but close enough that the user perceives them as touching.

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
dsnopek: Sure, it could make sense to implement double clicking in the pinch detector too. But I'm still not sure it makes sense to try and share code

demo/addons/common/function_pointer/function_pointer.gd
dsnopek: They are only called in one place here, but that's because I removed the WebXR support. I think I'd prefer to keep the functions themselves for some compatibility with the original code (so I can still copy back and forth :-)) but they could be merged into a single method with a boolean parameter

demo/addons/common/view_pivot/view_pivot.gd
dsnopek: I think that'll basically be the case? We default `_current_direction` to `Vector3(0, 0, -1)` (which is `Vector3.FORWARD`), so if there are no cameras that's what will be returned. Otherwise, if there was a camera in the past and this was calculated previously, it'll return what we had last, which I think is fine

demo/addons/common/view_pivot/view_pivot.gd
dsnopek: > `_get_camera_position()` is assigned to `camera_direction` on line 88, but to `_current_position` here.

I'm not sure what you're referring to?

> It's hard to tell if `_current_position` refers to the camera's current position or to this node's current position.

This node is "chasing" the camera position. Once we decide to move this node (because we've gotten over the threshold and the timeout has happened), we set `_current_position` to the new position (which will be the camera position), but we won't necessarily update the node's position right away, we may animate it there

So, `_current_position` is the like the current "effective" position, but we may not have actually finished moving the node there yet

plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
dsnopek: Yes. This is basically the same as doing `is_instance_valid()` in GDScript, which also works by looking up the object ID in `ObjectDB`

dsnopek:
@dhoverb Thanks for the review! I've made a PR based on your feedback for `viewport_2d_in_3d` and friends here: https://github.com/GodotVR/godot_openxr_vendors/pull/426

demo/addons/common/view_pivot/view_pivot.gd
dhoverb: > I'm not sure what you're referring to?

I meant that on line 88 `_get_camera_position()` is being assigned to something relating to direction.
Then here, `_get_camera_position()` is being assigned to something relating to position.
It seemed inconsistent; though my confusion may have been due to not understanding what `_current_position` represented.

> This node is "chasing" [...]

Ah, in the past I've named this kind of thing "target position".  As in, your current position (global_position) wants to eventually equal the target position (by lerp or some other easing means).

I guess another name that is a little more self explanatory would be "destination position" (or verbosely; "lerp to position", "ease to position", "animate to position", etc, though these all sound like a name for a bool and not a Vector3)

demo/addons/common/view_pivot/view_pivot.gd
dhoverb: Sounds good.  It was just a (probably extreme, or non-existent) edge case.

demo/addons/common/function_pointer/function_pointer.gd
dhoverb: Not sure what I saw when I read this originally, but I see that now.  Looks good!

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
dhoverb: > Sure, having both methods could be useful! Maybe something for a follow-up?

Yes, that is fine since it is refactor.

> I don't understand what you mean by this
The CURSOR_DISTANCE is the distance to put the cursor is away from the quad in the quad's local +Z axis. It's there to put it far enough away that there's no Z fighting, but close enough that the user perceives them as touching.

Ah, I was thinking `CURSOR_DISTANCE` was the distance from the hand to the quad, which would always be a different number.  What you said makes sense.

demo/addons/common/view_pivot/view_pivot.gd
dsnopek: > > I'm not sure what you're referring to?
> 
> I meant that on line 88 `_get_camera_position()` is being assigned to something relating to direction.

It isn't, though. This is line 88 in `master` which should be the same as it was here:

https://github.com/GodotVR/godot_openxr_vendors/blob/6b8b2e63428b1e175f9819d225021e998630951a/demo/addons/common/view_pivot/view_pivot.gd#L88-L93

Maybe you were misreading the `_get_camera_direction()` as `_get_camera_position()`?

demo/addons/common/viewport_2d_in_3d/viewport_2d_in_3d.gd
dhoverb: If I ever find time I could take a look at refactoring (assuming it makes sense).

I'm always looking for ways to have "one definition" or "one implementation" once I see code that is similar to other code.  This is true when the similar code adds new behavior or implements similar logic differently.  My thought is that it's easier to maintain code when the implementation is in one place, though usually needs to be commented clearly since it is a generalization.

demo/addons/common/view_pivot/view_pivot.gd
dhoverb: Ah, you're right.

plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
dsnopek: Updated in latest push!

samples/androidxr-light-estimation-sample/main.gd
dsnopek: Updated in latest push!

dsnopek:
This PR is basically ready for review at this point! It's only still in DRAFT because it depends on https://github.com/GodotVR/godot_openxr_vendors/pull/426 (and it only depends on that because of the moved files - otherwise, it doesn't need the changes from there)

I'm feeling more confident about the math (although, not 100% confident because rendering isn't my strongest area). I stole some code from [one of the official ARCore samples](https://github.com/google-ar/arcore-android-sdk/tree/52c722e43cd8ce546eea5dc4587e70e0c7f2c006/samples/hello_ar_kotlin), which appears to match the data from the OpenXR extension. And the result does look much better to me!

I've also added API documentation. This could use a tutorial to really explain how it works, but I'll save for a follow-up PR.

demo/addons/common/viewport_2d_in_3d/function_pointer.gd
```diff
@@ -66,23 +81,23 @@ func _exit_tree() -> void:
...
+					_cur_layer.pointer_leave(self)
+				_cur_layer = layer
 				_update_pointer_visibility()
-				_update_pointer_length(pointer_length)
+			break
+		elif _cur_layer == layer:
+			_cur_layer.pointer_leave(self)
+			_cur_layer = null
+			_update_pointer_visibility()
+			_update_pointer_length(pointer_length)
```
dhoverb, line 100: Isn't there a chance for a flicker between this and the next time _update_pointer_length_for_intersection() is called? 

For example:

1. The user is pointing at a layer
2. The user points off the layer and it happens to intersect a new layer sitting behind it (so, there was no empty space between the two layers.
3. This code happens to see `elif _cur_layer == layer:` first, so it calls `_update_pointer_length(pointer_length)`
4. Then, still in the same frame, this code sees `layer.pointer_intersects(self)` is true, and sets the new layer
5. A frame goes by before `_update_pointer_length_for_intersection()` is called to set another pointer length.

So steps 2-4 are in one frame, and step 5 is in the next frame.
This means the pointer length goes from intersecting the original layer, to `pointer_length` in one frame, then the next frame it is set to `_update_pointer_length_for_intersection()`.  Though, this is assuming `_update_pointer_length_for_intersection()` for the newly-intersecting layer is called in the next frame; I don't know if you can guarantee it though.

I think this will appear as a flicker if `pointer_length` is shorter than the distance to the new layer (if you test this, try setting `pointer_length` to a small value, like 0.1, to make it more obvious).
Also, since the layers are sorted, I think this'll only happen when the first layer is closer than the second layer.

plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
```diff
@@ -0,0 +1,361 @@
...
+#include <godot_cpp/classes/shader_material.hpp>
+#include <godot_cpp/classes/sky.hpp>
+#include <godot_cpp/classes/world_environment.hpp>
+#include <godot_cpp/classes/xr_server.hpp>
+
+#include "extensions/openxr_android_light_estimation_extension_wrapper.h"
+
+using namespace godot;
+
+// Lifted from ARCore's hello_ar_kotlin sample.
```
dhoverb: Maybe say `Copied`? (if you change it, don't forget to also change samples/androidxr-light-estimation-sample/custom_ambient_material.gdshader and samples/androidxr-light-estimation-sample/main.gd)

samples/androidxr-light-estimation-sample/main.gd
```diff
@@ -0,0 +1,107 @@
...
+	# Reset the original values.
+	directional_light.transform = directional_light_orig_transform
+	directional_light.light_color = directional_light_orig_color
+	directional_light.light_energy = directional_light_orig_energy
+
+	# Then change the mode.
+	$OpenXRAndroidLightEstimation.directional_light_mode = p_mode
+
+	# Goes as far as disabling the directional light if we're not using it.
+	%DirectionalLight3D.visible = (p_mode != OpenXRAndroidLightEstimation.DIRECTIONAL_LIGHT_MODE_DISABLED)
```
dhoverb, line 80: nit: parenthesis are not needed, but fine to leave it.

demo/addons/common/viewport_2d_in_3d/function_pointer.gd
dsnopek: I don't think so because it's `layer.pointer_intersects(self)` that calls to `_update_pointer_length_for_intersection()`, so there shouldn't a frame delay between intersecting the new layer and the pointer length being updated for that layer

plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
dsnopek: Sure, "copied" is more straight forward language than "lifted" :-)

doc_classes/OpenXRAndroidLightEstimation.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidLightEstimation" inherits="Node3D" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
```
m4gr3d, line 2: Does this node need to be a `Node3D`?

dsnopek: Right now, it's using the node's visibility (so `node.visible`) to control if it's enabled or disabled, and plain `Node` doesn't have that property. So, assuming we like this pattern, then, yes, it does need to be `Node3D` :-)

This is also done in `OpenXRMetaEnvironmentDepth` and `OpenXRAndroidEnvironmentDepth`, which is what I modeled the implementation off of

m4gr3d (approved):

PR #424, opened 2026-01-01
Update environment_depth.rst with permissions info. Fix code typo.

The-Cyber-Captain:
Picked up on a couple of typos discovered when I tried to actually follow the documented "Accessing the depth map on the CPU" example as a jumping off point for my project.

Clarified requirements for using Meta Environment Depth / OpenXRMetaEnvironmentDepthExtensionWrapper.start_environment_depth(); including necessary Android permissions.

With those in mind, veered away from the prior suggestion of triggering start_environment_depth() via the OpenXRInterface,session_began signal. As, surely, attempting to secure the permissions before the session has started will - at best - flood the logs with Condition "image_acquired" is true. Returning: true errors [<C++ Source>  modules/openxr/openxr_api.cpp:175 @ acquire()]?

//sigh// The legacy of evolving Android permissions - with or without Godot! - is such a ballache! 😜

### DIFF
--- a/docs/manual/meta/environment_depth.rst
+++ b/docs/manual/meta/environment_depth.rst
@@ -24,7 +24,7 @@ Before Meta Environment Depth can be used, it needs to be started:
 	if OpenXRMetaEnvironmentDepthExtensionWrapper.is_environment_depth_supported():
 		OpenXRMetaEnvironmentDepthExtensionWrapper.start_environment_depth()
 
-This will only work if there is both an active OpenXR session, and relevant Android permissions have all been granted in advance of the call. Namely, explicitly, "dangerous" permission ``android.permission.CAMERA``, and runtime permission ``USE_SCENE``. The latter can be automatically requested via the extension settings, Project Settings > OpenXR > Extensions > Automatically Request Runtime Permissions, or explicitly in code like the camera one.
+This will only work if there is an active OpenXR session. You can connect to the ``OpenXRInterface.session_begun`` signal to run code right when the session starts.
 
 There is a performance cost to using Meta Environment Depth, so you should only start it when needed, and stop it when no longer needed. For example, if your application has both a VR and AR mode, you should make sure that Meta Environment Depth is only running in AR mode.
 


docs/manual/meta/environment_depth.rst
```diff
@@ -24,7 +24,7 @@ Before Meta Environment Depth can be used, it needs to be started:
 	if OpenXRMetaEnvironmentDepthExtensionWrapper.is_environment_depth_supported():
 		OpenXRMetaEnvironmentDepthExtensionWrapper.start_environment_depth()
 
-This will only work if there is an active OpenXR session. You can connect to the ``OpenXRInterface.session_begun`` signal to run code right when the session starts.
+This will only work if there is both an active OpenXR session, and relevant Android permissions have all been granted in advance of the call. Namely, explicitly, "dangerous" permission ``android.permission.CAMERA``, and runtime permission ``USE_SCENE``. The latter can be automatically requested via the extension settings, Project Settings > OpenXR > Extensions > Automatically Request Runtime Permissions, or explicitly in code like the camera one.
```
dsnopek: As far as I know, the CAMERA permission isn't needed for the depth API to work. We aren't requesting it in the sample project, and it's always worked for me

I'll double check what happens when first granting USE_SCENE when I have a chance. Recommending folks to take a look at `OS.request_permission()` and `MainLoop.on_request_permissions_result` could certainly make sense!

The-Cyber-Captain: Yeahhhh.... you can imagine my surprise at the 
```
ERROR: Failed to create environment depth provider: XR_ERROR_NOT_PERMITTED_PASSTHROUGH_FB
(137) DEBUG: timer timed out
```
logs when trying to incorporate the Manual page example into my own test. I thought I was pretty thorough in eliminating the 'variables' before putting the PR in, but can now confirm CAMERA is *not* required.
Exhibit A & B:

https://github.com/user-attachments/assets/0ee5374b-8b97-4369-b77f-305b24dd76dc

no_camera


I guess I did a boob somewhere with my Permissions boiler-plate; the ongoing saga of The-Captain vs OS.android perms and the Quest. :wink: (https://github.com/godotengine/godot/pull/105364)

Let me see if I / we can come up with a suitable update to this page again when I'm finished with my project for the day. Thanks - as always - Snopek. [Open to suggestions!!! 😁 ]

The-Cyber-Captain: Ahhhhhh. I think I know where my misconception came from; my permissions boilerplate, indeed. When I tweaked it so as *not* to ask for Camera - purely for diagnosis prior to this PR - an unexpected side-effect was that start_environment_depth() was called during the same frame the XR session fires up. If not deferred, or on a later frame... like naturally when user-intervention takes place... then this causes the app to bail out - which I mistook for lack of permissions.

All that aside... still don't know what to change the document to. 🤣 
I think I'll just fix the typos and leave any talk of permission and service startup to the professionals.... BRB.

PR #428, opened 2026-01-08
Add Valve vendor to project setup wizard

devloglogan:
Adds Valve vendor to project setup wizard, for the Steam Frame. :slightly_smiling_face: 

I don't have one of these so I'm unable to test it.

### DIFF
--- a/plugin/src/main/cpp/editor/xr_project_setup_dialog.cpp
+++ b/plugin/src/main/cpp/editor/xr_project_setup_dialog.cpp
@@ -63,6 +63,7 @@
 #include <godot_cpp/classes/window.hpp>
 #include <godot_cpp/core/error_macros.hpp>
 #include <godot_cpp/core/memory.hpp>
+#include <godot_cpp/templates/hash_map.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
 static const char *XR_STARTUP_SCENE_PATH = "res://xr_startup.tscn";
@@ -88,6 +89,7 @@ enum VendorType {
 	VENDOR_TYPE_LYNX,
 	VENDOR_TYPE_MAGIC_LEAP,
 	VENDOR_TYPE_ANDROID_XR,
+	VENDOR_TYPE_VALVE,
 };
 
 namespace godot_openxr_vendors {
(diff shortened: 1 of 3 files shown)


dsnopek:
> Also we may want to enhance our export check that a Linux export setup is allowed too

I don't know that we should have the wizard recommend a Linux build while that's still crashing - I think we need to fix that first

BastiaanOlij:
> I don't know that we should have the wizard recommend a Linux build while that's still crashing - I think we need to fix that first

Absolutely, but once we got that figured out, I do think it is something to consider.

devloglogan:
Added a check for OpenXR in the `xr_mode` export setting to all vendors (except Meta because it relies on the meta toolkit). I also removed the `class_name` from the start xr scripts included in the startup scenes.

plugin/src/main/cpp/editor/xr_project_setup_dialog.cpp
```diff
@@ -368,39 +389,139 @@ void XrProjectSetupDialog::_notification(uint32_t p_what) {
...
+					"\n   - Android"
+					"\n       - Meta Toolkit enabled";
+			recommendations.push_back(memnew(ExportSettingRecommendation("Meta Export", meta_export_preset_description, "Open", ALERT_TYPE_ERROR, PROJECT_TYPE_ANY, VENDOR_TYPE_META, false, { meta_export_preset_values })));
+
+			// HTC export preset error.
+			HashMap<String, Variant> htc_export_preset_values;
+			htc_export_preset_values["platform"] = "Android";
+#ifndef ANDROID_ENABLED
+			htc_export_preset_values["gradle_build/use_gradle_build"] = true;
+#endif // !ANDROID_ENABLED
```
dsnopek: I wonder if these checks for Gradle on Android should actually be like:

```suggestion
#ifdef ANDROID_ENABLED
			bool gradle_supported = godot::internal::godot_version.minor >= 6;
#else
			bool gradle_supported = true;
#endif // ANDROID_ENABLED

			if (gradle_supported) {
				htc_export_preset_values["gradle_build/use_gradle_build"] = true;
			}
```

Because we already know that Gradle on Android will be supported in Godot 4.6+

This would apply to all vendors on Android except for Meta, which has the Meta Toolkit that can support the pre-built APK

devloglogan: Good idea, I've applied this suggestion!

PR #432, opened 2026-01-13
Add manual page about Android XR Environment Depth

dsnopek:
This adds a manual page about the feature added in PR https://github.com/GodotVR/godot_openxr_vendors/pull/400

It also reworks the sample to be more in line with what is described in the manual page, and to be somewhat more similar to the equivalent Meta sample.

And I snuck in one teeny tiny change to `hand_pinch_detector.gd` allowing the pinch action to be a variable rather than hard-coded


docs/manual/index.rst
```diff
@@ -9,4 +9,4 @@ Manual
 
     hybrid_apps
     meta/index
-
+    androidxr/index
```
m4gr3d, line 12: Should we keep it alphabetically ordered?

dsnopek: The main thing I don't like about alphabetical is that it would put "hybrid_apps" in between the two vendors. It would be nice to group the vendor specific stuff separate from the non-vendor specific stuff, but we only have 3 items so making more nested layers seems superfluous at this point

Maybe something to punt on for now and think more about when we have items?

m4gr3d: Sounds good.

PR #434, opened 2026-01-14
Minor tweaks to `XR_FB_space_warp` delta pose calculation and the sample

dsnopek:
I've been digging into space warp a bunch recently, and noticed a couple of minor things that this PR tweaks, including:

- Makes the delta pose calculation somewhat more robust to floating point precision issues. Before this it's possible (but very unlikely) to end up with "no change" giving a delta that is so small it should round to zero, but doesn't.
- Changes the sample's background to a clear color, rather than a sky, in order to make it easier to test that the delta pose calculation is correct. The idea for this comes from this comment in the Meta space warp sample project [here](https://github.com/meta-quest/Meta-OpenXR-SDK/blob/a4413c233a875978a5dd11c01e154d48c2075e36/Samples/XrSamples/XrSpaceWarp/Src/XrSpaceWarp.c#L3602).
- Makes the locomotion in the sample relative to the camera direction. Before this, locomotion would move you along the X and Z axes in world space
- Adds an export preset for Pico, since `XR_FB_space_warp` is also supported on Pico headsets


devloglogan (approved):
Everything LGTM, tested on Quest 3 and Pico 4 Ultra!

PR #455, opened 2026-02-09
Implement XR_ANDROID_trackables

dhoverb:
Implements [XR_ANDROID_trackables](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_ANDROID_trackables)


plugin/src/main/cpp/include/classes/openxr_android_anchor_space_tracker.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <androidxr/androidxr.h>
+#include <godot_cpp/classes/xr_positional_tracker.hpp>
+
+#include "classes/openxr_android_space_location.h"
+#include "classes/openxr_android_tracker.h"
+
+using namespace godot;
+
+class OpenXRAndroidAnchorSpaceTracker : public XRPositionalTracker {
```
dsnopek: Could we just call this `OpenXRAndroidAnchorTracker`? I don't feel like the "space" part is useful for Godot developers, given that we've hidden most of the details of OpenXR from developers (including `XrSpace`)

```suggestion
class OpenXRAndroidAnchorTracker : public XRPositionalTracker {
```

If we decided to rename this, it should affect all the rest of the names too, like `create_anchor_tracker()` (instead of `create_anchor_space_tracker()`).

doc_classes/OpenXRAndroidAnchorSpaceTracker.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidAnchorSpaceTracker" inherits="XRPositionalTracker" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Wraps the [code]XrSpace[/code] handle.
```
dsnopek: We don't require developers to know low-level OpenXR concepts like `XrSpace`:


```suggestion
		An anchor tracker (from [code]XR_ANDROID_trackables[/code]).
```

doc_classes/OpenXRAndroidAnchorSpaceTracker.xml
```diff
@@ -0,0 +1,98 @@
...
+	</brief_description>
+	<description>
+		Wraps the [code]XrSpace[/code] handle.
```
dsnopek:
```suggestion
		An anchor tracker (from [code]XR_ANDROID_trackables[/code]).
```

doc_classes/OpenXRAndroidObjectTracker.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidObjectTracker" inherits="OpenXRAndroidTracker" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Wraps the [code]XrTrackableObjectANDROID[/code] struct.
```
dsnopek: Here too, for the rest of these descriptions. These should describe the function of this object, not that it wraps a particular OpenXR struct, because then folks need to go look that up in the OpenXR spec, which will be unfamiliar to most Godot developers

plugin/src/main/cpp/include/classes/openxr_android_space_location.h
```diff
@@ -0,0 +1,69 @@
...
+
+#ifndef OPENXR_ANDROID_SPACE_LOCATION_H
+#define OPENXR_ANDROID_SPACE_LOCATION_H
+
+#include <androidxr/androidxr.h>
+#include <godot_cpp/classes/ref_counted.hpp>
+
+using namespace godot;
+
+class OpenXRAndroidSpaceLocation : public RefCounted {
```
dsnopek: Maybe `OpenXRAndroidAnchorLocation` (again, to remove "space")?


```suggestion
class OpenXRAndroidAnchorLocation : public RefCounted {
```

Overall, though, I'm not sure this class is need? Couldn't these methods be added to the `OpenXRAndroidAnchorSpaceTracker` itself: there could be an `update_transform()` method which gets the data from OpenXR, and then `get_transform_flags()` and `get_transform()` that get the actual data, or something like that

doc_classes/OpenXRAndroidTrackablesExtension.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidTrackablesExtension" inherits="OpenXRExtensionWrapperExtension" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
```
dsnopek:
```suggestion
<class name="OpenXRAndroidTrackablesExtension" inherits="OpenXRExtensionWrapper" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
```

dhoverb: Done

plugin/src/main/cpp/include/classes/openxr_android_space_location.h
dhoverb: Done

doc_classes/OpenXRAndroidObjectTracker.xml
dhoverb: Done

doc_classes/OpenXRAndroidAnchorSpaceTracker.xml
dhoverb: Done

doc_classes/OpenXRAndroidAnchorSpaceTracker.xml
dhoverb: Done

plugin/src/main/cpp/include/classes/openxr_android_anchor_space_tracker.h
dhoverb: Done

doc_classes/OpenXRAndroidObjectTracker.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidObjectTracker" inherits="OpenXRAndroidTracker" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		A tracker for object (from [code]XR_ANDROID_trackables[/code]).
```
dsnopek:
```suggestion
		A tracker for an object (from [code]XR_ANDROID_trackables[/code]).
```

doc_classes/OpenXRAndroidObjectTracker.xml
```diff
@@ -0,0 +1,38 @@
...
+	</brief_description>
+	<description>
+		A tracker for object (from [code]XR_ANDROID_trackables[/code]).
```
dsnopek:
```suggestion
		A tracker for an object (from [code]XR_ANDROID_trackables[/code]).
```

doc_classes/OpenXRAndroidPlaneTracker.xml
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidPlaneTracker" inherits="OpenXRAndroidTracker" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		A tracker for planes (from [code]XR_ANDROID_trackables[/code]).
```
dsnopek:
```suggestion
		A tracker for a plane (from [code]XR_ANDROID_trackables[/code]).
```

doc_classes/OpenXRAndroidPlaneTracker.xml
```diff
@@ -0,0 +1,83 @@
...
+	</brief_description>
+	<description>
+		A tracker for planes (from [code]XR_ANDROID_trackables[/code]).
```
dsnopek:
```suggestion
		A tracker for a plane (from [code]XR_ANDROID_trackables[/code]).
```

doc_classes/OpenXRAndroidTrackablesExtension.xml
```diff
@@ -0,0 +1,113 @@
...
+	<methods>
+		<method name="create_anchor_tracker">
+			<return type="OpenXRAndroidAnchorTracker" />
+			<param index="0" name="pose" type="Transform3D" />
+			<param index="1" name="trackable" type="OpenXRAndroidTracker" default="null" />
+			<description>
+				Create a new [OpenXRAndroidAnchorTracker], positioned at [param pose].
+				Unlike trackables, anchor trackers are not "discovered", as in, they must always be explicitly created.
+				Note that the [OpenXRAndroidAnchorTracker]'s location may not be ready.  See also [method OpenXRAndroidAnchorTracker.get_location_pose].
+				If [param trackable] is [code]non-null[/code], then the [OpenXRAndroidAnchorTracker] will only be available when the trackable is found.
```
dsnopek:
```suggestion
				If [param trackable] is not [code]null[/code], then the [OpenXRAndroidAnchorTracker] will only be available when the trackable is found.
```

doc_classes/OpenXRAndroidAnchorTracker.xml
```diff
@@ -0,0 +1,79 @@
...
+		</constant>
+		<constant name="LOCATION_FLAGS_ORIENTATION_TRACKED" value="4" enum="LocationFlags" is_bitfield="true">
+			Indicates that the [member Transform3D.basis] returned by [method get_location_pose] represents an actively tracked orientation.
+		</constant>
+		<constant name="LOCATION_FLAGS_POSITION_TRACKED" value="8" enum="LocationFlags" is_bitfield="true">
+			Indicates that the [member Transform3D.origin] returned by [method get_location_pose] represents an actively tracked position.
+		</constant>
+		<constant name="LOCATION_FLAGS_MAX" value="8" enum="LocationFlags" is_bitfield="true">
+			The maximum flag bit.
+		</constant>
```
dsnopek: I don't think a `*_MAX` flag makes sense for a bitfield enum. Those are better for sequential enums, where we might iterate over the values or make an array that contains a value for each member of the enum

samples/androidxr-sceneplanes-sample/export_presets.cfg
```diff
@@ -0,0 +1,258 @@
+[runnable_presets]
+
+Android="Android"
```
dsnopek: This looks like a Godot 4.7 thing that snuck in here. The samples should be built with Godot 4.6

samples/androidxr-sceneplanes-sample/project.godot
```diff
@@ -0,0 +1,33 @@
...
+
+renderer/rendering_method="mobile"
+textures/vram_compression/import_etc2_astc=true
+environment/defaults/default_clear_color=Color(0, 0, 0, 0)
+
+[xr]
+
+openxr/enabled=true
+openxr/environment_blend_mode=2
+openxr/extensions/spatial_entity/enabled=true
```
dsnopek: Do the EXT spatial entities need to be enabled? I would have expected this to be disabled

doc_classes/OpenXRAndroidTrackablesExtension.xml
```diff
@@ -0,0 +1,113 @@
...
+			</description>
+		</method>
+		<method name="destroy_anchor_tracker">
+			<return type="void" />
+			<param index="0" name="anchor_tracker" type="OpenXRAndroidAnchorTracker" />
+			<description>
+				Destroyes the underlying xrAnchorSpace.
```
dsnopek:
```suggestion
				Destroys the underlying tracker.
```

doc_classes/OpenXRAndroidPlaneTracker.xml
```diff
@@ -0,0 +1,83 @@
...
+		See also [OpenXRAndroidTrackablesExtension].
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="get_extents">
+			<return type="Vector2" />
+			<description>
+				Returns the width and height, or [member Vector3.x] and [member Vector3.y] respectively.
```
dsnopek:
```suggestion
				Returns the width and height, or [member Vector2.x] and [member Vector2.y] respectively.
```

doc_classes/OpenXRAndroidAnchorTracker.xml
```diff
@@ -0,0 +1,79 @@
...
+	</brief_description>
+	<description>
+		An anchor tracker (from [code]XR_ANDROID_trackables[/code]).
+
+		[OpenXRAndroidAnchorTracker]s can be created by [method OpenXRAndroidTrackablesExtension.create_anchor_tracker].
+		All newly created [OpenXRAndroidAnchorTracker]s have up-to-date state, as in, no need to call  [method get_location_flags] with [code]update == true[/code].
+		When an [OpenXRAndroidAnchorTracker] is created by [method OpenXRAndroidTrackablesExtension.create_anchor_tracker], the [code]trackable[/code] parameter can be set to either [code]null[/code] or a plane [OpenXRAndroidTracker].
+		The [OpenXRAndroidAnchorTracker] is called a "spatial anchor" when created with a [code]null[/code] [OpenXRAndroidTracker].
+		The [OpenXRAndroidAnchorTracker] is called a "plane anchor" when created with a plane [OpenXRAndroidTracker].
+		The [OpenXRAndroidAnchorTracker] cannot be created with other [OpenXRAndroidTracker]s types.
```
dsnopek:
```suggestion
		The [OpenXRAndroidAnchorTracker] cannot be created with other [OpenXRAndroidTracker] types.
```

plugin/src/main/cpp/include/classes/openxr_android_anchor_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+		LOCATION_FLAGS_ORIENTATION_TRACKED = 4,
+		LOCATION_FLAGS_POSITION_TRACKED = 8,
+
+		LOCATION_FLAGS_MAX = 8
+	};
+
+	BitField<LocationFlags> get_location_flags(bool p_update = false);
+	Transform3D get_location_pose(bool p_update = false);
+
+	// TODO is there a better place to put these?
```
dsnopek: We probably don't want to merge this TODO :-)

plugin/src/main/cpp/include/classes/openxr_android_anchor_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+	static XrSpaceLocation create_empty_location();
+
+protected:
+	static void _bind_methods();
+
+private:
+	void _update_location(bool p_update);
+
+	XrSpace space = XR_NULL_HANDLE;
+	XrSpaceLocation location;
```
dsnopek: This should be zero'd out here or in the constructor:


```suggestion
	XrSpaceLocation location = {};
```

plugin/src/main/cpp/include/classes/openxr_android_plane_tracker.h
```diff
@@ -0,0 +1,105 @@
...
+
+	static Ref<OpenXRAndroidPlaneTracker> create(XrTrackableANDROID p_trackable);
+
+	Vector2 get_extents();
+	PlaneType get_plane_type();
+	PlaneLabel get_plane_label();
+	Ref<OpenXRAndroidPlaneTracker> get_subsumed_by_plane();
+	PackedVector3Array get_vertices();
+	Ref<ArrayMesh> get_mesh();
+	Ref<ConcavePolygonShape3D> get_shape(real_t p_thickness);
```
dsnopek: Does this really need to be `real_t`? That creates a difference between 32 and 64 bit builds, which I suspect we don't need here. If 64-bit precision is needed, just use a `double`

plugin/src/main/cpp/include/classes/openxr_android_plane_tracker.h
```diff
@@ -0,0 +1,105 @@
...
+
+	XrTime last_updated_time_vertices = -1;
+
+	Vector2 extents;
+	PlaneType plane_type = PLANE_TYPE_ARBITRARY;
+	PlaneLabel plane_label = PLANE_LABEL_UNKNOWN;
+	Ref<OpenXRAndroidPlaneTracker> subsumed_by_plane;
+	PackedVector3Array vertices;
+
+	XrTrackablePlaneANDROID xr_trackable_plane;
```
dsnopek: Either here or in the constructor:

```suggestion
	XrTrackablePlaneANDROID xr_trackable_plane = {};
```

plugin/src/main/cpp/classes/openxr_android_anchor_tracker.cpp
```diff
@@ -0,0 +1,181 @@
...
+	}
+	location.pose = new_location.pose;
+
+	location_pose.origin.x = location.pose.position.x;
+	location_pose.origin.y = location.pose.position.y;
+	location_pose.origin.z = location.pose.position.z;
+	location_pose.basis = Basis{ Quaternion{ location.pose.orientation.x, location.pose.orientation.y, location.pose.orientation.z, location.pose.orientation.w } };
+
+	XRPose::TrackingConfidence confidence;
+	if (0 == (location.locationFlags & XR_SPACE_LOCATION_POSITION_TRACKED_BIT) || 0 == (location.locationFlags & XR_SPACE_LOCATION_POSITION_TRACKED_BIT)) {
```
dsnopek: This checks for `XR_SPACE_LOCATION_POSITION_TRACKED_BIT` twice. Maybe should be:


```suggestion
	if (0 == (location.locationFlags & XR_SPACE_LOCATION_POSITION_TRACKED_BIT) || 0 == (location.locationFlags & XR_SPACE_LOCATION_ORIENTATION_TRACKED_BIT)) {
```

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
```diff
@@ -0,0 +1,463 @@
...
+	if (trackable_count_output != trackable_count) {
+		WARN_PRINT("OpenXR: trackable query count differs from actual query");
+	}
+
+	// Run through the queried trackables.
+	// Absent trackables will remain in p_trackables_to_delete (and will be deleted later)
+	// New trackables will be added to the XRServer in get_or_create_tracker_and_update()
+	// All found trackables will be updated
+	for (int i = 0; i < trackable_count_output; ++i) {
+		Ref<OpenXRAndroidPlaneTracker> tracker = get_or_create_tracker_and_update(trackables[i], p_trackable_type, p_update_trackers);
```
dsnopek: Is this really meant to use `Ref<OpenXRAndroidPlaneTracker>`? This means that it will be `null` if a different kind of tracker is returned from `get_or_create_tracker_and_update()`, and this code doesn't seem to be specific to planes. I suspect it should be:


```suggestion
		Ref<OpenXRAndroidTracker> tracker = get_or_create_tracker_and_update(trackables[i], p_trackable_type, p_update_trackers);
```

samples/androidxr-sceneplanes-sample/main.gd
```diff
@@ -0,0 +1,48 @@
...
+
+const PLANE_TRACKABLE_VISUAL = preload("res://plane_trackable_visual/plane_trackable_visual.tscn")
+
+var _plane_visuals_dict: Dictionary
+
+
+func _ready():
+	super._ready()
+
+	OS.request_permissions()
```
dsnopek: Should `AndroidXREditorExportPlugin` be adding any permissions to the `AndroidManifest.xml` if this extension is enabled?

doc_classes/OpenXRAndroidTracker.xml
```diff
@@ -0,0 +1,44 @@
...
+	</signals>
+	<constants>
+		<constant name="TRACKING_STATE_PAUSED" value="0" enum="TrackingState">
+		</constant>
+		<constant name="TRACKING_STATE_STOPPED" value="1" enum="TrackingState">
+		</constant>
+		<constant name="TRACKING_STATE_TRACKING" value="2" enum="TrackingState">
+		</constant>
+		<constant name="TRACKING_STATE_UNKNOWN" value="3" enum="TrackingState">
+		</constant>
```
dsnopek, line 45: Not documented

plugin/src/main/cpp/classes/openxr_android_tracker.cpp
```diff
@@ -0,0 +1,130 @@
...
+	center_pose.origin.y = p_center_pose.position.y;
+	center_pose.origin.z = p_center_pose.position.z;
+	center_pose.basis = Basis{ Quaternion{ p_center_pose.orientation.x, p_center_pose.orientation.y, p_center_pose.orientation.z, p_center_pose.orientation.w } };
+
+	XRPose::TrackingConfidence confidence;
+	switch (get_tracking_state()) {
+		case TRACKING_STATE_PAUSED:
+		case TRACKING_STATE_STOPPED:
+		case TRACKING_STATE_TRACKING:
+			confidence = XRPose::TrackingConfidence::XR_TRACKING_CONFIDENCE_HIGH;
```
dsnopek: Should `TRACKING_STATE_PAUSED` and `TRACKING_STATE_STOPPED` really lead to _high_ confidence?

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	_on_process_anchors();
+	_on_process_trackables();
+}
+
+void OpenXRAndroidTrackablesExtension::_on_session_destroyed() {
+	_maybe_destroy_trackable_tracker(&plane_trackable_tracker);
+	_maybe_destroy_trackable_tracker(&depth_trackable_tracker);
+	_maybe_destroy_trackable_tracker(&object_trackable_tracker);
```
dsnopek: Should this also clean up `current_anchor_trackers` and `current_trackables`?

plugin/src/main/cpp/include/classes/openxr_android_object_tracker.h
```diff
@@ -0,0 +1,71 @@
...
+	static void _bind_methods();
+
+private:
+	void _ensure_updated() override;
+	bool _get_xrtrackable_info(XrTrackableObjectANDROID &p_xr_trackable_object);
+
+	uint64_t last_frame = -1;
+
+	Vector3 extents;
+	ObjectLabel object_label;
```
dsnopek: Either here or in the constructor:

```suggestion
	ObjectLabel object_label = OBJECT_LABEL_UNKNOWN;
```

plugin/src/main/cpp/classes/openxr_android_object_tracker.cpp
```diff
@@ -0,0 +1,163 @@
...
+void OpenXRAndroidObjectTracker::_ensure_updated() {
+	// This function can be called multiple times per frame, however it will only do work once per
+	// frame
+	uint64_t frame = Engine::get_singleton()->get_process_frames();
+	if (last_frame == frame) {
+		return;
+	}
+	last_frame = frame;
+
+	OpenXRAndroidTrackablesExtension *wrapper = OpenXRAndroidTrackablesExtension::get_singleton();
```
dsnopek: Unused variable

plugin/src/main/cpp/include/extensions/openxr_android_trackables_extension.h
```diff
@@ -0,0 +1,111 @@
...
+	HashMap<String, XrPath> paths;
+	HashMap<String, bool *> request_extensions;
+	bool available = false;
+	bool obj_available = false;
+
+	HashMap<XrTrackableANDROID, Ref<OpenXRAndroidTracker>> current_trackables;
+	int trackable_discovery_cooldown = 60;
+	int trackable_discovery_cooldown_cur = 0;
+	XrTrackableTrackerANDROID plane_trackable_tracker = XR_NULL_HANDLE;
+	XrTrackableTrackerANDROID depth_trackable_tracker = XR_NULL_HANDLE;
```
dsnopek: This member doesn't seem to be used here? Is this for a future PR?

dsnopek (commented):
Thanks!

At a high-level, this is looking pretty good to me! I tested the sample, and other than the plane tracking not being super accurate (which I don't think is the fault of this code), it worked great :-)

Other than my notes about some tiny details and fixing CI's complaints, I think all this really needs is a review from @BastiaanOlij on if the API defined here is close enough to using the EXT spatial entities, such that folks will be able to easily transition

doc_classes/OpenXRAndroidObjectTracker.xml
```diff
@@ -0,0 +1,38 @@
...
+	<constants>
+		<constant name="OBJECT_LABEL_UNKNOWN" value="0" enum="ObjectLabel">
+		</constant>
+		<constant name="OBJECT_LABEL_KEYBOARD" value="1" enum="ObjectLabel">
+		</constant>
+		<constant name="OBJECT_LABEL_MOUSE" value="2" enum="ObjectLabel">
+		</constant>
+		<constant name="OBJECT_LABEL_LAPTOP" value="3" enum="ObjectLabel">
+		</constant>
+	</constants>
```
dsnopek: Not documented

doc_classes/OpenXRAndroidPlaneTracker.xml
```diff
@@ -0,0 +1,83 @@
...
+		<constant name="PLANE_LABEL_UNKNOWN" value="0" enum="PlaneLabel">
+		</constant>
+		<constant name="PLANE_LABEL_WALL" value="1" enum="PlaneLabel">
+		</constant>
+		<constant name="PLANE_LABEL_FLOOR" value="2" enum="PlaneLabel">
+		</constant>
+		<constant name="PLANE_LABEL_CEILING" value="3" enum="PlaneLabel">
+		</constant>
+		<constant name="PLANE_LABEL_TABLE" value="4" enum="PlaneLabel">
+		</constant>
```
dsnopek, line 92: Not documented

doc_classes/OpenXRAndroidAnchorTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidAnchorTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidObjectTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidObjectTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidObjectTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidPlaneTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidPlaneTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidPlaneTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidPlaneTracker.xml
dhoverb: Done.

doc_classes/OpenXRAndroidTrackablesExtension.xml
dhoverb: Done.

doc_classes/OpenXRAndroidTrackablesExtension.xml
dhoverb: Done.

doc_classes/OpenXRAndroidTracker.xml
dhoverb: Done.

plugin/src/main/cpp/classes/openxr_android_anchor_tracker.cpp
dhoverb: Done.

plugin/src/main/cpp/classes/openxr_android_object_tracker.cpp
dhoverb: Done.

plugin/src/main/cpp/classes/openxr_android_tracker.cpp
dhoverb: I was thinking that it is high confidence that they are paused/stopped (i.e their last position is high confidence, even though they're paused/stopped).

Though, I see what you mean since the position is probably stale after some time.  Do you think these should be low confidence?

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
dhoverb: Done.

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
dhoverb: This code is correct, and getting `null` due to it not being a `OpenXRAndroidPlaneTracker` is intentional.

However...this code was making an assumption that is no longer true (the assumption was "xrGetAllTrackablesANDROID doesn't return subsumed planes", which is incorrect).
I don't recall when that assumption was made, but I refactored and tested this code and it seems to work fine.

plugin/src/main/cpp/include/classes/openxr_android_anchor_tracker.h
dhoverb: Heh, I still don't know where to put these functions.
I'll remove the TODO.

plugin/src/main/cpp/include/classes/openxr_android_anchor_tracker.h
dhoverb: Done.

plugin/src/main/cpp/include/classes/openxr_android_object_tracker.h
dhoverb: Done.

plugin/src/main/cpp/include/classes/openxr_android_plane_tracker.h
dhoverb: This was copy/paste from https://github.com/godotengine/godot/blob/master/modules/openxr/extensions/spatial_entities/openxr_spatial_plane_tracking.h#L167

I'll just change to float

plugin/src/main/cpp/include/classes/openxr_android_plane_tracker.h
dhoverb: Done.

plugin/src/main/cpp/include/extensions/openxr_android_trackables_extension.h
dhoverb: Yes, the next PR (XR_ANDROID_raycast).  I'll remove it from this PR.

samples/androidxr-sceneplanes-sample/main.gd
dhoverb: Yes, it does though I currently don't think it's technically correct.

Right now project.godot includes this line:
```
openxr/extensions/spatial_entity/enabled=true
```

which ultimately requests SCENE_UNDERSTANDING_COARSE in the AndroidManifest.xml
https://github.com/GodotVR/godot_openxr_vendors/blob/master/plugin/src/main/cpp/export/android_xr_export_plugin.cpp#L338

I'll update the check to look like this so we don't need to enable spatial entities:
```
} else if ((bool)project_settings->get_setting_with_override("xr/openxr/extensions/spatial_entity/enabled") ||
           (bool)project_settings->get_setting_with_override("xr/openxr/extensions/androidxr/light_estimation") ||
           (bool)project_settings->get_setting_with_override("xr/openxr/extensions/androidxr/trackables")) {
  contents += "    <uses-permission android:name=\"android.permission.SCENE_UNDERSTANDING_COARSE\" />\n";
}
```

samples/androidxr-sceneplanes-sample/project.godot
dhoverb: I responded to your other comment, asked a question (not submitted), and you answered it here, haha.

Yeah I'll remove this - it is from an early implementation that was refactored.

plugin/src/main/cpp/classes/openxr_android_tracker.cpp
dsnopek: I interpret the confidence to be about how confident we are that the transform is correct, so, yeah, my inclination would be to say we have low confidence (or maybe even "none" if the transform gets cleared in this case?) when the tracker is paused or stopped

samples/androidxr-sceneplanes-sample/export_presets.cfg
dhoverb: Done.

plugin/src/main/cpp/classes/openxr_android_tracker.cpp
dhoverb: Done.

dhoverb:
> One more question: How does a developer check if Android XR trackables are available? I don't see a method for that

I added `is_trackables_supported()` (and similar for raycast and anchor persistence in the other PRs).

plugin/src/main/cpp/include/extensions/openxr_android_trackables_extension.h
```diff
@@ -0,0 +1,150 @@
...
+	// If we do decide to add it, use OpenXRAndroidTrackablesObjectExtension as an example of which
+	// APIs and members to add, and for raycasts we'll have to search all plane-contexts to find the
+	// matching XrTrackableANDROID
+	HashMap<XrTrackableANDROID, Ref<OpenXRAndroidTrackableTracker>> current_plane_trackables;
+	XrTrackableTrackerANDROID plane_trackable_tracker = XR_NULL_HANDLE;
+	int next_plane_tracker_id = 1;
+	int plane_trackable_discovery_cooldown = 60;
+	int plane_trackable_discovery_cooldown_cur = 0;
+
+	// Achors
```
dsnopek:
```suggestion
	// Anchors
```

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
```diff
@@ -0,0 +1,573 @@
...
+		return;
+	}
+
+	_on_process_anchors();
+	_on_process_plane_trackers();
+}
+
+void OpenXRAndroidTrackablesExtension::_on_session_destroyed() {
+	maybe_destroy_trackable_tracker(plane_trackable_tracker, current_plane_trackables);
+	current_anchor_trackers.clear();
```
dsnopek, line 255: Does this need to remove all the trackers from `XRServer` and call `xrDestroySpace()` on each of the spaces? The latter might be optional, given that the session is being destroyed

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
```diff
@@ -0,0 +1,573 @@
...
+	return ret;
+}
+
+void OpenXRAndroidTrackablesExtension::destroy_anchor_tracker(Ref<OpenXRAndroidAnchorTracker> p_anchor_tracker) {
+	if (p_anchor_tracker.is_null()) {
+		return;
+	}
+
+	if (!current_anchor_trackers.has(p_anchor_tracker->get_xrspace())) {
+		UtilityFunctions::printerr("OpenXR: Failed to destroy anchor space; where did this tracker come from?");
```
dsnopek:
```suggestion
		UtilityFunctions::printerr("OpenXR: Failed to destroy anchor space; has it already been destroyed?");
```

dsnopek (commented):
Thanks!

I've reviewed this PR enough times that I think my brain is getting a little bit numb to it. :-) Even so, this is looking really good to me! Only the 2nd of my notes really _needs_ to be looked into; the other two are just nitpicks. But after that, I'm happy to give my approval

However, this could still use a review from @BastiaanOlij, to see if he thinks the API is close enough to the EXT spatial entities API

plugin/src/main/cpp/include/extensions/openxr_android_trackables_extension.h
dhoverb: Done

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
dhoverb: Nice catch, done.

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
dhoverb: Done.

dsnopek (approved):
Thanks! Looks good to me :-)

BastiaanOlij (approved):
This looks good to me. Am I correct that the sample only focuses on plane tracking and doesn't show off anchors?

Only nitpick is that I think the documentation could use some info on the relationship (or lack there of) between this and spatial entities. Possibly that is something that we should add in the help pages.

dhoverb:
> This looks good to me. Am I correct that the sample only focuses on plane tracking and doesn't show off anchors?

Yes.
Anchors (and persistence anchors) are tested in the androidxr-anchors-sample in https://github.com/GodotVR/godot_openxr_vendors/pull/457

> Only nitpick is that I think the documentation could use some info on the relationship (or lack there of) between this and spatial entities. Possibly that is something that we should add in the help pages.

There are some similarities w.r.t API and default behavior but the underlying implementations and configurability for them is fairly (extremely) different.
From what I'm thinking it would be a bit of work to get it straightened out (and might quickly change from upcoming changes).  Can this be addressed in a separate PR?

BastiaanOlij:
> From what I'm thinking it would be a bit of work to get it straightened out (and might quickly change from upcoming changes). Can this be addressed in a separate PR?

Oh absolutely, I don't think this is something that should prevent the current work from being merged. To be exact, the documentation part of this might not even end up in the class documentation but instead on manual pages, completely separate.

samples/androidxr-sceneplanes-sample/export_presets.cfg
```diff
@@ -0,0 +1,255 @@
+[preset.0]
+
+name="Android"
```
dsnopek: We've standardized on calling these presets "Android XR":


```suggestion
name="Android XR"
```

samples/androidxr-sceneplanes-sample/export_presets.cfg
```diff
@@ -0,0 +1,255 @@
...
+platform="Android"
+runnable=false
```
dsnopek:
```suggestion
runnable=true
```

samples/androidxr-sceneplanes-sample/export_presets.cfg
dhoverb: Done.

samples/androidxr-sceneplanes-sample/export_presets.cfg
dhoverb: Done.

dsnopek:
One thing I noticed when re-testing with the renamed sample app: the first time you run it, it won't detect any planes. But if you quit and restart the app, then it works.

I assume this is because of the permission check?

If so, we should have something that connects to `MainLoop.on_request_permissions_result` and restarts trackables once the permission is granted. We do this for `XR_FB_hand_tracking_mesh`, for example, see [`OpenXRFbHandTrackingMesh::on_request_permissions_result()`](https://github.com/GodotVR/godot_openxr_vendors/blob/c7e70348a8eb645f623a8b4fe90b18e007b15c18/plugin/src/main/cpp/classes/openxr_fb_hand_tracking_mesh.cpp#L68)

dhoverb:
> One thing I noticed when re-testing with the renamed sample app: the first time you run it, it won't detect any planes. But if you quit and restart the app, then it works.
> 
> I assume this is because of the permission check?

Thanks @dsnopek !
I didn't catch this because I install with `-g` (which grants every requested permission).  I am able to repro without the `-g` option.

I spent all day today getting something to handle these cases:
1: launch app without permissions, then accept permission from dialog
2: launch app without permissions, then decline permission from dialog.  Open settings app, navigate to app, grant permission, then resume the app
3: launch app without permissions, then decline permission from dialog.  `adb shell pm grant com.godotopenxrvendors.openxrandroidtrackablessample android.permission.SCENE_UNDERSTANDING_COARSE`.
4: launch app with permissions, then `adb shell pm revoke com.godotopenxrvendors.openxrandroidtrackablessample android.permission.SCENE_UNDERSTANDING_COARSE`.  This terminates the app.  Resuming the app is identical to installing the app without permissions.

Number 3 required me to un-focus then re-focus the app in order to apply the permissions.  The  `on_request_permissions_result` is not called when a permission is granted on the command line, so the updated code does not connect to that signal.

dsnopek (approved):
Thanks!

I did some simple tests of the permission change, and it worked for me

IMO, this PR is ready to merge :-)

BastiaanOlij:
I'm happy for this one to be merged as well!

PR #458, opened 2026-02-10
Implement XR_ANDROID_unbounded_reference_space

devloglogan:
This extension isn't currently a part of our AndroidXR headers or `openxr.h`, but appears to be implemented and works fine on my Galaxy XR headset. For implementation I've copied the def/enums `XR_ANDROID_UNBOUNDED_REFERENCE_SPACE_EXTENSION_NAME` and `XR_REFERENCE_SPACE_TYPE_UNBOUNDED_ANDROID` into `openxr_android_unbounded_reference_space_extension.h`. I figure that's fine and whenever these values are added via a proper header, these copied ones will just be deleted.

### DIFF
--- a/doc_classes/OpenXRAndroidUnboundedReferenceSpaceExtension.xml
+++ b/doc_classes/OpenXRAndroidUnboundedReferenceSpaceExtension.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidUnboundedReferenceSpaceExtension" inherits="OpenXRExtensionWrapper" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Wraps the [code]XR_ANDROID_unbounded_reference_space[/code] extension.
+	</brief_description>
+	<description>
+		Wraps the [code]XR_ANDROID_unbounded_reference_space[/code] extension.
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="is_play_space_unbounded">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if play space is set to unbounded reference space.
+			</description>
+		</method>
+		<method name="is_unbounded_reference_space_available">
+			<return type="bool" />
+			<description>
+				Returns [code]true[/code] if unbounded reference space is available for use.
+			</description>
+		</method>
+		<method name="set_unbounded_reference_space_enabled">
+			<return type="void" />
+			<param index="0" name="enable" type="bool" />
+			<description>
+				Sets the unbounded reference space as enabled or disabled.
+				Note: The AndroidXR Unbounded Reference Space project setting must be enabled in order to enable the reference space.
+			</description>
+		</method>
+	</methods>
+</class>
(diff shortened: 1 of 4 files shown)


doc_classes/OpenXRAndroidUnboundedReferenceSpaceExtension.xml
```diff
@@ -0,0 +1,32 @@
...
+	<brief_description>
+		Wraps the [code]XR_ANDROID_unbounded_reference_space[/code] extension.
+	</brief_description>
+	<description>
+		Wraps the [code]XR_ANDROID_unbounded_reference_space[/code] extension.
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="is_play_space_unbound">
```
dsnopek: This method doesn't appear to be present? But I do think it's a good idea, so the developer can know if this reference space was set successfully or not.

However, I think the name should be:


```suggestion
		<method name="is_play_space_unbounded">
```

plugin/src/main/cpp/include/extensions/openxr_android_unbounded_reference_space_extension.h
```diff
@@ -0,0 +1,84 @@
...
+#pragma once
+
+#include <openxr/openxr.h>
+#include <godot_cpp/classes/open_xr_extension_wrapper.hpp>
+#include <map>
+
+#include "util.h"
+
+#define XR_ANDROID_UNBOUNDED_REFERENCE_SPACE_EXTENSION_NAME "XR_ANDROID_unbounded_reference_space"
+static const XrReferenceSpaceType XR_REFERENCE_SPACE_TYPE_UNBOUNDED_ANDROID = (XrReferenceSpaceType)1000467000;
```
dsnopek: This should be in the .cpp file, not the header. Also, this should at least be `constexpr`:


```suggestion
#define XR_ANDROID_UNBOUNDED_REFERENCE_SPACE_EXTENSION_NAME "XR_ANDROID_unbounded_reference_space"
static constexpr XrReferenceSpaceType XR_REFERENCE_SPACE_TYPE_UNBOUNDED_ANDROID = (XrReferenceSpaceType)1000467000;
```

However, I could see an argument for using `#define` for both, since the first one is using `#define`

m4gr3d (approved):

doc_classes/OpenXRAndroidUnboundedReferenceSpaceExtension.xml
devloglogan: Whoops! Surprised that snuck into the XML, I'd only had in there for a bit while testing. But, you're right, I can see it being useful. So I've added it back in and changed the name.

plugin/src/main/cpp/include/extensions/openxr_android_unbounded_reference_space_extension.h
devloglogan: I've moved it to the cpp file and made it `constexpr`! This is just more in line with how they are defined in `openxr.h` but if you'd like it to be a `#define` instead, let me know.

dsnopek (approved):
Thanks! I haven't tested, but this looks good to me :-)

m4gr3d:
We should hide the boundary type setting and automatically set the manifest config to `XR_BOUNDARY_TYPE_NO_RECOMMENDATION` when this property is enabled.

dsnopek (approved):
LGMT!

m4gr3d (approved):

PR #466, opened 2026-02-25
Implement XR_ANDROID_trackables_object

dhoverb:
Implements [XR_ANDROID_trackables_object](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_ANDROID_trackables_object)
Rebased on https://github.com/GodotVR/godot_openxr_vendors/pull/455

### DIFF
--- a/doc_classes/OpenXRAndroidTrackableObjectTracker.xml
+++ b/doc_classes/OpenXRAndroidTrackableObjectTracker.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidTrackableObjectTracker" inherits="OpenXRAndroidTrackableTracker" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		A tracker for an object (from [code]XR_ANDROID_trackables[/code]).
+	</brief_description>
+	<description>
+		A tracker for an object (from [code]XR_ANDROID_trackables[/code]).
+		See also [OpenXRAndroidTrackablesExtension].
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="get_extents">
+			<return type="Vector3" />
+			<description>
+				Returns the width, height, and depth of the object, as [member Vector3.x], [member Vector3.y], and [member Vector3.z], respectively.
+				The width, height, and depth are centered on the pose from [method OpenXRAndroidTrackableTracker.get_center_pose].
+				For example, a width of [code]2[/code] means the object extends [code]1[/code] to the left and right of the center pose.
+			</description>
+		</method>
+		<method name="get_object_context">
+			<return type="RID" />
+			<description>
+				Get the [code]object context[/code] that was used to create this tracker.
+				See also [OpenXRAndroidTrackablesObjectExtension].
+			</description>
+		</method>
+		<method name="get_object_label">
+			<return type="int" enum="OpenXRAndroidTrackableObjectTracker.ObjectLabel" />
+			<description>
+				Return the [enum ObjectLabel] for this tracker.
+			</description>
+		</method>
+	</methods>
+	<constants>
+		<constant name="OBJECT_LABEL_UNKNOWN" value="0" enum="ObjectLabel">
+			The object is none of the other types.
+		</constant>
+		<constant name="OBJECT_LABEL_KEYBOARD" value="1" enum="ObjectLabel">
+			The object is a keyboard.
+		</constant>
+		<constant name="OBJECT_LABEL_MOUSE" value="2" enum="ObjectLabel">
+			The object is a computer mouse.
+		</constant>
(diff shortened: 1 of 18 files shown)


BastiaanOlij (approved):
Looks good to me! While related to the previous PR, I notice that the plane tracking still has problems with my L shaped desk :)

It picked up my keyboard and mouse right away, that was pretty impressive. This functionality would go well with our "hole punch" approach for immersive applications where we purely cut out the bounding box for the keyboard so you can see your keyboard.
That would actually be a nice sample application to add.

Code wise, this all looks good. Can't wait till we have similar support in spatial entities.

doc_classes/OpenXRAndroidTrackableObjectTracker.xml
```diff
@@ -0,0 +1,49 @@
...
+		A tracker for an object (from [code]XR_ANDROID_trackables[/code]).
+		See also [OpenXRAndroidTrackablesExtension].
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="get_extents">
+			<return type="Vector3" />
+			<description>
+				Returns the width, height, and depth, or [member Vector3.x], [member Vector3.y], and [member Vector3.z] respectively, of the object.
```
dsnopek:
```suggestion
				Returns the width, height, and depth of the object, as [member Vector3.x], [member Vector3.y], and [member Vector3.z], respectively.
```

doc_classes/OpenXRAndroidTrackableObjectTracker.xml
```diff
@@ -0,0 +1,49 @@
...
+				The width, height, and depth are centered on the pose from [method OpenXRAndroidTrackableTracker.get_center_pose].
+				For example, a width of [code]2[/code] means the extents extend [code]1[/code] to the left and right of the center pose.
```
dsnopek:
```suggestion
				For example, a width of [code]2[/code] means the object extends [code]1[/code] to the left and right of the center pose.
```

plugin/src/main/cpp/classes/openxr_android_trackable_object_tracker.cpp
```diff
@@ -0,0 +1,182 @@
...
+	// NOTE: unlike OpenXRAndroidTrackablePlaneTracker, we don't cache XrTrackableObjectANDROID
+	// because retrieving the info every time is not expensive (OpenXRAndroidTrackablePlaneTracker
+	// retrieves an array of vertices, which could be a lot of data)
+	XrTrackableObjectANDROID xr_trackable_object;
+	if (!_get_xrtrackable_info(xr_trackable_object)) {
+		_set_tracking_state(XR_TRACKING_STATE_STOPPED_ANDROID);
+		return;
+	}
+
+	// always set pose, even when old last updated time == xr_trackable_plane.lastUpdatedTime, since
```
dsnopek:
```suggestion
	// always set pose, even when old last updated time == xr_trackable_object.lastUpdatedTime, since
```

plugin/src/main/cpp/extensions/openxr_android_trackables_object_extension.cpp
```diff
@@ -0,0 +1,259 @@
...
+	object_trackable_discovery_cooldown_cur = object_trackable_discovery_cooldown;
+
+	discover_object_trackers(true);
+}
+
+void OpenXRAndroidTrackablesObjectExtension::_on_session_destroyed() {
+	OpenXRAndroidTrackablesExtension *wrapper = OpenXRAndroidTrackablesExtension::get_singleton();
+	ERR_FAIL_NULL(wrapper);
+
+	set_default_object_context_enabled(false);
```
dsnopek: While this is probably fine (because we can't restart a session anyway), this wipes out the value of `default_object_context_enabled` to do this clean-up. We can avoid that via:


```suggestion
	if (default_object_context.is_valid()) {
		free_object_context(default_object_context);
		default_object_context = RID();
	}
```

dsnopek (approved):
Thanks! The code looks good to me

I also tested on the Samsung Galaxy XR, and it seemed to work fine!

It was able to detect one of my mice very reliably, but could only detect my keyboard sporadically (which isn't surprising - it's a split keyboard), but my main mouse it never detected at all. Anyway, I don't see any obvious issues with the PR, I think that's just the detection only being so reliable

doc_classes/OpenXRAndroidTrackableObjectTracker.xml
dhoverb: Done

doc_classes/OpenXRAndroidTrackableObjectTracker.xml
dhoverb: Done

plugin/src/main/cpp/classes/openxr_android_trackable_object_tracker.cpp
dhoverb: Done

plugin/src/main/cpp/extensions/openxr_android_trackables_object_extension.cpp
dhoverb: Done

BastiaanOlij:
Seems ok to merge to me

PR #470, opened 2026-03-04
Add energy multiplier properties to `OpenXRAndroidLightEstimation` node

devloglogan:
This PR adds an `ambient_light_energy_multiplier` and `directional_light_energy_multiplier` property to the `OpenXRAndroidLightEstimation` node. I'm using light estimation in a project and, while it's working correctly, I really wanted to be able to increase the amount of ambient light energy while still letting light estimation affect the world environment.

### DIFF
--- a/doc_classes/OpenXRAndroidLightEstimation.xml
+++ b/doc_classes/OpenXRAndroidLightEstimation.xml
@@ -9,11 +9,18 @@
 	<tutorials>
 	</tutorials>
 	<members>
+		<member name="ambient_light_energy_multiplier" type="float" setter="set_ambient_light_energy_multiplier" getter="get_ambient_light_energy_multiplier" default="1.0">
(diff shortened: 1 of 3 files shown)


dsnopek:
Thanks!

I haven't tested, but the code looks good to me.

However, I wonder if we should also apply this with `AMBIENT_LIGHT_MODE_SPHERICAL_HARMONICS`? It would require passing another uniform to the shader, and then adding a little code to multiply the color.

But I think it would be just as useful as when using only "Color", because the spherical harmonics also give color (just with different colors from different directions) and you could also encounter a situation where the color is too dark and you want to increase it.

m4gr3d (approved):
The code looks good to me!

dsnopek (approved):
Thanks! I haven't tested, but the code looks great to me :-)

PR #474, opened 2026-03-13
Implement XR_ANDROID_mouse_interaction

devloglogan:
What the title says!

Marked as draft because this PR relies on https://github.com/godotengine/godot/pull/117399 which exposes `OpenXRInteractionProfileMetadata` so that the interaction profile can actually be registered.

This requires the `extension_api.json` to be updated, as well as the `godot-cpp` commit which I've updated to `master` for the time being.

It is worth noting that in the XR_ANDROID_mouse_interaction spec the [scroll action](https://developer.android.com/develop/xr/openxr/extensions/XR_ANDROID_mouse_interaction#scroll-action) is specified as a float type with path `…/input/scroll_android/value`, however from my testing this did not work. Instead what worked is a vec2 type with the path `…/input/scroll_android`, the y-value of the vector contained the changing value.


plugin/src/main/cpp/include/classes/openxr_fb_spatial_anchor_manager.h
```diff
@@ -32,6 +32,7 @@
 #include <openxr/openxr.h>
 
 #include <godot_cpp/classes/node.hpp>
+#include <godot_cpp/classes/ref.hpp>
```
devloglogan, line 35: Updating `godot-cpp` creates an include error that required this addition.

dsnopek:
Thanks!

This looks good to me. :-) Now that it can still work on Godot 4.6+, should this be taken out of DRAFT?

I also wonder if we should write anything in `thirdparty/godot_cpp_gdextension_api/README.md` about the hack added here, so that we don't lose track of it?

devloglogan:
@dsnopek both good points! I've taken the PR out of DRAFT and added a note to the `README.md` about the added hack.

plugin/src/main/cpp/extensions/openxr_android_mouse_interaction_extension.cpp
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+void OpenXRAndroidMouseInteractionExtension::_on_instance_destroyed() {
+	cleanup();
+}
+
+bool OpenXRAndroidMouseInteractionExtension::initialize_android_mouse_interaction_extension(const XrInstance &p_instance) {
+	return true;
+}
```
m4gr3d: Is this needed? If there's no initialization to be done, we can just omit the call and remove `_on_instance_created(...)`.

m4gr3d (approved):

plugin/src/main/cpp/extensions/openxr_android_mouse_interaction_extension.cpp
```diff
@@ -0,0 +1,120 @@
...
+	ClassDB::bind_method(D_METHOD("is_enabled"), &OpenXRAndroidMouseInteractionExtension::is_enabled);
+}
+
+bool OpenXRAndroidMouseInteractionExtension::is_enabled() const {
+	return android_mouse_interaction_ext;
+}
+
+PackedStringArray OpenXRAndroidMouseInteractionExtension::_get_suggested_tracker_names() {
+	PackedStringArray arr = {
+		"/user/mouse",
```
dsnopek, line 69: Should we include some indication that this is an Android XR extension in the name? Like, maybe:


```suggestion
		"/user/android_mouse",
```

I worry about the name conflicting with a future mouse extension from another vendor, or even a future EXT or KHR extension

dsnopek (commented):
Now that the Godot change is merged, it should be OK to merge this one too!

plugin/src/main/cpp/extensions/openxr_android_mouse_interaction_extension.cpp
dsnopek: I attempted to make this work, but just couldn't. However, it's `/user/mouse` [in the main spec](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_ANDROID_mouse_interaction) now, so I guess if there's ever an EXT or KHR version, that'll be up to the OpenXR WG to figure out

dsnopek (changes requested):
I did some testing with this PR on Samsung Galaxy XR and it worked great!

Since this PR was created, the `XR_ANDROID_mouse_interaction` extension was [added to the main spec](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_ANDROID_mouse_interaction) and, in comparing that with what's implemented here, there's two differences (one of which has already been brought up):

- The interaction profile path is `/interaction_profiles/android/mouse_interaction_android` (rather than `/interaction_profiles/android/mouse_interaction` as given in the spec)
- The scroll input is a vector2f at `/input/scroll_android` (rather than a float at `/input/scroll_android/value` as given in the spec)

While I think this PR is ready to go as far as code and functionality, I'm not sure we should actually merge it until we can get some feedback from Google on if the spec is incorrect, or if Android XR is going to get updated later to match the spec. I'd like to avoid merging something that's just going to break later.

(I've marked this as "Request changes" only so that we don't accidentally merge this before getting that feedback)

dsnopek (approved):
We got some feedback from Google that there will be some updates, but what we have in this PR should continue to be supported for backwards compatibility, so it should be safe to merge :-)

m4gr3d:
@devloglogan Let's us know if the PR is good to go, and we can merge it!

PR #478, opened 2026-03-25
Do not initialize light estimation extension if unsupported

ashwinvbs:


plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
```diff
@@ -137,6 +137,11 @@ Dictionary OpenXRAndroidLightEstimationExtension::_get_requested_extensions(uint
 }
 
 bool OpenXRAndroidLightEstimationExtension::start_light_estimation() {
+	if (!is_light_estimation_supported()) {
+		UtilityFunctions::printerr("Failed to start light estimation: light estimation not supported on device");
+		return false;
+	}
```
dsnopek: What's here is fine, but I think this could use the error macro:

```suggestion
	ERR_FAIL_COND_V(!is_light_estimation_supported(), false);
```

The programmer calling this method without checking on their own makes this error macro worthy IMO :-)

dsnopek (approved):
Ah, thanks! We should absolutely check this here - sorry for not catching in the original implementation

plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
ashwinvbs: Done

PR #483, opened 2026-04-02
Add OpenXRSpatialEntitiesUnified

dhoverb:
Add a new singleton that can be used for customized spatial entities setup and update.

This singleton is intended to be an alternative to the existing `builtin` project settings (like `enable_builtin_anchor_detection`, `enable_builtin_plane_detection`, and `enable_builtin_marker_tracking`).  The overall benefit is allowing more control over the spatial entity configuration without needing to reimplement context creation + query loop for every app.

Notable changes
* Any number of "unified" spatial contexts can be created, at any time, and can be destroyed at any time
* Unified spatial contexts are just a spatial context created with 1 or more capabilities (like anchors + planes).
* Automatic updates can be toggled at any time, and can be set per-capability.
* Almost all capability options (see `OpenXRSpatialEntitiesUnified::CapabilityOptions`) can be toggled at any time, and other initialization-only capability options can now be configured to use a value other than the project-setting value.
* Previously un-customizable options are now customizable.  One example is that apps can enable/disable polygon2d for planes (this wasn't possible before without reimplementing the context creation + query loop + other data structures).

This PR depends on https://github.com/godotengine/godot/pull/118128


dhoverb:
Documentation check may succeed once https://github.com/GodotVR/godot_openxr_vendors/pull/480 is merged.

plugin/src/main/cpp/include/extensions/openxr_ext_spatial_entities_unified.h
```diff
@@ -0,0 +1,180 @@
...
+	void _on_spatial_discovery_recommended(RID p_spatial_context);
+
+	template <typename T>
+	void _try_start_discovery(RID p_unified_spatial_context, Ref<UnifiedSpatialContextData> p_unified_spatial_context_data, BitField<CapabilityFlags> p_capability_flags, Ref<OpenXRStructureBase> p_next_snapshot_create, Ref<OpenXRStructureBase> p_next_snapshot_query);
+
+	HashMap<RID, Ref<UnifiedSpatialContextData>> unified_spatial_contexts;
+};
+
+class UnifiedSpatialContextData : public RefCounted {
+	GDCLASS(UnifiedSpatialContextData, RefCounted);
```
dsnopek: If this class is registered with Godot, then it should be prefixed with `OpenXR*` like all the other classes in this extension

But does it even need to be exposed to users? If not, it maybe this doesn't need to be a registered Godot class at all? It could just be a plain C++ `struct` or `class`. Given that all it's methods and fields are public, it feels very `struct`-like

dhoverb: I can't pass a plain C++ class/struct to `Callable::bind()`.  It has to be a Godot-type as far as I know.
For now I deleted `GDREGISTER_CLASS(UnifiedSpatialContextData)`.

Is there another option that you had in mind?

dhoverb: Whoops, nevermind, I get an error that the class is missing if I delete that line.
I uploaded a new patchset with the rename.

dsnopek: If it needs to be a Godot class, but we still want to discourage users from interacting with, you can register it with `GDREGISTER_INTERNAL_CLASS()`. It will still be in `ClassDB` (so we still need to prefix with `OpenXR*` so it can't clash with other classes), but you won't need to document it, and developers won't be able to create new instances of it (but, it'll still be bound, such that they could potentially receive an instance of it).

If using it with a `Callable` is the only concern, you could also get away with having the callable actually take `uint64_t` (or, if it will let you, a `void *`) that you cast to/from the real pointer. We already do this for OpenXR structs and a handful of other things in Godot.

dhoverb: > you could also get away with having the callable actually take uint64_t (or, if it will let you, a void *) that you cast to/from the real pointer.

That works for me, thanks!

thirdparty/godot_cpp_gdextension_api/extension_api.json
m4gr3d, line 1: Can you update the `README.md` in the same directory to provide information about how this version was generated.

dhoverb: Done.

doc_classes/OpenXRSpatialEntitiesUnified.xml
```diff
@@ -0,0 +1,268 @@
...
+	<description>
+		The original motivation for this class was to create a single spatial context with multiple capabilities since otherwise some extensions would not work.
+		In addition to unifying multiple spatial capabilities, it also provides more granular control over spatial capabilities.
+		For example, maybe an app wants to discover anchors and planes, but for planes they just want to be able to toggle meshing and automatic updates.
+		[codeblock]
+		func _ready():
+		    var flags := OpenXRSpatialEntitiesUnified.CAPABILITY_FLAGS_PLANE_TRACKING | OpenXRSpatialEntitiesUnified.CAPABILITY_FLAGS_ANCHOR_TRACKING
+		    var options := {
+		        OpenXRSpatialEntitiesUnified.CAPABILITY_OPTIONS_ENABLE_PLANE_MESH_2D : false,
+		        OpenXRSpatialEntitiesUnified.CAPABILITY_OPTIONS_ENABLE_PLANE_POLYGON_2D : false,
```
dsnopek:
```suggestion
		        OpenXRSpatialEntitiesUnified.CAPABILITY_OPTIONS_ENABLE_PLANE_MESH_2D: false,
		        OpenXRSpatialEntitiesUnified.CAPABILITY_OPTIONS_ENABLE_PLANE_POLYGON_2D: false,
```

doc_classes/OpenXRSpatialEntitiesUnified.xml
```diff
@@ -0,0 +1,268 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRSpatialEntitiesUnified" inherits="OpenXRExtensionWrapper" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		An API to simplify code that requires spatial context(s) to be created with multiple spatial capabilities.
+	</brief_description>
+	<description>
+		The original motivation for this class was to create a single spatial context with multiple capabilities since otherwise some extensions would not work.
```
dsnopek: I don't think we should talk about "the original motivation", and instead just state what this class does.

Ex:


```suggestion
		This class allows creating a single spatial context with multiple capabilities, which is required by some OpenXR extensions.
```

doc_classes/OpenXRSpatialEntitiesUnified.xml
```diff
@@ -0,0 +1,268 @@
...
+			<param index="3" name="enable_automatic_discovery_query" type="int" enum="OpenXRSpatialEntitiesUnified.CapabilityFlags" is_bitfield="true" default="127" />
+			<param index="4" name="enable_automatic_update_query" type="int" enum="OpenXRSpatialEntitiesUnified.CapabilityFlags" is_bitfield="true" default="127" />
+			<description>
+				Creates a spatial context with the capabilities specified in [param capability_flags].
+				[param callback] is called when the context is created, with the first parameter being the spatial context [RID].
+				[param capability_options] is a [Dictionary], where each key is [enum CapabilityOptions] and the value depends on the option.  See [enum CapabilityOptions].
+				Any capabilities set in [param enable_automatic_discovery_query] and [param enable_automatic_update_query] will automatically be updated after the context is created, assuming those capabilities are available by the XR runtime.
+				[b]NOTE:[/b] this does not return [OpenXRFutureResult] because sometimes it has async operations that must complete before the context is created.  A valid [param callback] must be provide for this function to be useful.
```
dsnopek:
```suggestion
				[b]NOTE:[/b] this does not return [OpenXRFutureResult] because sometimes it has async operations that must complete before the context is created.  A valid [param callback] must be provided for this function to be useful.
```

doc_classes/OpenXRSpatialEntitiesUnified.xml
```diff
@@ -0,0 +1,268 @@
...
+			</description>
+		</method>
+		<method name="destroy_unified_spatial_context">
+			<return type="void" />
+			<param index="0" name="unified_spatial_context" type="RID" />
+			<description>
+				Destroy the spatial context created by [method create_unified_spatial_context].
+				[b]NOTE:[/b] Even though the spatial context can also be destroyed by [method OpenXRSpatialEntityExtension.free_spatial_context], be sure to call this method to ensure proper cleanup.
```
dsnopek:
```suggestion
				[b]NOTE:[/b] Even though the spatial context can also be destroyed by [method OpenXRSpatialEntityExtension.free_spatial_context], be sure to call this method to ensure proper clean-up.
```

doc_classes/OpenXRSpatialEntitiesUnified.xml
```diff
@@ -0,0 +1,268 @@
...
+				Get the [enum OpenXRSpatialEntitiesUnified.CapabilityFlags] flags that the [param unified_spatial_context] uses.  Note that these flags are either equal, or a subset, of the flags passed in [method create_unified_spatial_context] due to availability of extensions at runtime.
+			</description>
+		</method>
+		<method name="get_capability_options" qualifiers="const">
+			<return type="Array" />
+			<param index="0" name="unified_spatial_context" type="RID" />
+			<description>
+				Returns an [Array] with two [Dictionary]s.
+				Index 0 is a [Dictionary] containing capability options for discovery, where the [Dictionary] maps [enum CapabilityFlags] to a [Dictionary], which maps [enum CapabilityOptions] to a [code]boolean[/code] indicating if it's enabled or not.
+				Index 1 is a [Dictionary] containing capability options for update, where the [Dictionary] maps [enum CapabilityFlags] to a [Dictionary], which maps [enum CapabilityOptions] to a [code]boolean[/code] indicating if it's enabled or not.
```
dsnopek: Why return an array with two dictionaries, rather than two different methods which each return a dictionary?

doc_classes/OpenXRSpatialEntitiesUnified.xml
```diff
@@ -0,0 +1,268 @@
...
+			<param index="0" name="unified_spatial_context" type="RID" />
+			<param index="1" name="capability_flags" type="int" enum="OpenXRSpatialEntitiesUnified.CapabilityFlags" is_bitfield="true" />
+			<description>
+				Check if the given [param capability_flags] are enabled for [param unified_spatial_context].
+				All of the flags must be present to return [code]true[/code].
+				This method is provided for convenience; it turns this
+				[codeblock]
+				var flags := OpenXRSpatialEntitiesUnified.CAPABILITY_FLAGS_ANCHOR | OpenXRSpatialEntitiesUnified.CAPABILITY_FLAGS_PLANE
+				if flags == (OpenXRSpatialEntitiesUnified.get_capability_flags(context) &amp; flags):
+				    # do something
```
dsnopek, line 133: So that this is valid GDScript:

```suggestion
				    # do something
				    pass
```

doc_classes/OpenXRSpatialEntitiesUnified.xml
```diff
@@ -0,0 +1,268 @@
...
+
+				[/codeblock]
+				Into
+				[codeblock]
+				if OpenXRSpatialEntitiesUnified.is_capability_supported(context, OpenXRSpatialEntitiesUnified.CAPABILITY_FLAGS_ANCHOR | OpenXRSpatialEntitiesUnified.CAPABILITY_FLAGS_PLANE):
+				    # do something
```
dsnopek, line 140:
```suggestion
				    # do something
				    pass
```

doc_classes/OpenXRSpatialEntitiesUnified.xml
```diff
@@ -0,0 +1,268 @@
...
+				Set [param discovery] to [code]false[/code] to set automatic updates for update query.
+			</description>
+		</method>
+	</methods>
+	<constants>
+		<constant name="CAPABILITY_FLAGS_NONE" value="0" enum="CapabilityFlags" is_bitfield="true">
+			This is a value to check for the absence of flags, for example:
+			[codeblock]
+			if my_capability_flags == CAPABILITY_FLAGS_NONE:
+			    # there are no flags set
```
dsnopek, line 165:
```suggestion
			    # there are no flags set
			    pass
```

plugin/src/main/cpp/include/extensions/openxr_ext_spatial_entities_unified.h
```diff
@@ -0,0 +1,238 @@
...
+using namespace godot;
+
+// Unified anchor capability configuration
+class OpenXRSpatialEntitiesUnifiedConfigurationAnchor : public OpenXRSpatialCapabilityConfigurationBaseHeader {
+	GDCLASS(OpenXRSpatialEntitiesUnifiedConfigurationAnchor, OpenXRSpatialCapabilityConfigurationBaseHeader);
+
+public:
+	virtual uint64_t _get_configuration() override;
+
+	void init(bool p_enable_persistence);
```
dsnopek: This might be personal preference, but I would give this a longer name:


```suggestion
	void initialize(bool p_enable_persistence);
```

I don't have the data to back this up, but I feel we've got more `initialize()` than `init()` in Godot's public API

plugin/src/main/cpp/include/extensions/openxr_ext_spatial_entities_unified.h
```diff
@@ -0,0 +1,238 @@
...
+};
+
+// Unified plane capability configuration
+class OpenXRSpatialEntitiesUnifiedConfigurationPlane : public OpenXRSpatialCapabilityConfigurationBaseHeader {
+	GDCLASS(OpenXRSpatialEntitiesUnifiedConfigurationPlane, OpenXRSpatialCapabilityConfigurationBaseHeader);
+
+public:
+	virtual uint64_t _get_configuration() override;
+
+	void init(bool p_supports_mesh_2d, bool p_supports_polygons, bool p_supports_labels);
```
dsnopek:
```suggestion
	void initialize(bool p_supports_mesh_2d, bool p_supports_polygons, bool p_supports_labels);
```

plugin/src/main/cpp/extensions/openxr_ext_spatial_entities_unified.cpp
```diff
@@ -0,0 +1,822 @@
...
+	BIND_ENUM_CONSTANT(CAPABILITY_OPTIONS_ENABLE_MARKER_ARUCO);
+	BIND_ENUM_CONSTANT(CAPABILITY_OPTIONS_ENABLE_MARKER_APRIL);
+}
+
+bool OpenXRSpatialEntitiesUnified::UnifiedSpatialContextData::add_discovery_capability_option(BitField<OpenXRSpatialEntitiesUnified::CapabilityFlags> p_capability_flags, CapabilityOptions p_capability_option, const Dictionary &p_capability_options, bool p_adding_required) {
+	return discovery_datas[p_capability_flags].enable_capability_option(p_capability_option, !p_capability_options.has(p_capability_option) || (bool)p_capability_options[p_capability_option], true, p_adding_required);
+}
+
+bool OpenXRSpatialEntitiesUnified::UnifiedSpatialContextData::add_update_capability_option(BitField<OpenXRSpatialEntitiesUnified::CapabilityFlags> p_capability_flags, CapabilityOptions p_capability_option, const Dictionary &p_capability_options, bool p_adding_required) {
+	return update_datas[p_capability_flags].enable_capability_option(p_capability_option, !p_capability_options.has(p_capability_option) || !(bool)p_capability_options[p_capability_option], true, p_adding_required);
```
dsnopek: Is the second `!` correct here? Or should this be:


```suggestion
	return update_datas[p_capability_flags].enable_capability_option(p_capability_option, !p_capability_options.has(p_capability_option) || (bool)p_capability_options[p_capability_option], true, p_adding_required);
```

plugin/src/main/cpp/extensions/openxr_ext_spatial_entities_unified.cpp
```diff
@@ -0,0 +1,822 @@
...
+}
+
+void OpenXRSpatialEntitiesUnified::enable_update_capability_option(RID p_unified_spatial_context, BitField<OpenXRSpatialEntitiesUnified::CapabilityFlags> p_capability_flags, CapabilityOptions p_capability_option, bool p_enable) {
+	_enable_capability_option<false>(p_unified_spatial_context, p_capability_flags, p_capability_option, p_enable);
+}
+
+void OpenXRSpatialEntitiesUnified::_on_create_unified_spatial_context_dependency_completed(RID p_completed_dependency, uint64_t p_unified_spatial_context_data, const Callable &p_user_callback, const Dictionary &p_capability_options, TypedArray<OpenXRSpatialCapabilityConfigurationBaseHeader> p_capability_configurations, Ref<OpenXRStructureBase> p_next, BitField<CapabilityFlags> p_capability_flags_completed, BitField<CapabilityFlags> p_capability_flags_just_completed) {
+	UnifiedSpatialContextData *unified_spatial_context_data = (UnifiedSpatialContextData *)p_unified_spatial_context_data;
+	if (CAPABILITY_FLAGS_ANCHOR_TRACKING != (p_capability_flags_completed & CAPABILITY_FLAGS_ANCHOR_TRACKING) && unified_spatial_context_data->capability_flags.has_flag(CAPABILITY_FLAGS_ANCHOR_TRACKING)) {
```
dsnopek, line 378: Is the `CAPABILITY_FLAGS_ANCHOR_TRACKING != (p_capability_flags_completed & CAPABILITY_FLAGS_ANCHOR_TRACKING)` check correct?

`CAPABILITY_FLAGS_ANCHOR_TRACKING` is a combination of flags, and this would check that _none_ of them are enabled? Although, I'm very unsure I understand what's going on here, so maybe this is correct :-)

But if it is incorrect, this pattern is repeated a couple times below

plugin/src/main/cpp/extensions/openxr_ext_spatial_entities_unified.cpp
```diff
@@ -0,0 +1,822 @@
...
+		// So we don't do this again in any other potential recursive calls
+		p_capability_flags_completed.set_flag(CAPABILITY_FLAGS_MARKER_TRACKING);
+	}
+
+	// Now that capability_flags are finalized, ensure the automatic update flags are set correctly
+	unified_spatial_context_data->capability_flags_automatic_discovery_query = unified_spatial_context_data->capability_flags & unified_spatial_context_data->capability_flags_automatic_discovery_query;
+	unified_spatial_context_data->capability_flags_automatic_update_query = unified_spatial_context_data->capability_flags & unified_spatial_context_data->capability_flags_automatic_update_query;
+
+	OpenXRSpatialEntityExtension *se_extension = OpenXRSpatialEntityExtension::get_singleton();
+	ERR_FAIL_NULL(se_extension);
```
dsnopek: If we return here, will the `unified_spatial_context_data` be freed anywhere?

dsnopek (commented):
Thanks!

I'm not super familiar with EXT spatial entities in general, so I think @BastiaanOlij will ultimately need to review this

But I skimmed the code a bit and had a couple notes/questions :-)

plugin/src/main/cpp/extensions/openxr_ext_spatial_entities_unified.cpp
dhoverb: I would've preferred to make this clearer but wasn't sure how :/

The code in this function is organized like:

```c++
if (/* need to process anchors */) {
  // collect capability option(s) related to anchors, might have a recursive call
} 

if (/* need to process planes */) {
  // collect capability option(s) related to planes, might have a recursive call
}

if (/* need to process markers */) {
  // collect capability option(s) related to markers, might have a recursive call
}

// create spatial context with the collected capability options
```

One way to make this function clearer is to turn each of these massive blocks into their own functions, though that would only clarify what `_on_create_unified_spatial_context_dependency_completed` is doing, not so much their internal logic.
(the functions may have a bunch of out-vars since they modify some incoming state too, so it may be (even more) complex/difficult to read)

Anyway, at the end of each block, you see this:

```c++
  // similar for planes and markers
  p_capability_flags_completed.set_flag(CAPABILITY_FLAGS_ANCHOR_TRACKING);
```

All this says is that we've completed anchors and they can be skipped in potential future recursive calls.

Which is why you see this check before process anchors (and similar for planes and markers):
```c++
// do we still need to process anchors?
if (CAPABILITY_FLAGS_ANCHOR_TRACKING != (p_capability_flags_completed & CAPABILITY_FLAGS_ANCHOR_TRACKING /* maybe more checks*/ )) {
  // anchors are not done
}
```

Currently only anchors has a recursive call (persistence has to wait for the persistence context to be created before continuing).  So these checks for planes and markers is only for consistency and have no additional usefulness currently.

plugin/src/main/cpp/extensions/openxr_ext_spatial_entities_unified.cpp
dhoverb: No, fixed.

plugin/src/main/cpp/extensions/openxr_ext_spatial_entities_unified.cpp
dhoverb: Done.

doc_classes/OpenXRSpatialEntitiesUnified.xml
dhoverb: Done.

doc_classes/OpenXRSpatialEntitiesUnified.xml
dhoverb: Done.

doc_classes/OpenXRSpatialEntitiesUnified.xml
dhoverb: I ended up rephrasing this sentence slightly.

doc_classes/OpenXRSpatialEntitiesUnified.xml
dhoverb: Done.

doc_classes/OpenXRSpatialEntitiesUnified.xml
dhoverb: Done.

I made this function follow the pattern of the other functions, where one parameter is `bool p_discovery`.  If it's true, it'll get the discovery data, and false it'll get the update data.

doc_classes/OpenXRSpatialEntitiesUnified.xml
dhoverb: Done.

doc_classes/OpenXRSpatialEntitiesUnified.xml
dhoverb: Done.

doc_classes/OpenXRSpatialEntitiesUnified.xml
dhoverb: Done.

plugin/src/main/cpp/include/extensions/openxr_ext_spatial_entities_unified.h
dhoverb: Done.

plugin/src/main/cpp/include/extensions/openxr_ext_spatial_entities_unified.h
dhoverb: Done.

BastiaanOlij:
Couldn't get this to compile (yet), might be a step I missed with it complaining about a variant conversion.

Anyway, code wise there isn't much I can add that @dsnopek hasn't already said. I don't know if we had any discussion in the past yet for the use of `auto` and using  Yoda notation. That's generally frowned upon in core but I haven't got a strong opinion either way for using it in vendors.

Functionally I think I follow how this all works. It reproduces what we do in core but trying to combine everything in a single unified class. I'm not sure which approach is better, it feels to me we're still in the same boat I was when implementing things in core. It's hard to see how this can extend once more types of capabilities and features are added. Especially with the issues we've had in the past when combining things that can't be combined.
Instead of writing our own classes to manage a new context for a new type of capability, this class will have to be enhanced to understand those capabilities.
The flip side of that argument is that we currently have limits enhancing the current capabilities with new features.  We can't do this in core, this will potentially allow us to do so on the vendor side.

I don't have an answer here yet but I'm worried about us going in a direction that long term may be incorrect.

dhoverb:
> Especially with the issues we've had in the past when combining things that can't be combined.

If something cannot be combined with something else, this class would block or limit it, or make it clear to the user why two particular extensions cannot be used together.

> Instead of writing our own classes to manage a new context for a new type of capability, this class will have to be enhanced to understand those capabilities.

That's correct.  As new spatial entities are added, this class must be updated to support those new spatial entities.

> The flip side of that argument is that we currently have limits enhancing the current capabilities with new features.

This is exactly why I wrote this, since it's otherwise impractical for anyone to use new future spatial entities that depends on existing spatial entities.  It's definitely possible (you could write OpenXRSpatialEntitiesUnified in GDScript), but it would be comparatively inefficient performance-wise as well as programmer-time-wise since _every_ app will have to implement some form of it.

The reason why I say "every app will have to implement some form of it" is because there is no automatic update/discovery query for custom spatial entity contexts.  The moment you need to combine capabilities, you're on your own to correctly query updates and discoveries (and all of the bookkeeping that goes along with it).  Since spatial entities can build on other spatial entities, this would mean every app would have to reimplement some form of OpenXRSpatialEntitiesUnified.

Some alternatives to this PR:
1. put it into a new GDExtension (in a new github repo).  That way godot-xr-vendors does not commit to it, and, developers have an option for using new spatial entity extensions without needing to reimplement update/discovery logic.
2. update the godot-core implementation to allow automatic update/discovery queries for custom spatial entity contexts.  It's been some time since I've looked at spatial entities, so it wouldn't surprise me if there is a massive caveat that would make this solution not work in the long term.  If you're interested in this one, I can take a deeper look to determine if this option is practical.
3. Do both options 1 and 2

I'm open to other suggestions as well.

PR #487, opened 2026-04-23
Add safety checks in `OpenXRFbSpaceWarpExtension`

dsnopek:
For reasons I still haven't figured out, when using Godot 4.7-dev5 or `master` with space warp, it will crash, because `_on_pre_render()` will run once before `_on_main_swapchains_created()`.

Regardless, `OpenXRFbSpaceWarpExtension::_on_pre_render()` was using a bunch of stuff without checking if it existed, so this PR adds those checks, which we should probably do anyway.

**UPDATE:** I switched from `_on_pre_render()` to `_on_pre_draw_viewport()` - I think it makes more sense than `_on_pre_render()`, since `_on_pre_render()` will still run even if there is no viewport with `use_xr = true`

The one that prevents the crash is:

```
	int view_count = openxr_interface->get_view_count();
	ERR_FAIL_COND(view_count != space_warp_info.size());
```

Because `_on_main_swapchains_created()` hasn't run yet, the `space_ware_info` is still empty

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_fb_space_warp_extension.cpp
@@ -193,27 +193,39 @@ void OpenXRFbSpaceWarpExtension::_on_main_swapchains_created() {
 	}
 }
 
-void OpenXRFbSpaceWarpExtension::_on_pre_render() {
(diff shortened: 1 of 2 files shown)


m4gr3d (approved):
The code looks good.

PR #491, opened 2026-05-07
Add the Meta anchor API permission when the scene API is enabled

dsnopek:
Per @BastiaanOlij's experience on stream, the anchor API permission may be needed, even if you only want to use the scene API

### DIFF
--- a/plugin/src/main/cpp/export/meta_export_plugin.cpp
+++ b/plugin/src/main/cpp/export/meta_export_plugin.cpp
@@ -405,8 +405,9 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
 		}
 	}
 
-	// Check for anchor api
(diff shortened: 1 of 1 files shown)


plugin/src/main/cpp/export/meta_export_plugin.cpp
```diff
@@ -407,6 +407,7 @@ String MetaEditorExportPlugin::_get_android_manifest_element_contents(const Ref<
 
 	// Check for anchor api
```
m4gr3d: can you update the comment too

m4gr3d (approved):

BastiaanOlij (approved):
Works like a charm!

plugin/src/main/cpp/export/meta_export_plugin.cpp
dsnopek: Comment updated in my latest push!

PR #497, opened 2026-05-15
Fix permission handling in AndroidXR Depth Texture Extension

ashwinvbs:
On starting app that uses depth extension for the first time, permission prompts are displayed. As necessary permissions are absent during this period, depth texture extension fails to fully initialize. Specifically, supported depth buffer resolutions are successfully populated but further operations fail.

Discarding the resolutions on this failure prevents recovering from this state.

This patch also updates the depth texture sample to recover on initial run after permission is granted.


dsnopek (approved):
Thanks! The changes look good, and fixed the issue in my testing :-)

PR #498, opened 2026-05-15
Add logic to manage XR Streaming client in supported scenarios

ashwinvbs:
XR Streaming client is auto started if the following conditions are met
1. XR/OpenXR/AndroidXR/Auto Start direct Preview Client is enabled in editor settings.
2. Android XR Streaming runtime is being used by app debugged in editor.
3. Android XR device is connected and is the only device adb can connect to. Scenarios where multiple devices are detected by adb is not handled.
4. XR Streaming client is installed on the connected device.

If client was not running on device before the debugged app is started, the client is started on app start, and closed on app exit.

If the client was running on device before the debugged app started, the client is brought into focus (if not already), but left open on app exit.

The debugged application relies on XrInstanceProperties.runtimeName to determine if XR Streaming runtime is being used. But before editor singletons are registered, allowing communication with debug interface, Godot halts, waiting for client connection. This forces us to start the client from the OpenXRExtensionWrapper implementation.

On exit, the debugged process is killed. As the cleanup functions are not called, the function to stop the client (when appropriate) is called from editor debugger plugin.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_session_helper_extension.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_session_helper_extension.cpp
@@ -1,0 +1,97 @@
+/**************************************************************************/
+/*  openxr_session_helper_extension.cpp                                   */
+/**************************************************************************/
+/*                       This file is part of:                            */
+/*                              GODOT XR                                  */
+/*                      https://godotengine.org                           */
+/**************************************************************************/
+/* Copyright (c) 2022-present Godot XR contributors (see CONTRIBUTORS.md) */
+/*                                                                        */
+/* Permission is hereby granted, free of charge, to any person obtaining  */
+/* a copy of this software and associated documentation files (the        */
+/* "Software"), to deal in the Software without restriction, including    */
+/* without limitation the rights to use, copy, modify, merge, publish,    */
+/* distribute, sublicense, and/or sell copies of the Software, and to     */
+/* permit persons to whom the Software is furnished to do so, subject to  */
+/* the following conditions:                                              */
+/*                                                                        */
+/* The above copyright notice and this permission notice shall be         */
+/* included in all copies or substantial portions of the Software.        */
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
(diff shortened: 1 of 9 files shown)


plugin/src/main/cpp/extensions/openxr_session_helper_extension.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include "openxr_android_xrs_utils.h"
+
+using namespace godot;
+
+OpenXRSessionHelperExtension *OpenXRSessionHelperExtension::singleton = nullptr;
+
+OpenXRSessionHelperExtension *OpenXRSessionHelperExtension::get_singleton() {
+	if (singleton == nullptr) {
+		memnew(OpenXRSessionHelperExtension());
```
m4gr3d, line 47:
```suggestion
		singleton = memnew(OpenXRSessionHelperExtension());
```

plugin/src/main/cpp/openxr_android_xrs_utils.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+	Ref<EditorSettings> editor_settings = editor_interface->get_editor_settings();
+	ERR_FAIL_COND_V(editor_settings.is_null(), "");
+
+	return editor_settings->get_setting("export/android/android_sdk_path");
+}
+
+// This method is called from the debugged application as well as the editor process
+String get_adb_path() {
+	// In the editor, we have access to Editor settings and can get adbn path from there.
```
m4gr3d:
```suggestion
	// In the editor, we have access to Editor settings and can get adb path from there.
```

plugin/src/main/cpp/extensions/openxr_session_helper_extension.cpp
ashwinvbs: wouldn't this be setting singleton twice?
we set singleton in the OpenXRSessionHelperExtension constructor.

plugin/src/main/cpp/openxr_android_xrs_utils.cpp
ashwinvbs: done

plugin/src/main/cpp/extensions/openxr_session_helper_extension.cpp
BastiaanOlij: Hmm, in most places we don't lazy instantiate singletons in the `get_singleton` method. Risks re-instantiating the singleton after cleanup. So we're mixing different patterns here.

Super nit-pick though, I think what you're doing right now is fine and I'd keep it with the singleton being assigned in the constructor.

BastiaanOlij (approved):
As discussed externally, ideally this logic should be handled by the OpenXR runtime. It is weird having to do this on the editor side. This is a feature that has benefits far beyond the editor.

But seeing that is not a short term option, and with every other platform having the same really irritating behaviour of failing if a dev hasn't started the client app on headset (which often has timed out and shut down between tests), the benefits of having this feature far outweighs any principle standpoint.

Hopefully as devs find out how useful this is, and vendors realise how frustrating the status quo is, it may move Google and others to finally adopt a feature like this in their OpenXR runtimes.

PR #501, opened 2026-05-19
Handle runtime changes to app permissions in face and eye tracking extensions

ashwinvbs:
Creating tracker in _on_state_focused allows handling the following scenarios
1. During first app launch when permissions are granted from the prompt interactively.
2. If the permissions are initially denied, but granted from app settings while app is running.


dsnopek (approved):
Thanks!

I tested this by uninstalling and then reinstalling the body tracking sample (so permissions would be cleared), and it started working immediately after the permissions were approved in the usual prompts

m4gr3d (approved):
Good idea! We should update the sample apps to use the same pattern.

PR #503, opened 2026-05-19
Implement `XR_ANDROID_enumerate_system_extension_properties`

dsnopek:
This PR is attempting to implement `XR_ANDROID_enumerate_system_extension_properties` per the spec in the Android XR docs:

https://developer.android.com/develop/xr/openxr/extensions/XR_ANDROID_enumerate_system_extension_properties

~~Unfortunately, I'm getting a segfault at the moment, so marking as a DRAFT until I figure out the issue~~

### DIFF
--- a/plugin/src/main/cpp/register_types.cpp
+++ b/plugin/src/main/cpp/register_types.cpp
@@ -50,6 +50,7 @@
 #include "export/validation_layers_export_plugin.h"
 
 #include "extensions/openxr_android_device_anchor_persistence_extension.h"
+#include "extensions/openxr_android_enumerate_system_extension_properties_extension.h"
 #include "extensions/openxr_android_environment_depth_extension.h"
 #include "extensions/openxr_android_eye_tracking_extension.h"
 #include "extensions/openxr_android_face_tracking_extension.h"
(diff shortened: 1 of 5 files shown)


dsnopek:
Ugh, the segfault turned out to be a really stupid type-o on my part: I had `get_instance_id()` rather than `get_instance()` :-/

Anyway, taking out of DRAFT because it seems to be working now, in my limited testing

plugin/src/main/cpp/register_types.cpp
```diff
@@ -373,6 +375,10 @@ void initialize_plugin_module(ModuleInitializationLevel p_level) {
 				_register_extension_with_openxr(OpenXRAndroidFaceTrackingExtension::get_singleton());
 			}
 
+			if (_get_bool_project_setting("xr/openxr/extensions/androidxr/enumerate_system_extension_properties")) {
+				_register_extension_with_openxr(OpenXRAndroidEnumerateSystemExtensionPropertiesExtension::get_singleton());
+			}
```
m4gr3d: I'm wondering whether this should be behind a project setting given the functionality it provides is only indirectly 'visible' to end users.
By that I mean that if a runtime supports the extension, I'd expect our logic to always enable it and use it to validate the list of the other extensions to enable.

dsnopek: Yeah, I thought about removing the project setting too...

> the functionality it provides is only indirectly 'visible' to end users.

This isn't true at the moment. This PR only exposes the extension to be directly used - we don't have any internal uses for it yet, because nothing has come up so far.

m4gr3d: Let's remove the project setting gate then. We can always add it back later if we find a need for it.

m4gr3d (approved):

plugin/src/main/cpp/register_types.cpp
dsnopek: I've removed the setting in my latest push!

m4gr3d (approved):

PR #506, opened 2026-06-26
Improve dynamic resolution sample

ashwinvbs:
Adds controls for dynamically loading CPU and GPU.

A static load might not trigger resolution changes on devices with better hardware and/or low default resolution. Allowing for dynamic loading lets the user increase load until the hardware limits are reached.


dsnopek (approved):
Thanks!

I've tested the new sample, and it is a genuinely better test of this functionality :-)

As we discussed elsewhere, we should move this into a new `tests/` directory in a follow-up, since this is more of a "test project" rather than a "sample project"

PR #508, opened 2026-07-10
Warn users that frame synthesis requires "Flip Y" on Android XR

dsnopek:
Adds the warning suggested by @m4gr3d on https://github.com/godotengine/godot/pull/119056#issuecomment-4928242988

~~This also switches to using `EditorExportPreset::get_project_setting()` rather than `ProjectSettings::get_setting_with_override()` so that it takes into account settings that are using feature tags that are specific to the export preset. This way, it'll correctly see that `flip_y.androidxr` is enabled, even if `flip_y` (with no feature tag) is disabled.~~

~~We really should do this for all of our export plugins now that we depend on Godot 4.6+, but I'll do that in another PR~~

**UPDATE:** This PR is now rebased on https://github.com/GodotVR/godot_openxr_vendors/pull/509 which updates all our export plugins to use `EditorExportPreset::get_project_setting()`

Marking as DRAFT until these PRs are merged:

- https://github.com/GodotVR/godot_openxr_vendors/pull/509
- https://github.com/godotengine/godot/pull/119056

### DIFF
--- a/plugin/src/main/cpp/export/android_xr_export_plugin.cpp
+++ b/plugin/src/main/cpp/export/android_xr_export_plugin.cpp
@@ -142,6 +142,12 @@ String AndroidXREditorExportPlugin::_get_export_option_warning(const Ref<EditorE
 		if (!openxr_enabled && _get_bool_option(option)) {
 			return "\"Use experimental features\" is only valid when \"XR Mode\" is \"OpenXR\".\n";
 		}
+	} else if (option == "xr_features/enable_androidxr_plugin") {
+		if (godot::gdextension_interface::godot_version.minor >= 8) {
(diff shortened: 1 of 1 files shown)


m4gr3d (approved):

dsnopek:
Something just occurred to me: we only need to flip Y on Android XR when using Vulkan. So, I need to update this to not show the warning when configured to use OpenGL

dsnopek:
I've updated it to not give the warning when using OpenGL in my latest push

m4gr3d (approved):

dsnopek:
Rebased and taking out of DRAFT now that all this PRs dependencies are merged!

PR #509, opened 2026-07-10
Use `EditorExportPreset::get_project_setting()` on export plugins

dsnopek:
This switches to using `EditorExportPreset::get_project_setting()` rather than `ProjectSettings::get_setting_with_override()` in export plugins, so that it takes into account settings that are using feature tags that are specific to the export preset.

This way, it'll correctly see settings like `flip_y.androidxr` as enabled on the Android XR preset, even if `flip_y` (with no feature tag) is disabled.

**AI Disclosure:** I did the changes to `android_xr_export_plugin.cpp` by hand and also did manual testing to make sure everything still worked as expected (the warnings, option visibility and even `AndroidManifest.xml` changes). However, I then asked Claude Code to make the same changes to all the other `*_export_plugin.cpp` classes based on my diff. I've reviewed (and fixed) the result and I believe this counts as "menial" and similar to "find and replace" per [the 2nd bullet point on this blog post](https://godotengine.org/article/contribution-policy-2026/).

### DIFF
--- a/plugin/src/main/cpp/export/android_xr_export_plugin.cpp
+++ b/plugin/src/main/cpp/export/android_xr_export_plugin.cpp
@@ -29,6 +29,7 @@
 
 #include "export/android_xr_export_plugin.h"
 
+#include <godot_cpp/classes/editor_export_preset.hpp>
 #include <godot_cpp/classes/project_settings.hpp>
 
 using namespace godot;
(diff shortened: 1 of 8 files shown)


plugin/src/main/cpp/export/android_xr_export_plugin.cpp
```diff
@@ -29,6 +29,7 @@
 
 #include "export/android_xr_export_plugin.h"
 
+#include <godot_cpp/classes/editor_export_preset.hpp>
 #include <godot_cpp/classes/project_settings.hpp>
```
m4gr3d, line 33: Are the `project_settings` includes still needed?

m4gr3d (approved):

plugin/src/main/cpp/export/android_xr_export_plugin.cpp
dsnopek: For the code as it is, yes - see the end of the constructor:

```
ProjectSettings::get_singleton()->connect("settings_changed", callable_mp(this, &AndroidXREditorExportPlugin::_project_settings_changed));
```

However, I'm honestly not sure why that is needed?

dsnopek: Oh, it looks like it's due to this: https://github.com/GodotVR/godot_openxr_vendors/pull/272#issuecomment-2769702590

We apparently need that so that the export warnings/visibility will change if the project settings change

PR #511, opened 2026-07-14
Use system property structure to determine passthrough camera support

ashwinvbs:


dsnopek (approved):
Thanks!

I didn't test, but the code looks good to me :-)

PR #510, opened 2026-07-14
Improve AndroidXR extension initialization

ashwinvbs:
1. Add missed availability checks during extension initialization.
2. Check permissions only on Android platforms.
3. Misc fix to avoid accessing uninitialized data after error.

PR #512, opened 2026-07-15
Add permission check for XR_ANDROID_light_estimation

dhoverb:
Add a check for the required permissions for XR_ANDROID_light_estimation.

The pattern this follows is similar to the trackables PRs [like this one](https://github.com/GodotVR/godot_openxr_vendors/blob/master/plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp#L209)

Also updated the sample app and the `OpenXRAndroidLightEstimation` node to adjust to the new permission check.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
@@ -30,6 +30,7 @@
 #include "extensions/openxr_android_light_estimation_extension.h"
 
 #include <godot_cpp/classes/open_xrapi_extension.hpp>
+#include <godot_cpp/classes/os.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
 using namespace godot;
@@ -103,6 +104,37 @@ void OpenXRAndroidLightEstimationExtension::_on_instance_created(uint64_t instan
 	}
 }
 
+void OpenXRAndroidLightEstimationExtension::_on_state_focused() {
+	if (!is_light_estimation_supported()) {
+		return;
+	}
+
+	OS *os = OS::get_singleton();
+	ERR_FAIL_NULL(os);
+
+	if (os->get_name() != "Android") {
+		permissions_granted = true;
+		return;
+	}
+
+	// always check for permissions for every focused event, since this is possible:
+	// 1: the app launched, but required permissions are not available
+	// 2: the user opens Android settings app and enables the permission (pauses this app)
+	// 3: the user switches back to this app (resumes and focuses this app)
+	//
+	// NOTE: the MainLoop's "on_request_permissions_result" signal is not always called, which is why
+	//       we are always checking here.
+	// NOTE: depending on the app's manifest, it may be restarted if permissions are removed or
+	//       enabled
+
+	PackedStringArray granted_permissions = os->get_granted_permissions();
+	permissions_granted = granted_permissions.has("android.permission.SCENE_UNDERSTANDING_COARSE");
+
+	if (!permissions_granted) {
+		WARN_PRINT("OpenXR: XR_ANDROID_light_estimation requires android.permission.SCENE_UNDERSTANDING_COARSE; waiting for it to be granted");
+	}
+}
+
 void OpenXRAndroidLightEstimationExtension::_on_instance_destroyed() {
 	cleanup();
 }
@@ -139,7 +171,11 @@ Dictionary OpenXRAndroidLightEstimationExtension::_get_requested_extensions(uint
 bool OpenXRAndroidLightEstimationExtension::start_light_estimation() {
 	ERR_FAIL_COND_V(!is_light_estimation_supported(), false);
 
-	if (light_estimator != XR_NULL_HANDLE) {
+	if (!permissions_granted) {
+		return false;
+	}
+
+	if (is_light_estimation_started()) {
 		return true;
 	}
 
@@ -159,7 +195,7 @@ bool OpenXRAndroidLightEstimationExtension::start_light_estimation() {
 }
 
 void OpenXRAndroidLightEstimationExtension::stop_light_estimation() {
-	if (light_estimator == XR_NULL_HANDLE) {
+	if (!is_light_estimation_started()) {
 		return;
 	}
 
--- a/plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
+++ b/plugin/src/main/cpp/classes/openxr_android_light_estimation.cpp
@@ -149,7 +149,7 @@ void OpenXRAndroidLightEstimation::_notification(int p_what) {
 			if (xr_server) {
 				Ref<OpenXRInterface> openxr_interface = xr_server->find_interface("OpenXR");
 				if (openxr_interface.is_valid()) {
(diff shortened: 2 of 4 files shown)


plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
```diff
@@ -103,6 +104,48 @@ void OpenXRAndroidLightEstimationExtension::_on_instance_created(uint64_t instan
...
 }
 
+void OpenXRAndroidLightEstimationExtension::_on_session_created(uint64_t p_session_instance) {
+	android_light_estimation_ext = android_light_estimation_ext && system_light_estimation_properties.supportsLightEstimation;
+	if (!android_light_estimation_ext) {
+		return;
+	}
+
+	// assume unavailable until the permission is granted (see _on_state_focused())
+	android_light_estimation_ext = false;
```
dsnopek: I'm not crazy about clearing this flag temporarily. In the past, we only ever cleared them for the whole session, if we discovered that we can't use the extension at all for some reason (like, if one of its functions can't be loaded)

Of course, this is consistent with all the Android XR trackables extensions, so I guess it's fine. But I'd really rather have another flag for this purpose (both here and in trackables).

I find it very confusing that the flag ends up having multiple meanings. I think it'd be clearer if `android_light_estimation_ext` was only about whether or not the OpenXR extension was enabled, and then we had a `has_android_permission` flag (or something) that we could use to track whether we have the permission or not.

plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
```diff
@@ -103,6 +104,48 @@ void OpenXRAndroidLightEstimationExtension::_on_instance_created(uint64_t instan
 	}
 }
 
+void OpenXRAndroidLightEstimationExtension::_on_session_created(uint64_t p_session_instance) {
+	android_light_estimation_ext = android_light_estimation_ext && system_light_estimation_properties.supportsLightEstimation;
```
dsnopek: Similarly here. I'm not crazy about overwriting the `android_light_estimation_ext` flag to depend on the flag from system properties.

I'd rather consistently check `android_light_estimation_ext && system_light_estimation_properties.supportsLightEstimation` - we frequently have an `is_enabled()` or `is_supported()` method to do this

dsnopek (commented):
Thanks!

Overall, I think having the permission check in the extension, rather than user code, does make sense. However, there's some things I'm not crazy about with how the permission check messes with the `android_light_estimation_ext` flag.

Of course, this is consistent with how the Android XR trackables extensions do it, but I guess I never really looked closely enough at that when reviewing the trackables PRs :-)

dhoverb:
Thanks @dsnopek !
I think the latest patchset addresses your concerns.

To summarize:
1. `android_light_estimation_ext`: whether the ANDROID_XR API requested exists and the functions were loaded successfully
2. `is_light_estimation_supported()`: bullet 1 AND if the XrSystem properties enabled it.
3. `permissions_granted`: whether the necessary permissions were granted when the app is (re)focused (only checked and potentially set to true in `_on_focused()` when bullet 2 is true)

Doing this however results in some new errors in logcat when the permission is not granted:

The previous patchset emitted one error, which said:
> OpenXR: XR_ANDROID_light_estimation requires android.permission.SCENE_UNDERSTANDING_COARSE; waiting for it to be granted

The new patchset now emits multiple errors plus the error above, because bullet 1 and 2 are true, but bullet 3 could remain false resulting in this ERR_FAIL to trip:
```
bool OpenXRAndroidLightEstimationExtension::start_light_estimation() {
  ERR_FAIL_COND_V(!is_light_estimation_supported() || !permissions_granted, false);
```

I considered added a new bounded method like:
```
bool are_permissions_granted() const { return permissions_granted; }
```

and then modify the sample app and `OpenXRAndroidLightEstimation`, but I can also see a noisy logcat being useful too.  Let me know what you think since I think either option would work.

plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
```diff
@@ -137,9 +169,9 @@ Dictionary OpenXRAndroidLightEstimationExtension::_get_requested_extensions(uint
 }
 
 bool OpenXRAndroidLightEstimationExtension::start_light_estimation() {
-	ERR_FAIL_COND_V(!is_light_estimation_supported(), false);
+	ERR_FAIL_COND_V(!is_light_estimation_supported() || !permissions_granted, false);
```
dsnopek: I would maybe change this to:


```suggestion
	ERR_FAIL_COND_V(!is_light_estimation_supported(), false);

	if (!permissions_granted) {
		return false;
	}
```

I don't think we should print an error for something that the developer can't do anything about. I think returning `false` to indicate that it was unable to start is sufficient

We could go as far as adding the `are_permissions_granted()` method you suggested, and then the developer could do something about the error, but I'm not sure that's necessary

dsnopek (commented):
Thanks! That looks much better to me :-)

Does my suggestion below get rid of all the extra log noise? Or, is there more?

dhoverb:
> Thanks! That looks much better to me :-)
> 
> Does my suggestion below get rid of all the extra log noise? Or, is there more?

That is a good suggestion.

There is still an extra error (["Unable to start light estimation"](https://github.com/GodotVR/godot_openxr_vendors/pull/512/changes#diff-38b47101458f4cd9eb4197c555c753afc960d1ebbf2724c36b428ee575f8a26bL62)) when permissions aren't granted, but I think it is an improvement.

The only way the sample could prevent itself from printing that is by calling the `are_permissions_granted()` function (though I agree it's not really needed too).

Update patchset contains your suggestion, thanks!

plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
dhoverb: Agreed.  Done.

dsnopek (approved):
Thanks!

This looks good to me :-)

However, I think we should make similar changes to how the extension enabled flags are handled in the trackables extension wrappers (in a follow-up PR)

dhoverb:
> However, I think we should make similar changes to how the extension enabled flags are handled in the trackables extension wrappers (in a follow-up PR)

Agreed - been working on it this morning.  Will hopefully put up a PR before end of day today.

plugin/src/main/cpp/extensions/openxr_android_light_estimation_extension.cpp
```diff
@@ -103,6 +104,37 @@ void OpenXRAndroidLightEstimationExtension::_on_instance_created(uint64_t instan
...
 
+void OpenXRAndroidLightEstimationExtension::_on_state_focused() {
+	if (!is_light_estimation_supported()) {
+		return;
+	}
+
+	OS *os = OS::get_singleton();
+	ERR_FAIL_NULL(os);
+
+	if (os->get_name() != "Android") {
```
m4gr3d, line 115: That seems very brittle, can we check for the `android` feature instead.

m4gr3d: See [comments](https://github.com/GodotVR/godot_openxr_vendors/pull/510#pullrequestreview-4731110395).

PR #514, opened 2026-07-17
Update permission checks for AndroidXR Trackables

dhoverb:
Addresses https://github.com/GodotVR/godot_openxr_vendors/pull/512#pullrequestreview-4724449750

Updated the permissions checks for AndroidXR Trackables (planes, objects, raycast, anchors) using the same pattern as https://github.com/GodotVR/godot_openxr_vendors/pull/512

Additionally:
* Went through each bound method to see if `ERR_FAIL_*` was needed
* Added `are_permissions_granted()` since it seemed to make sense for Trackables
* Updated and tested the sample app (androidxr-trackables-sample) to react to the current state of the permissions
* Added missing `_register_extension_as_singleton()` for `OpenXRAndroidTrackablesObjectExtension` in `register_types.cpp` (it is trivial to move this change into a separate PR if necessary)

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_android_device_anchor_persistence_extension.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_android_device_anchor_persistence_extension.cpp
@@ -99,9 +99,7 @@ void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_instance_created(uint64_
 }
 
 void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_session_created(uint64_t p_session_instance) {
-	available = available && XR_TRUE == anchor_persistence_properties.supportsAnchorPersistence;
-
-	if (!available) {
+	if (!is_device_anchor_persistence_supported()) {
 		return;
 	}
 
@@ -150,21 +148,17 @@ void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_session_created(uint64_t
 		supported_trackable_types.insert(trackable_type);
 	}
 
-	// assume unavailable until the permission is granted (see _on_state_focused())
-	available = false;
-	check_for_permissions = true;
+	// wait for _on_state_focused() to correctly set permissions_granted
+	permissions_granted = false;
 }
 
 void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_state_focused() {
-	if (!available && !check_for_permissions) {
+	if (!is_device_anchor_persistence_supported()) {
 		return;
 	}
 
-	OS *os = OS::get_singleton();
-	ERR_FAIL_NULL(os);
-
-	if (os->get_name() != "Android") {
-		available = true;
+	if (!OpenXRUtilities::supports_runtime_permissions()) {
+		permissions_granted = true;
 		return;
 	}
 
@@ -178,19 +172,19 @@ void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_state_focused() {
 	// NOTE: depending on the app's manifest, it may be restarted if permissions are removed or
 	//       enabled
 
-	// assume unavailable until the permission is granted
-	available = false;
+	OS *os = OS::get_singleton();
+	ERR_FAIL_NULL(os);
 
 	PackedStringArray granted_permissions = os->get_granted_permissions();
-	available = granted_permissions.has("android.permission.SCENE_UNDERSTANDING_COARSE") || granted_permissions.has("android.permission.SCENE_UNDERSTANDING_FINE");
+	permissions_granted = granted_permissions.has("android.permission.SCENE_UNDERSTANDING_COARSE") || granted_permissions.has("android.permission.SCENE_UNDERSTANDING_FINE");
 
-	if (!available) {
-		WARN_PRINT("OpenXR: XR_ANDROID_device_anchor_persistence requires android.permission.SCENE_UNDERSTANDING_COARSE or android.permission.SCENE_UNDERSTANDING_FINE; waiting for the permission to be set before enabling");
+	if (!permissions_granted) {
+		WARN_PRINT("OpenXR: XR_ANDROID_device_anchor_persistence requires android.permission.SCENE_UNDERSTANDING_COARSE or android.permission.SCENE_UNDERSTANDING_FINE; waiting for one of them to be granted");
 	}
 }
 
 void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_process() {
-	if (persisted_anchor_discovery_cooldown < 0 || !OpenXRAndroidTrackablesExtension::get_singleton()->can_create_more_anchors()) {
+	if (!permissions_granted || !is_device_anchor_persistence_supported() || persisted_anchor_discovery_cooldown < 0 || !OpenXRAndroidTrackablesExtension::get_singleton()->can_create_more_anchors()) {
 		return;
 	}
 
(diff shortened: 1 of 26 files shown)


samples/androidxr-trackables-sample/menu_3d/menu_3d.gd
```diff
@@ -64,13 +70,40 @@ func refresh():
 	_push_menu_impl(_menu_stack.pop_back(), true)
 
 
+func _on_openxr_session_focused() -> void:
+	var ret: Array = [true]
+	get_tree().call_group("menu3d_group", "are_permissions_granted", ret)
```
dsnopek: This is an odd way to do this - in particular the `are_permissions_granted()` which sounds like it'll return a bool, but instead it modifies its argument.

Why not do something like:

```gdscript
for menu in get_tree().get_nodes_in_group("menu3d_group"):
  if menu.are_permissions_granted():
    # etc...
```

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
```diff
@@ -205,21 +201,20 @@ void OpenXRAndroidTrackablesExtension::_on_session_created(uint64_t p_session_in
...
-	// assume unavailable until the permission is granted (see _on_state_focused())
-	available = false;
-	check_for_permissions = true;
+	// wait for _on_state_focused() to correctly set permissions_granted
+	permissions_granted = false;
 }
 
 void OpenXRAndroidTrackablesExtension::_on_state_focused() {
-	if (!available && !check_for_permissions) {
+	if (!is_trackables_supported() || !is_anchors_supported()) {
```
dsnopek: This doesn't seem quite right. This will return if trackables is supported, but not anchors, whereas I think trackables on its own still needs the permission?

Should this just be:


```suggestion
	if (!is_trackables_supported()) {
```

samples/androidxr-trackables-sample/menu_3d/menu_3d.gd
dhoverb: Good idea, done.

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
dhoverb: I ended up changing it to this since I think both need the permission

```
	if (!is_trackables_supported() && !is_anchors_supported()) {
```

dsnopek (approved):
Thanks! I didn't test, but the changes look good to me :-)

plugin/src/main/cpp/extensions/openxr_android_device_anchor_persistence_extension.cpp
```diff
@@ -152,21 +150,20 @@ void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_session_created(uint64_t
...
 void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_state_focused() {
-	if (!available && !check_for_permissions) {
+	if (!is_device_anchor_persistence_supported()) {
 		return;
 	}
 
 	OS *os = OS::get_singleton();
 	ERR_FAIL_NULL(os);
 
 	if (os->get_name() != "Android") {
```
m4gr3d: See comments in https://github.com/GodotVR/godot_openxr_vendors/pull/510#pullrequestreview-4731110395

plugin/src/main/cpp/extensions/openxr_android_raycast_extension.cpp
```diff
@@ -163,21 +163,20 @@ void OpenXRAndroidRaycastExtension::_on_session_created(uint64_t p_session_insta
...
 void OpenXRAndroidRaycastExtension::_on_state_focused() {
-	if (!available && !check_for_permissions) {
+	if (!is_raycast_supported()) {
 		return;
 	}
 
 	OS *os = OS::get_singleton();
 	ERR_FAIL_NULL(os);
 
 	if (os->get_name() != "Android") {
```
m4gr3d: See comments in https://github.com/GodotVR/godot_openxr_vendors/pull/510#pullrequestreview-4731110395

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
```diff
@@ -205,21 +201,22 @@ void OpenXRAndroidTrackablesExtension::_on_session_created(uint64_t p_session_in
...
+	// The permissions are required for both, so still check permissions if at least one of them is
+	// supported
+	if (!is_trackables_supported() && !is_anchors_supported()) {
 		return;
 	}
 
 	OS *os = OS::get_singleton();
 	ERR_FAIL_NULL(os);
 
 	if (os->get_name() != "Android") {
```
m4gr3d: See comments in https://github.com/GodotVR/godot_openxr_vendors/pull/510#pullrequestreview-4731110395

plugin/src/main/cpp/extensions/openxr_android_trackables_object_extension.cpp
```diff
@@ -93,21 +93,20 @@ void OpenXRAndroidTrackablesObjectExtension::_on_instance_created(uint64_t p_ins
...
 void OpenXRAndroidTrackablesObjectExtension::_on_state_focused() {
-	if (!available && !check_for_permissions) {
+	if (!is_trackables_object_supported()) {
 		return;
 	}
 
 	OS *os = OS::get_singleton();
 	ERR_FAIL_NULL(os);
 
 	if (os->get_name() != "Android") {
```
m4gr3d: See comments in https://github.com/GodotVR/godot_openxr_vendors/pull/510#pullrequestreview-4731110395

m4gr3d (changes requested):

plugin/src/main/cpp/extensions/openxr_android_device_anchor_persistence_extension.cpp
dhoverb: Done

plugin/src/main/cpp/extensions/openxr_android_raycast_extension.cpp
dhoverb: Done

plugin/src/main/cpp/extensions/openxr_android_trackables_extension.cpp
dhoverb: Done

plugin/src/main/cpp/extensions/openxr_android_trackables_object_extension.cpp
dhoverb: Done

dsnopek (approved):

plugin/src/main/cpp/extensions/openxr_android_device_anchor_persistence_extension.cpp
```diff
@@ -152,21 +150,20 @@ void OpenXRAndroidDeviceAnchorPersistenceExtension::_on_session_created(uint64_t
...
+	if (!is_device_anchor_persistence_supported()) {
 		return;
 	}
 
 	OS *os = OS::get_singleton();
 	ERR_FAIL_NULL(os);
 
-	if (os->get_name() != "Android") {
-		available = true;
+	if (!os->has_feature("android")) {
```
m4gr3d: Can you add a `bool runtime_support_permissions() { return os->has_feature("android"); }` util method in `utils.[h|cpp]` and use it for these checks.

This will allow us to easily update the logic to check for permissions grants in one location if needed.
Feel free to use a different name if you have a better option.

m4gr3d (changes requested):

plugin/src/main/cpp/extensions/openxr_android_device_anchor_persistence_extension.cpp
dsnopek: Maybe `supports_runtime_permissions()` so it starts with a verb?

dhoverb: Done

m4gr3d (approved):
Looks good!

PR #520, opened 2026-07-30
Fix Meta environment depth CPU readback on Vulkan

### LINKED ISSUE #517: Meta environment depth CPU readback fails with Mobile/Vulkan because wrapped texture lacks TEXTURE_USAGE_CAN_COPY_FROM_BIT
**Environment**
Godot: 4.8.dev2.official.7220d456d (Win)
**Godot OpenXR Vendors: 5.1.0**
Device: Meta Quest 3
Extension: XR_META_environment_depth

Seems to date back at least as far as OpenXR Vendors 4.2.2 and godot 4.6.1-stable.
Encountered the error prototyping on latest versions, on Mobile - as now, generally, recommended. So I went back to a stable Compatibility project I have on the older versions, flicked that to Mobile, same problem.

Behaviour
GPU-based Meta environment occlusion works correctly, proving the provider and depth swapchain are active.
However, calling `OpenXRMetaEnvironmentDepthExtension.get_environment_depth_map_async(callback)` under the Mobile/Vulkan renderer invokes the callback but supplies empty Image objects.:

texture_get_data: Texture requires the
`RenderingDevice.TEXTURE_USAGE_CAN_COPY_FROM_BIT`
to be set to be retrieved.

Is it maybe as simple as updating the cpp to set the flag, or is the whole transport chain suspect?

A MRP on this is a little tricky. AFAIK Meta-simulator doesn't support Compatibility (and maybe doesn't populate the depth sensors anyway?)... so it's going to have to run on actual hardware. The plugin, and exports, and Android Build Templates are going to need plumbed in, etc. Discovered this whilst trying to Dev my own project (leveraging the feature), so already find myself unexpectedly under the gun now. Let me log the issue here. Will see if there's a demo scene I can use / mod to show it. Or check my own project is suitably anonymised to share that. The moment I get a sec and a clean env.

The-Cyber-Captain:
Fixes #517 

## Summary

Fix CPU readback of Meta environment-depth maps when using the Mobile/Vulkan renderer.

## Problem

`get_environment_depth_map_async()` previously obtained both depth layers
through `RenderingServer::texture_2d_layer_get()`.

This worked with Compatibility/OpenGL ES, but failed under Vulkan because:

1. The wrapped external depth texture did not include
   `RenderingDevice::TEXTURE_USAGE_CAN_COPY_FROM_BIT`.
2. After enabling copying, synchronous Vulkan readback derived its copyable
   layout from incomplete driver metadata for the externally created image.
   This produced an incorrect row pitch and diagonally wrapped/skewed output.

Using `RenderingDevice::texture_get_data_async()` avoids the layout problem
because it calculates copy regions and row pitches from the high-level texture
dimensions.

The asynchronous readback returns raw `VK_FORMAT_D16_UNORM` texels. These
cannot be used directly as `Image.FORMAT_RH`, whose data is stored as IEEE-754
half floats; directly reinterpreting the bytes produces near-binary and
apparently inverted output.

### DIFF
--- a/plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension.cpp
+++ b/plugin/src/main/cpp/extensions/openxr_meta_environment_depth_extension.cpp
@@ -46,6 +46,7 @@
 
 #include <godot_cpp/classes/camera3d.hpp>
 #include <godot_cpp/classes/engine.hpp>
+#include <godot_cpp/classes/image.hpp>
 #include <godot_cpp/classes/open_xrapi_extension.hpp>
 #include <godot_cpp/classes/project_settings.hpp>
 #include <godot_cpp/classes/rendering_device.hpp>
@@ -54,6 +55,7 @@
 #include <godot_cpp/classes/xr_interface.hpp>
 #include <godot_cpp/classes/xr_server.hpp>
 #include <godot_cpp/templates/vector.hpp>
+#include <godot_cpp/variant/packed_float32_array.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
 using namespace godot;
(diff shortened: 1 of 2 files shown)


The-Cyber-Captain:
Oh FFS. I've set my fork and upstream in parallel... evidently. Had intended to squash those two commits - forgetting I don't have authority to do so here in Official. In my defense, I'm very tired. And typically work in a browser / headset* 🤣

[*Yes, I can't *quite* get a full Dev and build environment running in the Quest. But I'm an early-adopter of that stack, and happy to go full standalone when we can!]

PR #524, opened 2026-08-18
Update to OpenXR 1.1.60 and switch mostly away from Meta's headers

### LINKED ISSUE #304: Remove check for Meta headers when features become part of OpenXR spec

dsnopek:
_Edit_: Closes #304

This updates to OpenXR 1.1.60, which is the same version used in Godot currently. (OpenXR 1.1.62 is the latest, but I don't want to get ahead of Godot.)

All of the OpenXR extensions we implemented using Meta's headers are now part of the OpenXR spec:

- `XR_META_boundary_visibility`
- `XR_META_body_tracking_fidelity`
- `XR_META_body_tracking_calibration`
- `XR_EXTX2_stationary_reference_space` (this became `XR_EXT_stationary_reference_space`)

For the first 3, I just switched them to use the official OpenXR header.

However, the last one (`XR_EXT_stationary_reference_space`) should really be implemented in Godot, because it's an EXT. But, also, I'm not entirely sure how it should be implemented in Godot, because it's a sort of weird reference space. So, I'm punting on that one for now, and continuing to use the Meta headers for the experimental extension.

But, the old URL we were using to download Meta's headers on CI no longer works, so I also updated that to use the GitHub project: https://github.com/meta-quest/Meta-OpenXR-SDK

This _almost_ fixes https://github.com/GodotVR/godot_openxr_vendors/issues/304 - should we count it as fixing it, and open a new issue for the `XR_EXT_stationary_reference_space` situation?


devloglogan (approved):
Thanks, David! LGTM! I agree stationary reference space should be moved to core, I'll make a new issue for this and have this PR close the old one.

m4gr3d (approved):
Looks good.

Merging the current version to unblock CI failures.

PR #526, opened 2026-08-21
Use 4.6 version of Godot Android Library

dsnopek:
We're still using the Godot 4.5 version of the Godot Android Library, even though godot_openxr_vendors v5 depends on Godot 4.6+

I noticed this because `Callable.call()` is private in the Godot 4.5 version, but public in the Godot 4.6 version, and I needed to call it :-)

### DIFF
--- a/config.gradle
+++ b/config.gradle
@@ -1,24 +1,24 @@
 ext {
     versions = [
             gradlePluginVersion     : '8.6.1',
-            compileSdk              : 35,
+            compileSdk              : 36,
             minSdk                  : 24,
-            targetSdk               : 35,
+            targetSdk               : 36,
             // Update the Java version in the github workflows when this is updated.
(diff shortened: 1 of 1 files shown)


m4gr3d (changes requested):
We should as well update the other dependencies to match the latest from the engine.

Let me know if you want me to do the update.

dsnopek:
The latest-latest? Or what we have in 4.6? (Since we still support 4.6+ in this version)

m4gr3d:
> The latest-latest? Or what we have in 4.6? (Since we still support 4.6+ in this version)

We can use the values in the latest 4.6 stable branch

https://github.com/godotengine/godot/blob/4.6.3-stable/platform/android/java/app/config.gradle

dsnopek:
Updated!

m4gr3d (approved):

PR #528, opened 2026-08-24
Add support for `XR_ANDROID_geospatial`

dsnopek:
This adds support for `XR_ANDROID_geospatial` (and `XR_ANDROID_google_cloud_auth` which it depends on)

It does NOT add support for `XR_ANDROID_geospatial_anchor` - that's for an eventual follow-up PR

The one thing I'm unsure about is the `get_geographic_location()` method which is implemented in Kotlin, in order to get the device's GPS location the normal Android way. This _could_ be implemented by the user using `JavaClassWrapper`, but _only_ in Godot 4.7+ (because it needs `create_sam_callback()` or `create_proxy()`), however, this version of godot_openxr_vendors supports Godot 4.6+

An alternative option is leaving that out entirely, and just telling folks they need to use Godot 4.7+ in order to do this? But that would affect the demo as well

What do folks think?


doc_classes/OpenXRAndroidGeospatialExtension.xml
```diff
@@ -0,0 +1,92 @@
...
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="check_vps_availability">
+			<return type="OpenXRFutureResult" />
+			<param index="0" name="latitude" type="float" />
+			<param index="1" name="longitude" type="float" />
+			<description>
+				Checks whether VPS (Visual Positioning System) data is available at the given latitude and longitude. Returns an [OpenXRFutureResult] whose result value is [code]true[/code] if VPS is available, or [code]false[/code] if not.
```
m4gr3d: This should return an OpenXRFutureResult wrapping an enum matching https://registry.khronos.org/OpenXR/specs/1.1/man/html/XrVPSAvailabilityANDROID.html.

This allows us to future-proof the api if the reference enum set ends up changing.

doc_classes/OpenXRAndroidGeospatialExtension.xml
```diff
@@ -0,0 +1,92 @@
...
+			</description>
+		</method>
+		<method name="geospatial_pose_to_transform">
+			<return type="Transform3D" />
+			<param index="0" name="pose" type="OpenXRAndroidGeospatialPose" />
+			<description>
+				Converts an [OpenXRAndroidGeospatialPose] to a [Transform3D] in the current play space. The geospatial tracker must be in the [constant GEOSPATIAL_STATE_RUNNING] state.
+			</description>
+		</method>
+		<method name="get_android_location">
```
m4gr3d:
```suggestion
		<method name="get_geographic_location">
```

doc_classes/OpenXRAndroidGeospatialExtension.xml
```diff
@@ -0,0 +1,92 @@
...
+			<return type="void" />
+			<param index="0" name="callback" type="Callable" />
+			<description>
+				Gets the location using normal Android APIs, not VPS.
```
m4gr3d:
```suggestion
				Gets the device's geographic location using normal Android APIs, not VPS.
```

plugin/src/main/cpp/export/android_xr_export_plugin.cpp
```diff
@@ -160,6 +164,8 @@ bool AndroidXREditorExportPlugin::_get_export_option_visibility(const Ref<godot:
 		return (bool)export_preset->get_project_setting("xr/openxr/extensions/eye_gaze_interaction") || (bool)export_preset->get_project_setting("xr/openxr/extensions/androidxr/eye_tracking");
 	} else if (p_option == "android_xr_features/hand_tracking") {
 		return (bool)export_preset->get_project_setting("xr/openxr/extensions/hand_tracking");
+	} else if (p_option == "android_xr_features/google_play_location_services") {
```
m4gr3d: Where is this option defined?

dsnopek: Whoops! That's left over from something that has been removed. I'll remove it in my next push

plugin/src/main/cpp/extensions/openxr_android_geospatial_extension.cpp
```diff
@@ -0,0 +1,324 @@
...
+	BIND_ENUM_CONSTANT(GEOSPATIAL_STATE_INITIALIZATION_FAILED);
+
+	ADD_SIGNAL(MethodInfo("openxr_android_geospatial_state_changed", PropertyInfo(Variant::INT, "state")));
+}
+
+bool OpenXRAndroidGeospatialExtension::is_geospatial_supported() const {
+	// This is what it should be:
+	//return android_geospatial_ext && system_geospatial_properties.supportsGeospatial;
+
+	// But for testing, let's not check the system properties.
```
m4gr3d: Adding a comment as a TODO to revert the testing change.

dsnopek: Yeah, I just noticed this one too a moment ago :-)

I'll remove it in my next push

doc_classes/OpenXRAndroidGeospatialExtension.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="OpenXRAndroidGeospatialExtension" inherits="OpenXRExtensionWrapper" api_type="extension" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+		Wraps the [code]XR_ANDROID_geospatial[/code] extension.
+	</brief_description>
+	<description>
+		Wraps the [code]XR_ANDROID_geospatial[/code] extension, which provides precise geo-location and orientation using a combination of motion tracking, GPS and other sensors, and Google's Visual Positioning System (VPS). This can provide sub-meter positional accuracy and sub-degree orientation accuracy.
+		Authentication must be configured via [OpenXRAndroidGoogleCloudAuthExtension] before using this extension.
```
m4gr3d, line 8: Is there an error value to indicate to users when they forget to configure the Google cloud auth extension?

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRUtilitiesInternal.kt
```diff
@@ -0,0 +1,121 @@
...
+ */
+class GodotOpenXRLocationFinderInternal(godot: Godot?) : GodotPlugin(godot) {
+	internal companion object {
+		private val TAG = GodotOpenXRLocationFinderInternal::class.java.simpleName
+	}
+
+	override fun getPluginName() = "GodotOpenXRLocationFinderInternal"
+
+	@UsedByGodot
+	@RequiresPermission(anyOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_COARSE_LOCATION])
```
m4gr3d, line 62: The export plugin logic checks for fine location; should coarse location be removed?

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRLocationFinderInternal.kt
```diff
@@ -0,0 +1,121 @@
...
+			invokeCallback(callback, false, 0f, 0f)
+		}
+	}
+
+	private fun hasLocationPermission(context: Context): Boolean {
+		return ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED ||
+			ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED
+	}
+
+	private fun invokeCallback(callback: Callable, success: Boolean, latitude: Float, longitude: Float) {
```
m4gr3d:
```suggestion
	private fun invokeCallback(callback: Callable, success: Boolean, latitude: Float = 0.0f, longitude: Float = 0.0f) {
```

With this update, you can just use `invokeCallback(callback, false)` in the logic above.

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRUtilitiesInternal.kt
dsnopek: I added this due to a linter error that happened if it wasn't there. I think both are needed for the linter to be happy, but I'll try removing coarse location and see if that works. As far as I know, geospatial doesn't actually need coarse location (only fine location)

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRUtilitiesInternal.kt
```diff
@@ -0,0 +1,121 @@
...
+			}
+		} catch (e: SecurityException) {
+			Log.e(TAG, "Unable to find location", e)
+			invokeCallback(callback, false, 0f, 0f)
+		}
+	}
+
+	private fun hasLocationPermission(context: Context): Boolean {
+		return ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED ||
+			ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED
```
m4gr3d, line 118: Similar comment about whether `ACCESS_COARSE_LOCATION` is enough.

dsnopek: For `get_geographic_location()`, you can have only coarse location. However, for geospatial I'm pretty sure fine location is required. Given that this is tied to geospatial, we can probably remove the coarse location check

plugin/src/main/AndroidManifest.xml
```diff
@@ -31,6 +31,10 @@
         <meta-data
             android:name="org.godotengine.plugin.v2.GodotOpenXRHybridAppInternal"
             android:value="org.godotengine.openxr.vendors.GodotOpenXRHybridAppInternal" />
+
+        <meta-data
+            android:name="org.godotengine.plugin.v2.GodotOpenXRLocationFinderInternal"
+            android:value="org.godotengine.openxr.vendors.GodotOpenXRLocationFinderInternal" />
```
m4gr3d: We may want to add this conditionally based on whether the extension is enabled. 

If we take that approach, the class will need to include documentation to include this manifest declaration manually for folks using the library version of the plugin.

dsnopek: Or, could we inject this using the export plugin? I can give that a try

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRLocationFinderInternal.kt
```diff
@@ -0,0 +1,121 @@
...
+import androidx.core.location.LocationManagerCompat
+import org.godotengine.godot.Godot
+import org.godotengine.godot.plugin.GodotPlugin
+import org.godotengine.godot.plugin.UsedByGodot
+import org.godotengine.godot.variant.Callable
+
+/**
+ * A plugin used internally to get the device's current location.
+ */
+class GodotOpenXRLocationFinderInternal(godot: Godot?) : GodotPlugin(godot) {
```
m4gr3d: We have two plugins at the moment:
- GodotOpenXR: specific to OpenXR functionality
- GodotOpenXRHybridAppInternal: specific to hybrid app functionality

Maybe we should add a third to cover (generic) Android functionality / APIs, where we'd dump all non OpenXR, Android specific APIs.

dsnopek: Maybe `GodotOpenXRUtilities`? Or, with `*Internal` at the end?

Having it be generic utilities could solve the previous point somewhat, because it'd be less weird to include a generic utilities singleton always?

doc_classes/OpenXRAndroidGeospatialExtension.xml
dsnopek: I did some experimentation, and the extension itself is super inconsistent about this. On some API calls (like creating a geospatial tracker) we'll get an error if the user isn't authenticated. But when checking for VPS availability, it'll just never resolve the OpenXR future

So, in my latest push, I added a way to check on our end if the user ever successfully authenticated through `OpenXRAndroidGoogleCloudAuthExtension` and give our own error about it

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRLocationFinderInternal.kt
m4gr3d: Yes that sounds good!

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRLocationFinderInternal.kt
dsnopek: Changed in my latest push

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRLocationFinderInternal.kt
dsnopek: Alright! In my latest push, it's now `GodotOpenXRUtilitiesInternal` and I updated some of the method names to account for this class not only being about finding the location

plugin/src/main/AndroidManifest.xml
dsnopek: Now that this is a general utility singleton, I think it's fine being added always

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRUtilitiesInternal.kt
dsnopek: So, geospatial requires fine location. However, this method can work with either coarse or fine location. Now that this is on a general utility class which isn't necessarily tied only to geospatial, I think we should just keep both coarse and fine here

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRUtilitiesInternal.kt
dsnopek: Same note as above: geospatial requires fine location. However, `getLocation()` can work with either coarse or fine location. Now that this is on a general utility class which isn't necessarily tied only to geospatial, I think we should just keep both coarse and fine here

plugin/src/main/cpp/extensions/openxr_android_geospatial_extension.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+bool OpenXRAndroidGeospatialExtension::_on_event_polled(const void *p_event) {
+	if (!android_geospatial_ext) {
+		return false;
+	}
+
+	if (static_cast<const XrEventDataBuffer *>(p_event)->type == XR_TYPE_EVENT_DATA_GEOSPATIAL_TRACKER_STATE_CHANGED_ANDROID) {
+		XrEventDataGeospatialTrackerStateChangedANDROID *geospatial_event = (XrEventDataGeospatialTrackerStateChangedANDROID *)p_event;
+		GeospatialState new_geospatial_state = GEOSPATIAL_STATE_STOPPED;
```
m4gr3d: I think this should be:

```suggestion
		GeospatialState new_geospatial_state = geospatial_state;
```

plugin/src/main/cpp/export/android_xr_export_plugin.cpp
```diff
@@ -148,6 +148,10 @@ String AndroidXREditorExportPlugin::_get_export_option_warning(const Ref<EditorE
 				return "\"Frame Synthesis\" requires enabling \"Flip Y\" on Android XR when using Vulkan.\n";
 			}
 		}
+		bool geospatial_enabled = export_preset->get_project_setting("xr/openxr/extensions/androidxr/geospatial");
```
m4gr3d, line 151: We should also add a warning if `xr/openxr/extensions/androidxr/google_cloud_auth` is not enabled.

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRUtilitiesInternal.kt
```diff
@@ -0,0 +1,121 @@
...
+class GodotOpenXRUtilitiesInternal(godot: Godot?) : GodotPlugin(godot) {
+	internal companion object {
+		private val TAG = GodotOpenXRUtilitiesInternal::class.java.simpleName
+	}
+
+	override fun getPluginName() = "GodotOpenXRUtilitiesInternal"
+
+	@UsedByGodot
+	@RequiresPermission(anyOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_COARSE_LOCATION])
+	private fun findLocation(callback: Callable) {
```
m4gr3d: Let's keep the method public.
```suggestion
	fun findGeographicLocation(callback: Callable) {
```

Can you also a javadocs that describe the function's functionality.

m4gr3d (approved):
The code looks good!

Can you also include a recording of the sample?
I think this is something we should start doing, and include the recording in the sample folder so folks checking out the samples can quickly see what the functionality is supposed to look like.

plugin/src/main/cpp/extensions/openxr_android_geospatial_extension.cpp
dsnopek: Thanks! Changed in my latest push

plugin/src/main/java/org/godotengine/openxr/vendors/GodotOpenXRUtilitiesInternal.kt
dsnopek: Oops! Not sure how that got marked as private. Updated in my latest push!

m4gr3d (approved):

PR #529, opened 2026-08-27
Allow configuring Android XR's spatial API version

dsnopek:
This allows the developer to set the "Spatial API Version" when exporting for Android XR, which is used by the Google Play Store to filter out games/apps that the current device isn't able to run (due to having an older version of the OpenXR runtime, where the required extensions haven't been implemented yet).

The PR can show export warnings, telling the developer that one of the OpenXR extensions they have enabled requires a later "Spatial API Version", and if the app needs that extension in order to function, they should increase the "Spatial API Version" to match.

It's based on the tables in the Android XR docs here:

https://developer.android.com/develop/xr/openxr/extensions

Here's a screenshot of what that looks like:

Selection_525

I'm happy to discuss changing the wording to try and make things clearer!


m4gr3d (approved):
Looks good!

PR #535, opened 2026-09-07
Add support for `XR_META_environment_raycast`

realThirdpartycookie:
Implements [XR_META_environment_raycast](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_META_environment_raycast)

Adds `OpenXRMetaEnvironmentRaycastExtension`, which casts rays against the live environment depth data and returns the hit pose (Z axis along the surface normal) in the play space. The raycaster is created asynchronously through Godot's `OpenXRFutureExtension` when the session starts. Enabling the `xr/openxr/extensions/meta/environment_raycast` project setting adds the `USE_SCENE` permission on export.

Includes the class reference, a manual page and `samples/meta-raycast-sample`, which casts the ray out of the right index finger (hand tracking), the right controller or the headset, shows a reticle on the hit point and places boxes with a pinch or the trigger.

Tested on Meta Quest 3 (Horizon OS v205). The Meta XR Simulator advertises the extension too, but only returns hits on Windows where it provides depth data.

https://github.com/user-attachments/assets/7608bd36-f079-4163-80c1-9d7b6573504f


plugin/src/main/cpp/extensions/openxr_meta_environment_raycast_extension.cpp
```diff
@@ -0,0 +1,251 @@
...
+bool OpenXRMetaEnvironmentRaycastExtension::initialize_meta_environment_raycast_extension(const XrInstance &p_instance) {
+	GDEXTENSION_INIT_XR_FUNC_V(xrCreateEnvironmentRaycasterAsyncMETA);
+	GDEXTENSION_INIT_XR_FUNC_V(xrCreateEnvironmentRaycasterCompleteMETA);
+	GDEXTENSION_INIT_XR_FUNC_V(xrDestroyEnvironmentRaycasterMETA);
+	GDEXTENSION_INIT_XR_FUNC_V(xrPerformEnvironmentRaycastMETA);
+	return true;
+}
+
+void OpenXRMetaEnvironmentRaycastExtension::_bind_methods() {
+	ClassDB::bind_static_method("OpenXRMetaEnvironmentRaycastExtension", D_METHOD("get_singleton"), &OpenXRMetaEnvironmentRaycastExtension::get_singleton);
```
m4gr3d: `get_singleton` should not be bound.

realThirdpartycookie: Removed the binding. The sample now grabs the singleton with `Engine.get_singleton("OpenXRMetaEnvironmentRaycastExtension")`, like the other Meta samples do, and the class reference no longer documents `get_singleton`. Retested on Quest 3.

PR #537, opened 2026-09-08
Update to godot-cpp v10.0-stable

dsnopek:
This doesn't actually update to v10.0-stable, because that release doesn't exist yet, but this is using current `master` which is likely to be very close :-)

~~Once godot-cpp v10.0-stable is actually released, I'll update this PR to use the released version, and then take it out of DRAFT~~

**UPDATE:** godot-cpp v10.0.0-stable is [released](https://github.com/godotengine/godot-cpp/releases/tag/10.0.0-stable)!

But I wanted to put this one up to make sure it passes CI. It seems to work fine in some quick local testing, though!

(FYI, we currently use godot-cpp v10.0-rc1, so we're already on 10.x, just not a stable release)

Almost all of the changes relate to code that was sync'd from Godot to godot-cpp.

The biggest set of changes are to `xr_project_setup_dialog.cpp` which used `HashMap`, but now `HashMap` is kind of a pain to copy around (it must be explicitly copied or moved), however, it didn't even really need `HashMap` anyway, so I replaced it with a `Vector<Pair<String, Variant>>`. This should be slightly more efficient, since we are just looping over them anyway, not doing arbitrary lookups.


dsnopek:
godot-cpp v10.0.0-stable has been [released](https://github.com/godotengine/godot-cpp/releases/tag/10.0.0-stable), so I've updated to the released version and taken this out of DRAFT

m4gr3d (approved):

PR #538, opened 2026-09-16
Update Android XR dynamic resolution sample for Godot 4.8

dsnopek:
Now that https://github.com/godotengine/godot/pull/122512 is merged, on Godot 4.8, the dynamic resolution sample won't show the actual resolution since the render target size won't actually get any smaller

This PR updates the sample to use the new `get_recommended_target_size()` method if it's available, so it'll show the right value on Godot 4.8, while still working for earlier Godot versions


m4gr3d (approved):

BastiaanOlij (approved):
LGTM

